PR #1, opened 2021-08-02
Create Xcode Workspace, Demo, & Package project

scannillo:
## Summary
This PR:
- Adds root-level PayPal.xcworkspace. This workspace will be the main workstation for this project.
   - The workspace contains two xcodeproj files. One is for the Demo, the other is for the framework's source code & tests.
   - The Demo app imports the `Card` and `PayPal` modules. 
- Adds framework targets for `PaymentsCore`, `Card`, and `PayPal` modules.
    - The code in here is "Dummy" code that is included for us to test that the project setup works (meant to be replaced).
    - The `Card` and `PayPal` modules depend on `PaymentsCore`. We can follow this pattern when adding new payment methods in the future.
- Adds unit test targets for each module.
- Adds Demo, UnitTest, Card, PayPal, and PaymentCore schemes.
- Adds a Package.swift & PayPal.podspec.

## Notes
This PR will be much easier to review by checking out the branch and opening `PayPal.xcworkspace`. The auto-generated Xcode code is _a lot_.

**This project requires Xcode 13-beta.**

## What's Next
- Add GitHub Actions CI
    - Add build commands per package manager
- Add UITest & IntegrationTest targets
- Create README from [HackMD iOS RFC ](https://hackmd.io/yFA9PLlsRpSvjxxTg5H-ng)
- Add Rakefile (or Fastlane) setup for running tests locally
    - Add instructions for running these commands in README
- Create new, shared, development team for Demo App signing. (Currently using Braintree Payment Solutions, LLC)
- Add Demo app Icon

## Screenshot
Screen Shot 2021-08-02 at 6 05 40 PM


jonathajones:
Not in the org yet so i can't see how we want to merge - but how do we feel about squashing PRs into single commits when they are merged into main / develop? 

Next question: Do we want to work on our own forks and then submit PRs to the main repo? I could see that being useful to keep the main repository cleaner, with just the main / develop branches, and maybe a few larger feature branches that would also get PR'd to

scannillo:
> Not in the org yet so i can't see how we want to merge - but how do we feel about squashing PRs into single commits when they are merged into main / develop?

If you check out the [HackMD](https://hackmd.io/FNus18IQRUaHkxTMldEfTQ?both) Tim started last week, most of us agreed on squashing PRs into a single commit before merging. I definitely like that approach, as well!

> Next question: Do we want to work on our own forks and then submit PRs to the main repo? I could see that being useful to keep the main repository cleaner, with just the main / develop branches, and maybe a few larger feature branches that would also get PR'd to

Also probably a convo for the HackMD doc. I personally think forks are confusing to work with and prefer just using feature branches off `main`. But we could chat about it further on that doc?

jonathajones (approved):
Not for this PR, but would be good to also discuss approaching UI programmatically vs using storyboard / nibs

sarahkoop:
I know renaming is a pain in Xcode but realized it might make more sense to call `PaymentsCore` -> `CorePayments` to align with the other modules `Payments` suffix. Also fine to keep modules named as is though.

Sources/PayPal/PayPal.swift
```diff
@@ -0,0 +1,10 @@
+#if !COCOAPODS
+import PaymentsCore
+#endif
```
scannillo, line 3: We can revisit if this syntax, or `canImport()` is easier to read / makes more sense.

scannillo: This import statement is not needed (won't compile) when building for CocoaPods. It _is_ needed for building our demo app. We will test in the future whether it is also needed for SPM & revisit the best way to structure these imports then.

PayPal.podspec
```diff
@@ -0,0 +1,27 @@
+Pod::Spec.new do |s|
+  s.name             = "PayPal"
+  s.version          = "0.0.1"
+  s.summary          = "The PayPal iOS SDK: Helps you accept card, PayPal, and alternative payments in your iOS app."
```
jaxdesmarais: nit: It looks like we use "alternative payment methods" instead of "alternative payments" in the docs for BT and PayPal. Should we update this to use the same language?

jaxdesmarais (approved):
This looks great! Thanks for getting this all structured and set up. I left one very small comment just on language, but approving either way.

scannillo:
> I know renaming is a pain in Xcode but realized it might make more sense to call PaymentsCore -> CorePayments to align with the other modules Payments suffix. Also fine to keep modules named as is though.

The other modules, currently don't have the `Payments` suffix. They're just `Card` and `PayPal` for now. I think we can leave as-is. In the future / once things are more fleshed out we can revisit naming.

PayPal.podspec
scannillo: Good call. Let's definitely keep it consistent with the existing docs.
```suggestion
  s.summary          = "The PayPal iOS SDK: Helps you accept card, PayPal, and alternative payment methods in your iOS app."
```

scannillo:
> @jonathajones Not for this PR, but would be good to also discuss approaching UI programmatically vs using storyboard / nibs

Definitely - we should have a convo about this. Do you mean deciding how we'll build the Demo app, or how we'll eventually build out the SDK's UI card fields/components?

jonathajones:
> Definitely - we should have a convo about this. Do you mean deciding how we'll build the Demo app, or how we'll eventually build out the SDK's UI card fields/components?


Both, I think - I've worked almost entirely with programmatic UI over the last couple of years, and I find it really clear and easily configurable.

scannillo:
> Both, I think - I've worked almost entirely with programmatic UI over the last couple of years, and I find it really clear and easily configurable.

Gotcha. Yeh I think BT has mostly leaned toward programatic for the Demo app as well. Currently though the BT Demo is made up of a bunch of different layout types - visual layout, auto layout, some anchor constraints, some xibs. It's really confusing. I am totally down for choosing one and sticking to it throughout!

PR #2, opened 2021-08-03
Add GH Actions CI

scannillo:
## Summary
- Add tests.yml for unit tests (integration & UI tests in the future)
- Add build.yml for testing package manager builds (currently CocoaPods only, SPM sample app to be added in follow up PR)
- Add README badge

## Notes
To learn more about GH Actions, they have great [docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions).


sarahkoop (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,13 @@
...
+on: [pull_request, workflow_dispatch]
+jobs:
+  cocoapods:
+    name: CocoaPods
+    runs-on: macOS-11
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Use Xcode 13
+        run: sudo xcode-select -switch /Applications/Xcode_13.0.app
```
jonathajones, line 11: Maybe a dumb question : What machine is actually running this job? If it's a local machine, would we all need to make sure our Xcode installs are named the same / stored in the same directory?

scannillo: Good question.

So, GitHub Action's macOS-11 image provides Xcode images as well (see docs [here](https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode)). So this isn't a local Xcode, but the path that the GH Actions docs want us to specify to use their hosted Xcode.

.github/workflows/build.yml
```diff
@@ -0,0 +1,13 @@
+name: Build
+on: [pull_request, workflow_dispatch]
```
jonathajones, line 2: When we run this action on a PR, does it automatically target the source branch when running this? Saw this parameter (https://github.com/actions/checkout#checkout-a-different-branch) in the actions and wondering if we would need this, or if github actions are smart enough to run commands on the source branch rather than on main

jonathajones: Actually reading a different part of the docs: 
```
    # The branch, tag or SHA to checkout. When checking out the repository that
    # triggered a workflow, this defaults to the reference or SHA for that event.
    # Otherwise, uses the default branch.
    ref: ''
```
So it seems like it would automatically use the source branch! Neat

.github/workflows/build.yml
```diff
@@ -0,0 +1,13 @@
...
+  cocoapods:
+    name: CocoaPods
+    runs-on: macOS-11
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Use Xcode 13
+        run: sudo xcode-select -switch /Applications/Xcode_13.0.app
+      - name: Run pod lib lint
+        run: pod lib lint
```
jonathajones, line 13: What version of cocoapods does github actions use? Can't seem to find it straightforwardly via documentation, though there are a lot of actions that will help to install and configure it

.github/workflows/build.yml
scannillo: Yup - I think we're good on this.

.github/workflows/build.yml
scannillo: I would assume it's just the latest version. We could include a Gemfile / bundler to make sure each of our development (and this CI) uses the same version.

jonathajones (approved):

.github/workflows/build.yml
jonathajones: I'm creating a gemfile to add Fastlane, so i can add this there

PR #3, opened 2021-08-06
Add Swift Package Manager Test App

scannillo:
### Reason for changes

We want our CI to test that any changes to our SDK will compile properly for SPM integrations. Adding new files & import statements across modules as the package grows can get tricky when supporting multiple package managers. 

_Or maybe I'm just scarred by Objective-C header files._

### Summary of changes

- Add a SPMTest application to the repo. This is a blank app that imports the PayPal iOS SDK via SPM.
- Add GH Actions task to build the SPMTest app based on the feature branch.
- Change `#if !COCOAPODS` to `#if canImport()` statements in source files. For some reason, SPM doesn't like the former.

### Notes

I did a deep dive into SPM command line tools. Unfortunately, there is no equivalent of `pod lib lint` in SPM world. This approach is, to my knowledge, the best we can do to mimic an actual integration.

### Checklist

- ~Added a changelog entry~


.github/workflows/build.yml
```diff
@@ -11,3 +11,17 @@ jobs:
         run: sudo xcode-select -switch /Applications/Xcode_13.0.app
       - name: Run pod lib lint
         run: pod lib lint
+  spm:
+    name: SPM
```
jonathajones: nit: Is the 'name' property ever referenced? maybe worthwhile using the full name rather than the acronym here, just for clarity

jonathajones (approved):

.github/workflows/build.yml
scannillo: This name is what shows up on the bottom of the PR (and in the Actions tab in the repo) to visualize the build. I think most folks looking at this PR will know what SPM is, but totally down to make it the full name if that is more straightforward!

scannillo:
```suggestion
    name: Swift Package Manager
```

jaxdesmarais (approved):
This looks good to me. It does look like there is a fastlane lane for spm: https://docs.fastlane.tools/actions/spm/. I'm not sure if that would be able to replace some aspects of the commands used in the github workflow, but thought I'd send it over incase it can! Approving in either case as I don't think this is a blocker.

scannillo:
> This looks good to me. It does look like there is a fastlane lane for spm: https://docs.fastlane.tools/actions/spm/. I'm not sure if that would be able to replace some aspects of the commands used in the github workflow, but thought I'd send it over incase it can! Approving in either case as I don't think this is a blocker.

Thanks for sending this link over. It seems like the commands they're offering essentially wrap the existing xcodebuild commands for SPM. So that still won't provide us with the ability to test an actual integration/import of the SPM package into an app (build & archive too). It would be awesome if we didn't need to create a separate app for this, but I think for now this is our only option.

jonathajones:
Not really for this PR - but maybe a follow up task could be to pull the SPM logic from github actions and into the fastfile. I can see the value in being able to trigger the test SPM integration locally with one line, and that would keep it consistent with the other actions

scannillo:
> Not really for this PR - but maybe a follow up task could be to pull the SPM logic from github actions and into the fastfile. I can see the value in being able to trigger the test SPM integration locally with one line, and that would keep it consistent with the other actions

Yeh, totally! This is a good call for a follow up PR.

PR #5, opened 2021-08-10
Create Project README

sarahkoop:
### Summary of changes

- Create project README based on RFC

### Checklist

- @scannillo 
- @sarahkoop

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,13 @@ cardClient = CardClient(config: paymentConfig)
 card = Card(number: 4111111111111111, cvv: 123, ...)
 
 // STEP 4: Call checkout method
-cardClient?.checkoutWithCard(orderID: ORDER_ID, card: card) { (result, error) in
-    if (error != nil) {
-        // handle checkout error
-        return
+cardClient?.checkoutWithCard(orderID: ORDER_ID, card: card) { result in
+    switch result {
+    case .success(let orderId):
+      // Send orderID to your server to process the payment
+    case .error(let error):
+      // handle checkout error
     }
-    
-    guard let orderID = result?.orderID else { return }
-    // Send orderID to your server to process the payment
 }
 
 // STEP 5: Send orderID to your server to capture/authorize
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 # PayPal iOS SDK
 
-Welcome to PayPal's iOS SDK. This library will help you accept card, PayPal, Venmo, and alternative payments in your iOS app.
+Welcome to PayPal's iOS SDK. This library will help you accept card, PayPal, Venmo, and alternative payment methods in your iOS app.
 
 ## Support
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,82 @@
...
+
+TBD until development progresses. We will use Rake, Fastlane, or some other tool for easy command-line build tasks.
+
+## Release Process
+
+This SDK follows [Semantic Versioning](https://semver.org/). The release process will be automated via GitHub Actions.
+
+## Analytics
+
+Client analytics will be collected via Lighthouse/FPTI.
```
jonathajones: Is this important to highlight? If the target audience for this readme is end merchants, do they need to know about it?

sarahkoop: This document is currently more of a mix of internal agreed upon tools, process, etc and the start of external facing documentation. I agree this info isn't needed for merchants, so if we don't think it's helpful internally either we can remove it.

README.md
```diff
@@ -1,5 +1,82 @@
...
+// STEP 4: Call checkout method
+cardClient?.checkoutWithCard(orderID: ORDER_ID, card: card) { (result, error) in
+    if (error != nil) {
+        // handle checkout error
+        return
+    }
+    
+    guard let orderID = result?.orderID else { return }
+    // Send orderID to your server to process the payment
+}
```
jonathajones: This snippet could be simplified if we use the `Result` type (even though this is just an example, i think this is clearer): 

```suggestion
cardClient?.checkoutWithCard(orderID: ORDER_ID, card: card) { result in
    switch result {
    case .success(let orderId):
      // Send orderID to your server to process the payment
    case .error(let error):
      // handle checkout error
    }
}
```

README.md
jonathajones: That works! I'm fine leaving it as is if this is for our team as well

README.md
scannillo: Good call! Let's definitely use the Swift `Result` type where we can.

README.md
```diff
@@ -1,5 +1,81 @@
...
+
+### Package Managers
+This SDK supports:
+
+* CocoaPods
+* Swift Package Manager
+
+### Languages
+
+This SDK supports Swift 5.1+. This SDK is written in Swift.
```
jonathajones, line 17: Should we also call out our minimum OS target here?

jonathajones (approved):

README.md
sarahkoop: That seems redundant to me since it's stated a few lines above - but can add it again if we think it's helpful.

jonathajones: Whoops, missed it above. No need!

README.md
```diff
@@ -1,5 +1,81 @@
 # PayPal iOS SDK
 
-![GitHub Actions Tests](https://github.com/paypal/iOS-SDK/workflows/Tests/badge.svg)
+Welcome to PayPal's iOS SDK. This library will help you accept card, PayPal, Venmo, and alternative payments in your iOS app.
```
jaxdesmarais: nit: can we update "alternative payments" => "alternative payment methods"? That appears to be the language we currently use in PayPal and Braintree docs for this.

jaxdesmarais (approved):
Just a small comment - looks good to me otherwise. Thanks for making these updates!

PR #7, opened 2021-08-16
Add unit test code coverage lane

jonathajones:
### Reason for changes
We've stated that we want to maintain high code coverage expectations for all submitted code. If we want to enforce code coverage standards, the best place to do so would be through CI and github actions.

### Summary of changes
- Adds a code coverage lane to our fastfile
- Coverage action will fail if the coverage percentage for any framework is below our minimum coverage threshold
- For any modules below the coverage threshold, the coverage of files within that module will be output as well
- Adds an associated github action
- Unrelated: Previously, the `Card` framework did not contain a shared scheme. This PR adds it

### Screenshots
| When all frameworks are fully tested | With under tested framework |
| --- | --- | 
| image | image |

When a framework is below the passing threshold: 

### Checklist

- [ ] Added a changelog entry

### Authors
- jonathajones


PayPal.xcodeproj/xcshareddata/xcschemes/Card.xcscheme
```diff
@@ -0,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jonathajones, line 1: Under `xcshareddata/xcschemes`, on main we have a scheme for `PaymentsCore` and `PayPal`, but not `Card`. This was automatically generated (maybe when I ran the `Card` scheme?). Not sure how it escaped source control til now.

PayPal.xcodeproj/xcshareddata/xcschemes/UnitTests.xcscheme
```diff
@@ -34,13 +34,28 @@
...
+            buildForArchiving = "NO"
+            buildForAnalyzing = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "PayPal::PayPalTests"
+               BuildableName = "PayPalTests.xctest"
+               BlueprintName = "PayPalTests"
+               ReferencedContainer = "container:PayPal.xcodeproj">
+            </BuildableReference>
+         </BuildActionEntry>
```
jonathajones, line 50: This was also automatically added -there were already entries here for `PaymentsCoreTests` and `CardTests`, so not sure why `PayPalTests` was treated differently until now.

fastlane/Fastfile
```diff
@@ -17,4 +17,69 @@ platform :ios do
       test | xcpretty
       ")
   end
+
+  desc "Runs the test suite and calculates code coverage percentages for each
+  module within the SDK"
+  lane :coverage do
+    MINIMUM_COVERAGE = 100.0
```
jonathajones: This was an assumption on my part

PayPal.xcodeproj/xcshareddata/xcschemes/PayPal.xcscheme
```diff
@@ -14,7 +14,7 @@
             buildForAnalyzing = "YES">
             <BuildableReference
                BuildableIdentifier = "primary"
-               BlueprintIdentifier = "PayPal::PayPal"
+               BlueprintIdentifier = "80DB6F0D26B89D9600277E54"
```
minhthenguyen, line 17: quick question, what is this ID, is it auto generated?

README.md
```diff
@@ -71,6 +71,12 @@ bundle install
 bundle exec fastlane unit_tests
 ```
 
+If you would like to get the code coverage for all of the modules within the `iOS-SDK`, run the following:
+```
+bundle install
+bundle exec fastlane coverage
```
minhthenguyen, line 77: nit: I tried running without `bundle exec` and looks like the outputs are the same.
For simplicity I think maybe `fastlane coverage` will do? :smile:

minhthenguyen (approved):
This looks great! :smile:
I have some questions but also approved.

fastlane/Fastfile
minhthenguyen: I like the 80% per DoD, curious to hear other thoughts :smile:

PayPal.xcodeproj/xcshareddata/xcschemes/PayPal.xcscheme
jonathajones: This is auto generated - Looking through why this change would happen here led me to [this](https://github.com/tuist/XcodeProj/pull/612#pullrequestreview-672012032). Basically, the previous value (`PayPal::PayPal`) is not a UUID, and Xcode generated an actual identifier for the scheme when I was looking at all the schemes through the editor.

 @scannillo - did you by chance use `swift package generate-xcodeproj` to create the project initially?

README.md
scannillo: Using `bundle exec` will ensure that we're all using the same version of Fastlane. If you don't use bundle exec, it will use whatever version you have on your machine (if you even have it on your machine).

PayPal.xcodeproj/xcshareddata/xcschemes/PayPal.xcscheme
scannillo: I definitely did use that command for part of the setup, yup!

README.md
minhthenguyen: Ohhhh I understand now!!!!
Thanks Sammy!! :smile:

jonathajones: Generally thinking i think that just using `fastlane coverage` is totally fine - the reason to use `bundle exec` is to ensure that the right gems are used when running the action. [this SO reply](https://stackoverflow.com/a/6588708) does a better job of explaining it than i can

fastlane/Fastfile
```diff
@@ -17,4 +17,69 @@ platform :ios do
...
+
+    # Run the test lane to get coverage profdata files
+    unit_tests
+
+    # Ensure the code coverage directory is empty before creating new coverage
+    sh("rm -rf coverage")
+
+    # Use xcov to get the coverage percentages
+    xcov(
+      scheme: "PaymentsCore",
```
scannillo: Why just the `PaymentsCore` scheme here? We have a `UnitTests` scheme if we want to hit all of the module's tests at once. Can we use that instead?

PayPal.xcodeproj/xcshareddata/xcschemes/PayPal.xcscheme
minhthenguyen: ah that makes sense!!! Thanks for explaining Jon!! I'll resolve these 2 comments :smile:

fastlane/Fastfile
```diff
@@ -17,4 +17,69 @@ platform :ios do
...
+              get_coverage(function, 2)
+            }
+          end
+        }
+      end
+      UI.message ""
+      coverage
+    }
+
+    #If the overall coverage percentage for any module is below our minimum threshhold, fail the test
```
scannillo, line 69: These comments throughout this file are super helpful 👍

fastlane/Fastfile
jonathajones: Good question! it has to do with how `xccov` (which underpins the `xcov` action) creates coverage data for all targets on build, so running this for each scheme gives us redundant reports (this is how i had it configured initially, and then did some digging to figure out why I was seeing xcov output for all targets three times)

jonathajones: You can configure a scheme to only collect partial coverage data for some targets, but since I wanted to analyze all of our targets for coverage data, i thought this was fine

fastlane/Fastfile
jaxdesmarais: If we wanted to try 100%, now is the time, then we can re-evaluate and go down a bit if we feel like 100% isn't necessary? 🤷‍♂️

fastlane/Fastfile
```diff
@@ -17,4 +17,69 @@ platform :ios do
...
+    unit_tests
+
+    # Ensure the code coverage directory is empty before creating new coverage
+    sh("rm -rf coverage")
+
+    # Use xcov to get the coverage percentages
+    xcov(
+      scheme: "PaymentsCore",
+      workspace: "PayPal.xcworkspace",
+      output_directory: "fastlane/coverage",
```
jaxdesmarais, line 36: This is more just a general question than anything, but with this running on CI, do we need an output? Or is this how the coverage report is generated/interpreted by CI?

jaxdesmarais (approved):
This looks great! Thanks for adding this code coverage, I think it'll be very helpful to have this to ensure proper testing out of the gate. I had a couple of just general questions/comments more for my own understanding than anything else.

fastlane/Fastfile
scannillo: I think 100% isn't always going to be possible, even with our best intentions and good practices. I bet there will end up being situations where we need to cover certain situations via an integration or UI test.

I think in general the code coverage metric isn't the "tell-all". It doesn't test for quality of test so the metric itself, it just checks which lines of code has an associated test. While it is helpful at a high level, we still need to make sure we're actually testing every case.

So with that, something like 80% seems more realistic to me. [Hacking With Swift take](https://www.hackingwithswift.com/articles/144/how-to-measure-code-coverage-in-xcode).

fastlane/Fastfile
scannillo: Good spot - we probably don't need that

fastlane/Fastfile
scannillo: 🤔  I'm still a little confused. This reads like we're only getting coverage for the `PaymentsCore` module.

fastlane/Fastfile
jonathajones: 80% works for me!

fastlane/Fastfile
scannillo: I imagine that it was getting duplicate reports since our modules have dependencies on each other? Maybe it would be more readable to use the partial coverage syntax, or maybe there is one that lets us specify the entire `PayPal.xcworkspace` proj?

fastlane/Fastfile
jonathajones: The coverage output is used on line 43 of this file to get the report JSON and iterate through the coverage data. The reason I went this route is because the default terminal output is minimal (it lists the top level coverage data for each framework, but nothing more granular). It has more detailed information in the reported html / json, so figured it'd be helpful to iterate through the JSON and print out the files / functions that are poorly tested

fastlane/Fastfile
jonathajones: Each scheme has this option in under the testing menu: 
image

By running the tests (happens in this file immediately prior to calling `xcov`) we get an `profdata` file for each target in our unit test suite

Since our schemes by default collect coverage data for every target, running `xcov` separately for each target would sort of re-compare every target to the slice of the `profdata` coverage data that xcode generates when running tests.

Maybe a way to make this clearer is to target `Demo` instead - the effect will be the same (all of our targets will be compared to their code coverage data gathered by the test suite)

fastlane/Fastfile
jaxdesmarais: Gotcha, that makes sense - thanks for the explanation @jonathajones!

fastlane/Fastfile
scannillo: Ah gotcha. Thanks for explaining this 👍

scannillo (approved):
_life in the fastlane_

PR #8, opened 2021-08-17
Add SwiftLint rules and add SwiftLint to GitHub Actions + Fastlane

jaxdesmarais:
### Reason for changes

We've indicated that we would like to use a linter and would like to currently only run it via CI and GitHub Actions. In the future, we can also add a Run Script to ensure linter rules are run at compile time as well (if we want) but for now this is only being added to CI + GitHub Actions.

### Summary of changes

- Adds [Ray Wenderlich swiftlint rules](https://github.com/raywenderlich/swift-style-guide/blob/master/com.raywenderlich.swiftlint.yml) as decided via [slack](https://paypal.slack.com/archives/C02A8BRBVGW/p1628801010052400)
- Adds a `lint` lane to our fastfile
- Adds an associated github action
- Resolves all current linter warnings
- Unrelated to this PR, but the import statements have been updated to `canImport` as this was causing errors locally

### Checklist

- [ ] Added a changelog entry

- jonathajones
- jaxdesmarais


.github/workflows/build.yml
```diff
@@ -11,3 +11,13 @@ jobs:
...
+
+    swiftlint:
+    runs-on: macOS-11
+    steps:
+    - name: Check out repository
+    - uses: actions/checkout@v2
+    - name: Install Bundle
+      run: bundle install
+    - name: Run SwiftLint
+      run: bundle exec fastlane lint
```
jonathajones: The indentation on this is off compared to the cocoapods job

fastlane/Fastfile
```diff
@@ -17,4 +17,15 @@ platform :ios do
       test | xcpretty
       ")
   end
+
+  desc "Runs SwiftLint and returns any errors if our linter rule aren't met"
+  lane :lint do
+    swiftlint(
+      mode: :lint,
+      output_file: "swiftlint.result.json",
```
jonathajones: I think we would want to add this file (`swiftlint.result.json`) to our .gitignore. What do you think?

fastlane/Fastfile
```diff
@@ -17,4 +17,15 @@ platform :ios do
       test | xcpretty
       ")
   end
+
+  desc "Runs SwiftLint and returns any errors if our linter rule aren't met"
+  lane :lint do
+    swiftlint(
```
jonathajones, line 88: Assuming we do not want warnings, would we also want the `strict` param? Assuming we want to have zero warnings in PRs

fastlane/Fastfile
jaxdesmarais: Thinking about it now, since this currently only runs in CI, idk how helpful an `output_file` will be. I think I was thinking about this in the sense of running the fastlane script manually via the command line, so I'm thinking for now we remove this line. Thoughts?

jonathajones: I'm good with removing it entirely 👍

.github/workflows/build.yml
```diff
@@ -11,3 +11,14 @@ jobs:
...
+  swiftlint:
+    name: SwiftLint
+    runs-on: macOS-11
+    steps:
+    - name: Check out repository
+    - uses: actions/checkout@v2
+    - name: Install Bundle
+      run: bundle install
+    - name: Run SwiftLint
+      run: bundle exec fastlane lint
```
jonathajones: Still needs updated indentation - also, `uses` does not need a leading hyphen

.swiftlint.yml
```diff
@@ -0,0 +1,110 @@
...
+  - strict_fileprivate
+  - toggle_bool
+  - trailing_closure
+  - unneeded_parentheses_in_closure_argument
+  - unused_import
+  - vertical_whitespace_closing_braces
+  - vertical_whitespace_opening_braces
+  - yoda_condition
+
+
```
scannillo:
```suggestion
```

.swiftlint.yml
```diff
@@ -0,0 +1,110 @@
...
+custom_rules:
+  array_constructor:
+    name: "Array/Dictionary initializer"
+    regex: '[let,var] .+ = (\[.+\]\(\))'
+    capture_group: 1
+    message: "Use explicit type annotation when initializing empty arrays and dictionaries"
+    severity: warning
+
```
scannillo, line 66:
```suggestion
```

Nitpick to keep this consistent across the file

Demo/Demo/SceneDelegate.swift
```diff
@@ -1,15 +1,13 @@
...
-
     var window: UIWindow?
 
-
     func scene(_ scene: UIScene, willConnectTo session: UISceneSession, options connectionOptions: UIScene.ConnectionOptions) {
         // Use this method to optionally configure and attach the UIWindow `window` to the provided UIWindowScene `scene`.
         // If using a storyboard, the `window` property will automatically be initialized and attached to the scene.
         // This delegate does not imply the connecting scene or session are new (see `application:configurationForConnectingSceneSession` instead).
-        guard let _ = (scene as? UIWindowScene) else { return }
+        guard let scene = (scene as? UIWindowScene) else { return }
```
scannillo: Won't this now lead to an XCode warning for an unused var?

README.md
```diff
@@ -59,6 +59,16 @@ cardClient?.checkoutWithCard(orderID: ORDER_ID, card: card) { result in
 
 ```
 
+## Prerequisites
```
scannillo:
```suggestion
## Development
```

Wonder if this should instead go in a `DEVELOPMENT.md` along with other contributing guidlines.

Sources/PayPal/PayPal.swift
```diff
@@ -1,4 +1,4 @@
-#if !COCOAPODS
+#if canImport(PaymentsCore)
```
scannillo: Oh yeh - I also had to make this change here https://github.com/paypal/iOS-SDK/pull/3. I think we need one more ✔️ 👀  on that one!

fastlane/Fastfile
```diff
@@ -17,4 +17,15 @@ platform :ios do
...
+      mode: :lint,
+      config_file: ".swiftlint.yml",
+      raise_if_swiftlint_error: true,
+      ignore_exit_status: true,
```
scannillo: 🤔  Don't we want this to be false? We don't want to ignore the exist status code for this job?

.swiftlint.yml
```diff
@@ -0,0 +1,110 @@
+# Required Swiftlint Version
```
scannillo, line 2: Maybe would be helpful in this file to include a reference to where we pulled it from with a URL link.

.swiftlint.yml
jaxdesmarais: Yep, makes sense - not sure why the [Ray Wenderlich yml](https://github.com/raywenderlich/swift-style-guide/blob/master/com.raywenderlich.swiftlint.yml) has these extra line breaks, but we certainly don't need them!

Demo/Demo/SceneDelegate.swift
jaxdesmarais: It doesn't look like I'm currently getting any issues in Xcode when building the project with this change 🤔

README.md
jaxdesmarais: Yeah, may be good to start a `DEVELOPMENT.md` for these types of potential setup/local testing steps. Thoughts on doing that now vs leaving in the README for now?

Sources/PayPal/PayPal.swift
jaxdesmarais: I'll take a look at that PR today!

fastlane/Fastfile
jaxdesmarais: Ope, yep you're correct, [this defaults to false](https://docs.fastlane.tools/actions/swiftlint/), so I'll remove this line

Demo/Demo/SceneDelegate.swift
jonathajones: It doesn't throw a swiftlint warning, but I do get an xcode warning on this line when i build the project

jaxdesmarais: 😕 I wonder why I'm not getting a warning when I build the project

jaxdesmarais: Ah, nvm, needed to build the demo not paypal target 🤔

README.md
scannillo: Hm - I would vote it can't hurt to start it now, but up to you really!

Sources/PayPal/PayPal.swift
scannillo: That would be awesome - thank you!

.github/workflows/build.yml
```diff
@@ -11,3 +11,16 @@ jobs:
...
+    runs-on: macOS-11
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Use Xcode 13
+        run: sudo xcode-select -switch /Applications/Xcode_13.0.app     
+      - name: Install Dependencies
+        run: bundle install
+      - name: Run SwiftLint
+        run: bundle exec fastlane lint
```
scannillo: 🤔  Maybe this job should go in the tests.yml instead? And have the build.yml be specific for builds per package manager?

Demo/Demo/SceneDelegate.swift
jaxdesmarais: Removing this line entirely for now to avoid xcode warnings or having to disable a linter rule since we don't need to reference this currently

README.md
jaxdesmarais: Added!

.github/workflows/build.yml
jaxdesmarais: 🤔  Yeah, it's not really a build or a test, so does feel weird having it in either of those. Looking online it looks like folks sometimes have a `.github/workflows/swiftlint.yml` but that seems too specific if we want to add other workflows that also aren't build or test specific. I'm not sure if a `run/compile.yml` or something along those lines makes sense for this type of workflow? Or maybe just doing `.github/workflows/swiftlint.yml` is fine for now and if we have too many workflows that aren't related to build or test, we could create a new workflow for those cases. That's a lot of thoughts, so curious if you/anyone else has other ideas! 😄

scannillo: A separate `swiftlint.yml` seems like a good idea to me!

.swiftlint.yml
```diff
@@ -0,0 +1,109 @@
+# URL: https://github.com/realm/SwiftLint
```
scannillo:
```suggestion
# Reference: https://github.com/realm/SwiftLint
```
nitpick

DEVELOPMENT.md
```diff
@@ -0,0 +1,12 @@
+# PayPal iOS Development Notes
+
+This document outlines development practices that we follow while developing this SDK.
+
+## SwiftLint
+
+If you would like to run SwiftLint outside of CI for all of the paths included in the `.swiftlint.yml`, run the following:
```
scannillo:
```suggestion
To run [SwiftLint](https://github.com/realm/SwiftLint) for all of the paths included in the `.swiftlint.yml`, run the following:
```

Take it or leave it comment

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,16 @@
+name: SwiftLint
+on: [pull_request, workflow_dispatch]
+jobs:
+  swiftlint:
+    name: SwiftLint
```
scannillo, line 5: At the bottom of the PR, the name it reads will be `SwiftLint / SwiftLint (pull_request) Successful in 36s`. Not sure if you want to update the two `name` values here to make that read diffferently.

scannillo (approved):
🏁

jonathajones (approved):

PR #9, opened 2021-09-16
Add Networking Layer

jonathajones:
### Reason for changes
This PR adds a networking layer in `PaymentsCore` that can be used by any of the other modules to handle networking requests

Expected use: 
For any network request, we can create an `APIRequest` type to define the endpoint, headers, body and expected response. Assumption is that the response type is a codable type

When a client needs to make a network request, the signature will look like this: 

```swift 
        let apiClient = APIClient(urlSession: URLSession.shared, environment: .sandbox)

        ...
        let request = AccessTokenRequest(clientID: "")
        apiClient.fetch(endpoint: request) { result, correlationId in
            switch result {
            case .success(let tokenResponse):
            // Can used the typed `tokenResponse` here
            case .failure(let error):
            // Something went wrong
            }
        }
```

In addition to the expected result, we also pass back a `correlationId`, since any request targeting  a PayPal API should also vend a correlation ID that can be used to debug merchant issues. The correlationIds can be saved or discarded by the FI clients.

### Summary of changes

- Created `APIClient` to handle network requests
- Added typed `fetch` function, which takes a requests with an associated response type
- Added mock network request responses

### Checklist

- [ ] Added a changelog entry 

### Authors

- @minhthenguyen 
- @jonathajones  
- @jaxdesmarais


Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
+import Foundation
+
+public class APIClient {
+    public typealias CorrelationId = String
```
jaxdesmarais: `CorrelationId` > `CorrelationID` to match the ID name-spacing used in other parts of the SDK like for `clientID`?

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+        completion: @escaping (Result<T.ResponseType?, CoreError>, CorrelationId?) -> Void
+    ) {
+        guard let request = endpoint.toURLRequest(environment: environment) else {
+            completion(.failure(.noUrlRequest), nil)
+            return
+        }
+
+        let task = urlSession.dataTask(with: request) { data, response, error in
+            guard let response = response as? HTTPURLResponse else { return }
+            let correlationId = response.allHeaderFields["Paypal-Debug-Id"] as? String
```
jaxdesmarais: `Id` > `ID`

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+
+            let finish: (Result<T.ResponseType?, CoreError>) -> Void = { result in
+                /// For discussion:
+                /// When a network request to a PayPal API fails, we have some associated error information in the body data of the response.
+                /// This error data doesn't have the same format, but we should discuss and plan for how we want to surface this to the merchant
```
jaxdesmarais, line 27: I wonder if this is worth bringing up in a meeting to discuss with the API team. It would be great if we had an expected model to follow for all error cases so maybe this is an update that could be made on their end to formalize and error format?

Sources/PaymentsCore/Networking/Enums/CoreError.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+// TODO: Refine how we want to do error handling
+/// For further reading, see https://developer.apple.com/documentation/swift/error
+/// For Discussion:
+/// Error conforming enum vs Error conforming Struct?
```
jaxdesmarais: I'm curious to hear some BT thoughts on this as I think errors are handled in BT as a whole a lot better than the PayPal SDKs! I think a Struct will give us a bit more flexibility to potentially return more details in the error.

Sources/PaymentsCore/Networking/Enums/Environment.swift
```diff
@@ -0,0 +1,20 @@
...
+    case stage
+    case production
+
+    //swiftlint:disable force_unwrapping
+    var baseURL: URL {
+        switch self {
+        case .sandbox:
+            return URL(string: "https://api.sandbox.paypal.com")!
+        case .stage:
+            return URL(string: "https://api.msmaster.qa.paypal.com")!
```
jaxdesmarais: I don't think msmaster is accessible off of PayPal VPN so we may want to either omit this and add a confluence note somewhere if folks want to test on stage or add a comment above this noting that unless you're a PayPal employee this will be of no use to you. 😄

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+public class AccessTokenResponse: Codable {
```
jaxdesmarais: Thoughts on making this a struct vs a class?

Sources/PaymentsCore/Networking/Enums/Environment.swift
minhthenguyen: ooohh this is a great point!!!
I wonder if it sufficient just testing on sandbox and remove stage completely considering our SDK being open source and this is public facing :-?

Sources/PaymentsCore/Networking/APIClient.swift
minhthenguyen: I'd love that (I think our GQL API works like that, and in NXO SDK we just have one error type to decode every errors => makes our lives much easier)
But at the same time, I think it'd be a big effort for the API team + considering Orders API is already live and there are merchants using it, I highly doubt it's gonna be viable... :(

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+
+    public init(urlSession: URLSession = .shared, environment: Environment) {
+        self.urlSession = urlSession
+        self.environment = environment
+    }
+
+    public func fetch<T: APIRequest>(
+        endpoint: T,
+        completion: @escaping (Result<T.ResponseType?, CoreError>, CorrelationId?) -> Void
+    ) {
```
scannillo: Nitpick - this spacing is strange to me.

I feel that one line is fine, since this isn't that long
```
    public func fetch<T: APIRequest>(endpoint: T, completion: @escaping (Result<T.ResponseType?, CoreError>, CorrelationId?) -> Void) {
```

If not, then two lines with `completion` starting the new line and lining up in the `:` for the input params.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Yeh, the errors I think unfortunately will be far from uniform. They can also change without warning so our SDK should be ready for that.

Sources/PaymentsCore/Networking/Enums/CoreError.swift
scannillo: Gonna check out this article & come back to this comment!

Sources/PaymentsCore/Networking/Enums/CoreError.swift
```diff
@@ -0,0 +1,13 @@
...
+public enum CoreError: Error {
+    case networkingError(Error)
+    case decodingError(Error)
+    case noResponseData
+    case noUrlRequest
```
scannillo:
```suggestion
    case noURLRequest
```

Sources/PaymentsCore/Networking/Enums/HTTPMethod.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+public enum HTTPMethod: String {
```
scannillo, line 7: We likely won't need all of these. Maybe let's include the ones we need for now, and then add the others as the need arises.

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
scannillo: struct!

scannillo: They're good for modeling data due to it's "value-type" nature

UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,50 @@
...
+class APIClientTests: XCTestCase {
+    
+    lazy var apiClient: APIClient = {
+        APIClient(
+            urlSession: URLSession(urlProtocol: URLProtocolMock.self),
+            environment: .sandbox
+        )
+    }()
+
+    func testAccessTokenMockRequest_Success() {
```
scannillo: I think we should establish a good test naming convention from the start.

Something like `test<Subject/MethodUnderTestName>_when<Condition>_<ExpectedResult>`

So something like this in these tests:
`testFetch_withValidAccessTokenRequest_returnsValidAccessToken`
`testFetch_withInvalidAccessTokenRequest_returnsError`

I think this will help make the tests readable and understandable as the SDK gains a ton of tests. I think we will thank ourselves later if they're uniformly named.

UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,50 @@
...
+
+        URLProtocolMock.requestResponses.append(MockAccessTokenRequestResponse(success: false))
+
+        apiClient.fetch(endpoint: AccessTokenRequest(clientID: "")) { result, correlationId in
+            switch result {
+            case .success:
+                XCTFail("Should not be able to successfully decode a result")
+            case .failure(let error):
+              XCTAssertNotNil(error)
+            }
```
scannillo:
```suggestion
            case .success:
                XCTFail("Should not be able to successfully decode a result")
            case .failure(let error):
                XCTAssertNotNil(error)
            }
```

UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        let expect = expectation(description: "Get mock response for access token request")
+
+        URLProtocolMock.requestResponses.append(MockAccessTokenRequestResponse(success: false))
+
+        apiClient.fetch(endpoint: AccessTokenRequest(clientID: "")) { result, correlationId in
+            switch result {
+            case .success:
+                XCTFail("Should not be able to successfully decode a result")
+            case .failure(let error):
+              XCTAssertNotNil(error)
```
scannillo: Can we assert on error type here instead of just "not nil"?

scannillo: The `fetch` method should have tests for each error type possible on `CoreError`

UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        APIClient(
+            urlSession: URLSession(urlProtocol: URLProtocolMock.self),
+            environment: .sandbox
+        )
+    }()
+
+    func testAccessTokenMockRequest_Success() {
+        let expect = expectation(description: "Get mock response for access token request")
+
+        URLProtocolMock.requestResponses.append(MockAccessTokenRequestResponse(success: true))
```
scannillo: I think a lot of value of a good unit test suite comes from a test being readable in isolation. So instead of needing a tests to be super "DRY", I think there is more value in being able to go right to a failing test, read it and understand what that test is doing. So like in this example, the `"TestToken"` value is set in some other class through a few  different abstraction levels. I think this might be prematurely optimizing and that it would be easier to read this test and the mock JSON values all inline.

I think as the SDK & test suite scales, this introduces a bigger headache. So let's say a code change or iOS update causes a unit test to fail, now we also have to decipher how that set is set up across files vs just being able to read this one test in order to understand what it is doing.

Let me know if where I am coming from makes sense!

Sources/PaymentsCore/Networking/APIClient.swift
minhthenguyen: I could be wrong, but I think lining up `:` is more of an ObjC convention isn't it? 
But I'm down having this one line :smile:

UnitTests/PaymentsCoreTests/APIClientTests.swift
minhthenguyen: can't speak for Jon and Jax, but I like this a lot!!! :smile:

UnitTests/PaymentsCoreTests/APIClientTests.swift
minhthenguyen: I think it makes sense and that's a great point!!!! What we're having is like the setup we're having in PayPalCheckout and I agree it doesn't look/feel nice asserting things from the mock in the test, but then having the mock object in another place.

What do you suggest when in the future we may use the same mock for multiple tests? Should we have a place dedicated for mocks, or will we agree with mocking similar responses in multiple tests?
I see it's the problem of readability vs clean code, but maybe for tests, readability should be higher priority? What do you think?

Sources/PaymentsCore/Networking/APIClient.swift
jaxdesmarais: Wouldn't this hit our `line_length` linter warning as one line? The [default is 120 characters](https://realm.github.io/SwiftLint/line_length.html) and this looks to be over that. I think maybe this example would be a good reason run our linter during build in Xcode since there are things like this that we may only potentially catch once pushing to GitHub. But that could be a separate discussion. 😄

Sources/PaymentsCore/Networking/Enums/Environment.swift
jaxdesmarais: Yeah, that's what I am more inclined to do as well! I think sandbox is typically sufficient for our testing purposes, but we could also add an internal confluence page for PayPal employees looking to test on msmaster with instructions on how to do that.

UnitTests/PaymentsCoreTests/APIClientTests.swift
jonathajones: I think this sounds really good. Will update!

UnitTests/PaymentsCoreTests/APIClientTests.swift
jonathajones: I think this is a great point, and we could maybe forego adding default mock objects as statics, and instead create them locally from a response code / response data mock. maybe the lines would blur here if  a lot of tests want to use the same mock, but i don't think that should happen too often?

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: > I could be wrong, but I think lining up : is more of an ObjC convention isn't it?

🤔 Pretty sure that rule can apply to Swift, as well! I think having the `(` and `)` on different lines of a function definition is strange and hard to read.

Sources/PaymentsCore/Networking/Enums/Environment.swift
scannillo: +1 to this!

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: lining up the `:` is more of an obj-c thing - xcode doesn't align newlined parameters like this by default in swift, and I haven't seen this before outside obj-c

re: newlining parameters - This was new to me when I came to paypal, but I've grown to really appreciate the way it visually organizes the code. for reference, this is also part of the google [swift style guide](https://google.github.io/swift/#line-wrapping), and they do a way better job explaining it than i can. either every parameter gets wrapped to a new line, or everything should fit on a single line - this can also prevent blocks of code hanging in freespace in the middle of the page,  a la 

```swift
    func someFunctionWithAVeryLongName(and many: String,
                                       additional: String,
                                       parameters: String) {
    }
```

jonathajones: but also agreed that it looked weird to me when I first came here. I would prefer to keep this codestyling, but the most important thing is that we agree on the style and all do the same thing

UnitTests/PaymentsCoreTests/APIClientTests.swift
jaxdesmarais: +1 to this as well, I think that sounds good!

jonathajones: updated!

UnitTests/PaymentsCoreTests/APIClientTests.swift
jonathajones: updated! I think we can create a follow up work item to define how we want to handle networking errors in the SDK. I feel like an `Error` struct makes sense, since we'll have additional context / data that we want to pass back to the merchant, which doesn't lend itself well to using an `enum`

UnitTests/PaymentsCoreTests/APIClientTests.swift
jonathajones: We updated the init for our mocking types to include a response body string and a status code to be vended. We kept the access token response (and error sample) around as a static member of the access token response mock in order to prevent us from copying the same JSON block in multiple tests. How does this sound @scannillo ?

scannillo:
I saw this rename `UnitTests/PayPalTests/PayPal_Tests.swift → PayPalTests/PayPal_Tests.swift` in the PR. I think we should leave them under a directory named `UnitTests` as we will add `IntegrationTests` per module later on. Also prevents too many dirs at the top level, which we will end up having.

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
```
scannillo: Don't think we need `Foundation` here? We should do an audit per file

jonathajones: yep, you're right - will update. good catch

jonathajones:
> I saw this rename `UnitTests/PayPalTests/PayPal_Tests.swift → PayPalTests/PayPal_Tests.swift` in the PR. I think we should leave them under a directory named `UnitTests` as we will add `IntegrationTests` per module later on. Also prevents too many dirs at the top level, which we will end up having.

Yep! this wasn't intentional - somehow moved it to the networking directory, so thought we were just moving it back. Will properly revert this, thanks for catching it

UnitTests/PaymentsCoreTests/EnvironmentTests.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  EnvironmentTests.swift
+//  EnvironmentTests
+//
+//  Created by Jones, Jon on 9/20/21.
+//
+
```
scannillo:
```suggestion
```

Sources/PaymentsCore/Networking/Protocols/APIRequest.swift
```diff
@@ -0,0 +1,40 @@
...
+    var path: String { get }
+    var method: HTTPMethod { get }
+    var headers: [HTTPHeader: String] { get }
+    var queryParameters: [String: String] { get }
+    var body: Data? { get }
+
+    func toURLRequest(environment: Environment) -> URLRequest?
+}
+
+public extension APIRequest {
```
scannillo, line 15: I had no idea this was possible - neat!

Sources/PaymentsCore/Networking/Protocols/APIRequest.swift
```diff
@@ -0,0 +1,40 @@
+import Foundation
+
+public protocol APIRequest {
```
scannillo, line 3: From reading through the code, I wouldn't think that `APIRequest` was a protocol but rather a class. I was thinking .. how can we name this to be more "protocolly". Here's what[ Swift's guidelines ](https://swift.org/documentation/api-design-guidelines/)say:

```
Protocols that describe what something is should read as nouns (e.g. Collection).

Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).
```

I guess this falls in the first category and that technically the name fits the description above. However, just posting this in case anyone else also has this confusion / ideas.

Sources/PaymentsCore/Networking/Protocols/APIRequest.swift
```diff
@@ -0,0 +1,40 @@
...
+    associatedtype ResponseType: Codable
```
scannillo, line 4: THIS IS COOl. WOW SWIFT IS MAGIC

Sources/PaymentsCore/Networking/Enums/CoreError.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+// TODO: Refine how we want to do error handling
+/// For further reading, see https://developer.apple.com/documentation/swift/error
+/// For Discussion:
+/// Error conforming enum vs Error conforming Struct?
+public enum CoreError: Error {
```
scannillo:
```suggestion
public enum NetworkingError: Error {
```

Rename suggestion. I think `Core` is too vague, as these errors are specific to Networking.

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
```diff
@@ -0,0 +1,78 @@
+import XCTest
+@testable import PaymentsCore
+
+class APIClientTests: XCTestCase {
```
scannillo:
```suggestion
class APIClient_Tests: XCTestCase {
```

Also suggestion to rename this file to the same. It matches the other modules test files and I think will be easier to read/sift through as tests grow.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+    public var urlSession: URLSession
+    public var environment: Environment
+
+    public init(urlSession: URLSession = .shared, environment: Environment) {
+        self.urlSession = urlSession
+        self.environment = environment
+    }
+
+    public func fetch<T: APIRequest>(
+        endpoint: T,
```
sshropshire, line 15: I do see how using generics here allows for strongly typed requests. On the other hand, I wonder if the same thing could be achieved with a `protocol` instead?

sshropshire: I feel like a more loosely typed `APIResponse` object may offer more flexibility to the caller in this case rather than immediately returning a strongly typed result.

UnitTests/PaymentsCoreTests/NetworkingTests/RequestTests/AccessTokenRequestTests.swift
```diff
@@ -0,0 +1,25 @@
...
+class AccessTokenRequestTests: XCTestCase {
+    func testAccessTokenRequest_toURLRequest_vendsCorrectURLString() throws {
+        let expectedUrl = "https://api.sandbox.paypal.com/v1/oauth2/token"
+        let environment = Environment.sandbox
+
+        let testClientId = "testClientId"
+        let accessTokenRequest = AccessTokenRequest(clientID: testClientId)
+
+        let urlRequest = try XCTUnwrap(accessTokenRequest.toURLRequest(environment: environment))
+
```
scannillo:
```suggestion
```

UnitTests/PaymentsCoreTests/NetworkingTests/RequestTests/AccessTokenRequestTests.swift
```diff
@@ -0,0 +1,25 @@
...
+        
+        XCTAssert(urlRequest.url?.absoluteString == expectedUrl)
```
scannillo:
```suggestion
        XCTAssertEqual(urlRequest.url?.absoluteString, expectedUrl)
```
We should use this when we can instead

UnitTests/PaymentsCoreTests/NetworkingTests/RequestTests/AccessTokenRequestTests.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Jones, Jon on 9/20/21.
+//
+
+import Foundation
+@testable import PaymentsCore
+import XCTest
+
+class AccessTokenRequestTests: XCTestCase {
+    func testAccessTokenRequest_toURLRequest_vendsCorrectURLString() throws {
```
scannillo: I think we want `<testSUT_condition_result>` for naming. So for this something like:

`testToURLRequest_withValidURLComponents_vendsCorrectURLString`.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+        completion: @escaping (Result<T.ResponseType, CoreError>, CorrelationID?) -> Void
+    ) {
+        guard let request = endpoint.toURLRequest(environment: environment) else {
+            completion(.failure(.noURLRequest), nil)
+            return
+        }
```
scannillo, line 21: This isn't tested

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+            let finish: (Result<T.ResponseType, CoreError>) -> Void = { result in
+                /// For discussion:
+                /// When a network request to a PayPal API fails, we have some associated error information in the body data of the response.
+                /// This error data doesn't have the same format, but we should discuss and plan for how we want to surface this to the merchant
+                completion(result, correlationID)
+            }
+
+            if let error = error {
+                finish(.failure(.networkingError(error)))
+            }
```
scannillo: Needs a test

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Our code should do its best to parse errors from the various types of PayPal responses, even if they aren't uniform. And then of course fail gracefully if we aren't able to parse the error appropriately. Probably will require investigation on our end to get test cards to trigger error cases / invalid API keys/ etc. In my experience, we need to do this work ourselves.

UnitTests/PaymentsCoreTests/NetworkingTests/NetworkingMocks/MockRequestResponse.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+
```
scannillo: Leaving a comment here to simplify this class and move some of this setup logic into the tests themselves, as discussed in previous comment.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,65 @@
...
+        endpoint: T,
+        completion: @escaping (Result<T.ResponseType, CoreError>, CorrelationID?) -> Void
+    ) {
+        guard let request = endpoint.toURLRequest(environment: environment) else {
+            completion(.failure(.noURLRequest), nil)
+            return
+        }
+
+        let task = urlSession.dataTask(with: request) { data, response, error in
+            guard let response = response as? HTTPURLResponse else { return }
```
scannillo: Error case not handled for nil response

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: `APIRequest` here is a generic protocol (it has an `associatedtype` constraint), so it can't be referenced as a non-generic function parameter - I think this works for us for the time being since we will know the response type of all the requests we make from our SDK (since the only ones using this `APIClient` would be the other modules in the SDK)

re more loosely typed response - we can definitely add an additional `fetch` function that vends a JSON object instead of a typed result. That would be easy enough to include, but didn't see an immediate reason to add it just yet, so left it off. figured we could add that when we would want to use it instead of now though

sshropshire: Got it yeah. We may be able to emulate this Kotlin but it may not be an exact one-to-one correlation. The two languages are similar, but they do diverge at times. We also don't have the standard lib `Result` type that Swift has.

I'm wondering if it would be beneficial to adopt a style that scales across both platforms or if it's best to lean into the strengths of each language when we see fit?

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Gotcha - thanks for the resources/ comments. Makes sense to me!

UnitTests/PaymentsCoreTests/APIClientTests.swift
scannillo: I agree - we'll likely want to provide additional context/info and should set up for that 👍

Sources/PaymentsCore/Networking/APIClient.swift
sshropshire: I'm doing some due diligence research here. This is a good breakdown of the [Result](https://www.swiftbysundell.com/articles/the-power-of-result-types-in-swift/#:~:text=Using%20a%20Result%20type%20can,while%20retaining%20full%20type%20safety.) type in Swift for the uninitiated. It appears there is a similar [Result](https://medium.com/@jcamilorada/arrow-try-is-dead-long-live-kotlin-result-5b086892a71e) type in Kotlin!

Sources/PaymentsCore/Networking/Enums/CoreError.swift
jonathajones: All good if we punt on this decision for now and handle as a follow up work item?

UnitTests/PaymentsCoreTests/APIClientTests.swift
jonathajones: We simplified the process for mocking requests - should be fairly easy to follow now. To create a mock response for a request:
```swift
        let mockRequest = MockRequestResponse(
            request: AccessTokenRequest(clientID: "123"), // this is any APIRequest 
            statusCode: 500, // this is the status code of the URL response
            responseString: nil, // this is the response data to vend (this should be a json string)
            error: serverError,  // this is the error object to vend
        )
```

This can be seen in the implementation of all our APIClient tests, and will prevent us from needing to create any special infrastructure for different request types while keeping the response easily discernible from the callsite in a test

Sources/PaymentsCore/Networking/Protocols/APIRequest.swift
jonathajones: naming is hard! 😆 open to alternatives, but it feels okay to me - it feels like it defines the contract for what constitutes an API request well

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: handled!

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: Nice! My opinion is that this is an area that we don't strictly tight alignment between platforms, in case this exact sort of API client doesn't make as much sense on Android. If looser typing is a better fit on Android, then I see no problem with deviation.

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: updated!

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,75 @@
+import Foundation
+
+public class APIClient {
```
jaxdesmarais: Would it be worth making this a final class?

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,75 @@
...
+                // TODO:
+                // Add networking error cases (ie more descriptive networking errors / handle 400 responses, 500 errors, etc
+                finish(.failure(.unknown))
+            }
+        }
+
+        task.resume()
+    }
+
+    public func parseDataObject<T: APIRequest>(_ data: Data?, type: T.Type) throws -> T.ResponseType {
```
jaxdesmarais: Does this need to be public since it's just used in this scope? Or will it also be used by the other modules eventually?

Tests/NetworkingMocks/Mocks/BasicRequestMocks.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  BasicRequestMocks.swift
+//  BasicRequestMocks
+//
+//  Created by Jones, Jon on 9/22/21.
+//
+
```
jaxdesmarais: Remove 😄

Tests/NetworkingMocks/Mocks/MockRequestResponse.swift
```diff
@@ -0,0 +1,66 @@
+import Foundation
+import PaymentsCore
+
+class MockRequestResponse<Response: Codable>: APIRequest, MockResponse {
```
jaxdesmarais: I really like this for mocking our responses!  🚀

jaxdesmarais (approved):
Approved from my side! couple of minor questions/comments but more for my own knowledge than anything else.

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: nope! changed to internal

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: done!

sshropshire (approved):
🌊 &nbsp; I think this is structurally sound. We made a lot of similar decisions on Android. We plan on naturally evolving the networking portion as we implement more features as well. This is awesome.

Sources/PaymentsCore/Networking/Enums/NetworkingError.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+// TODO: Refine how we want to do error handling
+/// For further reading, see https://developer.apple.com/documentation/swift/error
+/// For Discussion:
+/// Error conforming enum vs Error conforming Struct?
+public enum NetworkingError: Error {
+    case networkingError(Error)
```
scannillo: This name isn't clear, since it's the same as the class. What about `connectionIssue`

Sources/PaymentsCore/Networking/Enums/NetworkingError.swift
```diff
@@ -0,0 +1,16 @@
...
+    case decodingError(Error)
```
scannillo: Nitpick - `parsing` over `decoding` is more clear to me

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -0,0 +1,75 @@
...
+                completion(result, correlationID)
+            }
+
+            if let error = error {
+                finish(.failure(.networkingError(error)))
+                return
+            }
+
+            guard let response = response as? HTTPURLResponse else {
+                finish(.failure(.badURLResponse))
```
scannillo:
```suggestion
                finish(.failure(.invalidURLResponse))
```
nitpick

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
```diff
@@ -0,0 +1,189 @@
+import XCTest
+@testable import PaymentsCore
+
+class APIClient_Tests: XCTestCase {
```
scannillo, line 4: This test file looks awesome! Super readable and thorough 👍

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/RequestTests/AccessTokenRequest_Tests.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  APIRequestTests.swift
+//  APIRequestTests
+//
+//  Created by Jones, Jon on 9/20/21.
+//
+
```
scannillo:
```suggestion
```

Tests/NetworkingMocks/URLProtocolMock.swift
```diff
@@ -0,0 +1,76 @@
...
+                didFailWithError: Errors.noResponse
+            )
+            client.urlProtocolDidFinishLoading(self)
+            return
+        }
+        let requestResponse = Self.requestResponses.remove(at: firstIndex)
+        resolveRequest(with: requestResponse, client: client)
+    }
+
+    override func stopLoading() {
```
scannillo, line 49: Do we need to override all of these functions? If we can remove the unnecessary ones we should

scannillo:
This is nitpicky, but the naming of these is hard to follow. They all sound sort of the same. Can we name these in a way to make their responsibility more clear? Also totally cool if we wanna punt on it / leave a TODO!

Also, we go back and forth between `<ClassName>Mock` and `Mock<ClassName>`, let's pick one.

Screen Shot 2021-09-27 at 10 40 02 AM

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
```diff
@@ -0,0 +1,189 @@
...
+            guard case .success = result else {
+                XCTFail("Expected successful empty response")
+                return
+            }
+            expect.fulfill()
+        }
+        waitForExpectations(timeout: 1)
+    }
+
+    func testFetch_withError_failsSuccessfully() {
```
scannillo: Naming should follow pattern set by other tests and be more specific as to type of error

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
```diff
@@ -0,0 +1,189 @@
...
+        let expect = expectation(description: "Get empty response type for mock request")
+        let emptyRequest = MockEmptyRequestResponse()
+
+        URLProtocolMock.requestResponses.append(emptyRequest)
+
+        apiClient.fetch(endpoint: emptyRequest) { result, _ in
+            guard case .success = result else {
+                XCTFail("Expected successful empty response")
+                return
+            }
```
scannillo, line 113: This test should have a positive assertion on the result instead

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
```diff
@@ -0,0 +1,189 @@
...
+        // Mock request whose API object does not vend a URLRequest
+        let noURLRequest = MockNoURLRequest()
+
+        URLProtocolMock.requestResponses.append(noURLRequest)
+
+        apiClient.fetch(endpoint: noURLRequest) { result, _ in
+            guard case .failure(.noURLRequest) = result else {
+                XCTFail("Expected bad URLResponse error")
+                return
+            }
```
scannillo, line 173: These tests should assert that the error type is equal to  `.noURLRequest`. This gives the test less room to potentially give a false positive.

Tests/NetworkingMocks/URLProtocolMock.swift
minhthenguyen: we need this as it's required for URLProtocol implementation :smile:

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
```diff
@@ -0,0 +1,189 @@
...
+class APIClient_Tests: XCTestCase {
+
+    lazy var apiClient: APIClient = {
+        APIClient(
+            urlSession: URLSession(urlProtocol: URLProtocolMock.self),
+            environment: .sandbox
+        )
+    }()
+
+    func testFetch_withAccessTokenSuccessMockResponse_returnsValidAccessToken() {
```
scannillo: Another nitpicky naming comment. We can make the naming format the same for all tests. Currently some say "with" some don't. Some say "returns" some say "vends". Long test names is OK and preferred.

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
minhthenguyen: to do positive assertion, the NetworkingError'd need to conform to `Equatable`.
That'd allow us to do `XCTEqual(theError, theExpectedErrorCase)` for enum without associated value.
But still, for cases with associated value like `connectionIssue(Error)` or `parsingError(Error)`, we'd still need to do this `guard case`.

IMO the guard case here already assert that the error here needs to be `.noURLRequest` otherwise it fails, so I don't think we'll ever give a false positive.
What do you think we leave this as is?

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
minhthenguyen: similar to below comment, since `.success` has an associated value we'd need to do `guard case` here.
We can do another positive assertion below this like `XCTAssertTrue(result is EmptyResponse)` but that'd always be true due to `associatedtype` of `EmptyRequestResponseMock` is already `EmptyResponse`

Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift
scannillo: The `XCTAssertEqual` should work fine on the Enum that conforms to Error. I think where it goes wonky is because we have these two cases `case networkingError(Error)` & `case decodingError(Error)` which take an arg. If these cases didn't take params, then we could do `XCTAssertEqual(error, .noURLRequest)` etc.

I think we should be able to assert directly on our error types. Our merchants tests & apps should be able to do this as well, without needed to do this workaround via the guard.

Also, I know we have a TODO for follow up later to re-implement our Error type. We could punt on this until then and include a note on it in the TODO comment?

minhthenguyen: Without associated value, It still needs to be `Equatable`

Screen Shot 2021-09-27 at 2 46 50 PM

But yes, I'll put a TODO comment there for this.

Tests/NetworkingMocks/URLProtocolMock.swift
scannillo: Got it! This article helped me understand this test setup a bit more https://forums.raywenderlich.com/t/chapter-8-init-deprecated-in-ios-13/102050/8 for those also interested

scannillo: Might help this part of the code to be more readable if we add comments for which methods were required to overrride?

Tests/UnitTests/PaymentsCoreTests/PaymentsCore_Tests.swift
```diff
@@ -1,7 +1,7 @@
 import XCTest
 @testable import PaymentsCore
 
-final class PayPalTests: XCTestCase {
+final class PaymentsCore_Tests: XCTestCase {
```
scannillo: This file can be deleted

scannillo (approved):
🚀 This is awesome, thank you for addressing all of my feedback! This lays an awesome foundation.

Tests/NetworkingMocks/URLProtocolMock.swift
minhthenguyen: these are all the required methods :smile: 
I can add some documentation on what each does to make it clearer

PR #11, opened 2021-09-20
Demo App Networking

scannillo:
### Reason for changes

We want our demo app to mimic a merchant integration, so it should connect to a "sample" merchant server.

### Summary of changes

- Add Networking layer to connect with sample merchant server
    - Add `createOrderID` func
    - Add `processOrderID` func (supports authorize/capture intents) - will work once we add approve order/card module
- Add basic models mirroring responses from sample server

### Next Steps

In a follow up PR we will attach the DemoSettings to some toggles/settings in the UI of the Demo app. We will also add UI for a basic card & PayPal checkout feature.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo @jaxdesmarais


Demo/Demo/ViewController.swift
```diff
@@ -3,8 +3,29 @@ import Card
 import PayPal
 
 class ViewController: UIViewController {
```
scannillo, line 5: The linter is complaining here for `Vertical Whitespace after Opening Braces Violation: Don't include vertical whitespace (empty line) after opening braces. (vertical_whitespace_opening_braces)`. I don't think we should follow this rule for class declarations, and a few other potential scenarios such as protocols, structs, etc.. Having a newline is common here & helps with organization.

Rather than make this PR conform to the linter, I think we can use this as a time to modify the linting rules for what we think is actually useful to lint on.

I'd rather remove this rule from the linter than need to add a `// swiftlint` comment on top of every class. Thoughts?

Demo/Demo/DemoMerchantAPI.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+
+final class DemoMerchantAPI {
+
+    static let sharedService = DemoMerchantAPI()
+
+    private init() {}
```
minhthenguyen: what do you think of making this function `init(environment: DemoSettings.Environment)` and add a `var environment`?
so we have an easy way to toggle between environments in our test app if we wanna :smile:

Demo/Demo/DemoMerchantAPI.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    func createOrder(orderParams: CreateOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
+        // TODO: get environment from settings in Demo app
+        guard let url = URL(string: DemoSettings.Environment.sandbox.baseURL + "/order?countryCode=US") else {
```
minhthenguyen: can we turn this into a function 
`func url(endpoint: String, environment: DemoSettings.Environment) -> URL?`
or `func url(endpoint: String) -> URL?` if we have an instance variable `environment`
so we don't have to combine base url string everytime

Demo/Demo/Models/ProcessOrderParams.swift
```diff
@@ -0,0 +1,5 @@
+struct ProcessOrderParams: Codable {
+    let orderId: String
+    let intent: String
+    let countryCode: String
```
minhthenguyen, line 4: question: I dont remember having this field in capture/authorize endpoints, is this required in our test server? 🤔

minhthenguyen (commented):
This looks great!!!!
I just have some very minor comments :smile:

Demo/Demo/ViewController.swift
minhthenguyen: +1 to removing that lint rule!!!

jonathajones: No strong opinion from me - I prefer consistency over the linter, so maybe we can decide that we will always have an empty line at the top of protocols / classes ? I'm open to either, just want to do the same thing

Demo/Demo/DemoSettings.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    enum Environment {
+        case sandbox, production
+
+        var baseURL: String {
+            switch self {
+            case .sandbox:
+                return "https://ppcp-sample-merchant-sand.herokuapp.com"
+            case .production:
+                return "https://ppcp-sample-merchant-prod.herokuapp.com"
```
jonathajones, line 22: note for later - maybe we update the heroku app to remove `ppcp` from the URL string, if we're going to take over that merchant app?

Demo/Demo/DemoMerchantAPI.swift
```diff
@@ -0,0 +1,78 @@
+import Foundation
+
+final class DemoMerchantAPI {
+
+    static let sharedService = DemoMerchantAPI()
+
+    private init() {}
+
+    func createOrder(orderParams: CreateOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
```
jonathajones: Could we add a line or so of documentation just describing the process here? and so it's clear that this is not part of the SDK flow

Demo/Demo/DemoMerchantAPI.swift
```diff
@@ -0,0 +1,78 @@
...
+                let order = try JSONDecoder().decode(Order.self, from: data)
+                completion(.success(order))
+            } catch {
+                completion(.failure(.dataParsingError))
+            }
+        }
+        .resume()
+    }
+
+    func processOrder(processOrderParams: ProcessOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
```
jonathajones: Same note here - would be nice to have a line of doc just describing what happens in this function, making clear this isn't part of the SDK flow

jonathajones (approved):
Really minor things from me - but looks good!

Demo/Demo/DemoMerchantAPI.swift
jonathajones: I think since it's a singleton init, it wouldn't matter much what initialization parameters we add to it. Guessing this will eventually read from `DemoSettings` to get the current environment

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -93,17 +103,30 @@
...
 				806F1E3826B85367007A60E6 /* AppDelegate.swift */,
 				806F1E3A26B85367007A60E6 /* SceneDelegate.swift */,
 				806F1E3C26B85367007A60E6 /* ViewController.swift */,
-				806F1E3E26B85367007A60E6 /* Main.storyboard */,
-				806F1E4126B85369007A60E6 /* Assets.xcassets */,
-				806F1E4326B85369007A60E6 /* LaunchScreen.storyboard */,
-				806F1E4626B85369007A60E6 /* Info.plist */,
+				80F33CE726F8DE29006811B1 /* DemoMerchantAPI.swift */,
+				80F33CF226F8EA50006811B1 /* DemoSettings.swift */,
+				BE1766B226F911A2007EF438 /* URLResponseError.swift */,
```
jonathajones: Maybe tuck the `DemoMerchantAPI` and `URLResponseError` in a networking directory, since we're gonna have a lot more UI types / complexity introduced as we flesh out the demo app

Demo/Demo/DemoSettings.swift
```diff
@@ -0,0 +1,21 @@
+// swiftlint:disable convenience_type
+
+import Foundation
+
+class DemoSettings {
+
```
jonathajones, line 4: To address some of Minh's feedback about making using the environment - can we make `DemoSettings` a singleton + add an `environment` property to it ? That'll let the `DemoMerchantAPI` just read from the current settings, rather than assume `sandbox`. Guessing we're already thinking in this direction, just saving it for a future PR?

Demo/Demo/DemoMerchantAPI.swift
minhthenguyen: ah you're right!!!! Also agree with your suggestion on making `DemoSettings` a singleton too, it'd be nice having one place to store all settings :smile:

Demo/Demo/DemoSettings.swift
scannillo: Good call. Yeh, we have a few things to do with that server, including to move it to the PayPal github. We'll add this rename to this list too!

Demo/Demo/Models/ProcessOrderParams.swift
scannillo: Currently, the sample merchant server requires a countryCode. However, it doesn't need it. It was a requirement of a previous project.

We made a note do update & re-deploy the sample server removing it!

Demo/Demo/ViewController.swift
jaxdesmarais: Removed this linter rule in PR #12. Agree with Jon on having this standardized to ensure it's consistent throughout the project!

Demo/Demo/DemoMerchantAPI.swift
scannillo: Hm, not sure I understand what benefit that will really give us. The URL will be different for different request types, creatOrder, processOrder, etc.

Demo/Demo/DemoSettings.swift
scannillo: Yeh - we were planning on adding this logic into the follow up PR. Cool if we punt on it for now?

Demo/Demo/ViewController.swift
scannillo: Awesome - thanks for taking care of that.

Demo/Demo/DemoSettings.swift
jonathajones: yep sure thing

Demo/Demo/DemoMerchantAPI.swift
scannillo: Happy to add documentation if folks are confused. However, I think it's pretty clear that `DemoMerchantAPI` which lives in the `Demo` app is not part of SDK logic. The separate XCode projects for Demo & PayPal help achieve this split too.

Demo/Demo/DemoMerchantAPI.swift
scannillo:
```suggestion
    /// Sends order to sample server for authorize/capture
    func processOrder(processOrderParams: ProcessOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
```

Does this clarify?

Demo/Demo/DemoMerchantAPI.swift
scannillo: Maybe instead at top of class, does this help?

```
/// API Client used to create and process orders on sample merchant server
final class DemoMerchantAPI: { ... }
```

Demo/Demo/DemoMerchantAPI.swift
jonathajones: Yep!

Demo/Demo/DemoMerchantAPI.swift
jonathajones: Yep that's great

Demo/Demo/DemoMerchantAPI.swift
minhthenguyen: I think if we have that function, here we'll do this:
```
guard let url = url(endpoint: "/order?countryCode=US") else ...
```
that way this function and `func processOrder()` don't need to care what the baseURL is and do the string combining anymore, func `url(endpoint: String)` will handle it for us.
Also if in the future we wanna change/rename `DemoSettings.Environment.sandbox.baseURL`, we'll just need to make one change in `func url(endpoint: String)` instead of having to change in multiple functions (`createOrder()`, `processOrder()`, etc.)

minhthenguyen: but also this is a very minor change request, totally cool if we don't do this :smile:

scannillo: Jax & I tried this out. It doesn't really make the code more readable or much more DRY, since it's one line of code that is pretty simple. We opted to leave as-is.

minhthenguyen: ah gotcha! Yep then everything looks good to me!!! :smile:

minhthenguyen (approved):

scannillo:
Linting will fail until #12 is merged. Might as well wait for it.

PR #12, opened 2021-09-20
Add Swiftlint Build Phase Script

jaxdesmarais:
### Reason for changes

This will allow us to see the linter warnings via Xcode when building instead of needing to wait for CI to finish to see them. If you've _not_ installed swiftlint yet on your machine, to run the linter the first time you will need to install via:
```
brew install swiftlint
```

Once you install swiftlint via homebrew, you will not need to run the above command again.

### Summary of changes

- Display swiftlint warnings inline
- Make us all a little happier

### Checklist

- @jaxdesmarais


jonathajones (approved):

scannillo:
Can we incorporate some linter-file updates into this PR? Here is a previous discussion in a different PR (https://github.com/paypal/iOS-SDK/pull/11/files#r712484455) regarding the `vertical_whitespace_opening_braces` rule.

DEVELOPMENT.md
```diff
@@ -4,9 +4,10 @@ This document outlines development practices that we follow while developing thi
...
-To run [SwiftLint](https://github.com/realm/SwiftLint) for all of the paths included in the `.swiftlint.yml`, run the following:
+Ensure that you have [SwiftLint](https://github.com/realm/SwiftLint) installed as we utilize it within our project.
 
+To install via [Homebrew](https://brew.sh/) run:
 ```
-bundle install
-bundle exec fastlane lint
+brew install swiftlint
 ```
+Our Xcode project has a `Run Phase` which integrates in `SwiftLint` so the only prerequisite is installing via `Homebrew`.
```
scannillo:
```suggestion
Our Xcode workspace has a `Run Phase` which integrates in `SwiftLint` so the only prerequisite is installing via `Homebrew`.
```

scannillo: nitpick

DEVELOPMENT.md
```diff
@@ -4,9 +4,10 @@ This document outlines development practices that we follow while developing thi
...
 ## SwiftLint
 
-To run [SwiftLint](https://github.com/realm/SwiftLint) for all of the paths included in the `.swiftlint.yml`, run the following:
+Ensure that you have [SwiftLint](https://github.com/realm/SwiftLint) installed as we utilize it within our project.
 
+To install via [Homebrew](https://brew.sh/) run:
 ```
-bundle install
-bundle exec fastlane lint
+brew install swiftlint
```
scannillo, line 11: Can we use bundler instead to do this instead of needing everyone to also need homebrew? Since our Gemfile already has `fastlane` I think that contains `swiftlint`.

jaxdesmarais: Swiftlint [unfortunately isn't available via bundler using the Gemfile](https://github.com/realm/SwiftLint/issues/2093). 🙁  It looks like we _could_ use [this gem called Danger](https://rubygems.org/gems/danger-swiftlint/versions/0.26.0) if we want to stick to using bundler for all of these, but I'm not sure what other updates would be needed to use that gem instead.

jaxdesmarais: Swiftlint [unfortunately isn't available via bundler using the Gemfile](https://github.com/realm/SwiftLint/issues/2093). 🙁  It looks like we _could_ use [this gem called Danger](https://rubygems.org/gems/danger-swiftlint/versions/0.26.0) if we want to stick to using bundler for all of these, but I'm not sure what other updates would be needed to use that gem instead.

jaxdesmarais: Actually, looking into [Danger a bit further](https://github.com/ashfurrow/danger-ruby-swiftlint) it looks like you still need to do the following:

> SwiftLint also needs to be installed before you run Danger, which you can do via Homebrew or a Brewfile. 

So not sure that's really a viable solution either.

scannillo:
Hi sorry I accidentally merged main into your branch @jaxdesmarais - meant to do it in another. Reverted.

DEVELOPMENT.md
scannillo: Darn! Well thanks for investigating

scannillo:
Other lint rules I think are worth chatting more about:

- Line length
  - Is it too short? Jax & I felt like it had room to be expanded a little and that it's a bit a struggle to fit everything ATM.
- `Unused Optional Binding Violation`
  - This made us convert `let _ =` syntax to ` != nil` syntax. Is this useful?

.swiftlint.yml
```diff
@@ -53,7 +53,6 @@ opt_in_rules:
   - unneeded_parentheses_in_closure_argument
   - unused_import
   - vertical_whitespace_closing_braces
```
scannillo, line 55: I think we can remove this one too. So then we could do:

```
class MyClass {

    var property1
    var property2
    var property3

    func doStuff() { 
       ...     
    }
   
    fund doMoreStuff() { 
        ... 
    }

}
```

jonathajones: I say if we remove this, then we also decide that we're going to open every class with an empty space. Don't necessarily need a rule for that though. I'd pref to lose the empty space between the last function and the last brace. But i agree that having a space between the type definition and the first member can help with readability

jonathajones:
> * Is it too short? Jax & I felt like it had room to be expanded a little and that it's a bit a struggle to fit everything ATM.

Maybe we can bump to 160 ? I'm fine increasing it, just don't want to have to scroll in PR review to fit all the text
 
> * `Unused Optional Binding Violation`
>   
>   * This made us convert `let _ =` syntax to ` != nil` syntax. Is this useful?

i find `if statusCode != nil` to be clearer than `if let _ = statusCode` - to me it's easier to read and more immediately apparent what the condition is

.swiftlint.yml
jonathajones: Happy to cave to other opinions on this!

jaxdesmarais: Yeah, I think I'm a fan of getting rid of `vertical_whitespace_opening_braces` but keeping `vertical_whitespace_closing_braces`? I think for me not having a space between the last func and closing brace is a bit easier to read? Basically same as what @jonathajones said. 😄  But also happy to be overruled if anyone feels strongly about it.

jaxdesmarais:
> Maybe we can bump to 160 ? I'm fine increasing it, just don't want to have to scroll in PR review to fit all the text

After doing some testing, it looks like 160 is basically to the end of the screen on a 16 inch macbook with the inspector pane on the right side closed. So maybe we bump it to 140 and go from there? We can always bump it back down to 120 if 140 feels like it's still too long. Or we can keep it at 120 for now and always bump it back up if that feels to restrictive?

scannillo (approved):

PR #13, opened 2021-09-23
Add CoreConfig + CardClient + approve card order request

minhthenguyen:
### Reason for changes

This PR adds `CoreConfig`, `CardClient` and a few models needed for card payment (`Card`, `CardExpiry`, etc.) to allow merchants to approve a card order:

```
// 1. Init config

let config = CoreConfig(clientID: <clientID>, environment: .sandbox)

// 2. Init CardClient

// 3. Approve a card order

// Create a card from buyer's inputs
let card = Card(
    number: <cardNumber>,
    expiry: CardExpiry{
        month: <expiryMonth>,
        year: <expiryYear>
    ),
    securityCode: <securityCode>
)

// Approve the order with above card
cardClient.approveOrder(orderID: <orderID>, card: card) { result in
    switch result {
        case .success(let orderData):
            print(orderData.orderID)
            print(orderData.status)
        case .failure(let error):
            print(error)
    }
}
```

### Summary of changes

- Added:
    - `CoreConfig`
    - `CardClient` (with `approveOrder` function)
    - `Card`, `CardExpiry`, `OrderData`, `OrderStatus`
    - `CoreError` (currently having `encodingData` for cases when we fail to encode data to make a network request)
    - `ConfirmPaymentSourceRequest`, `ConfirmPaymentSourceResponse`

### Checklist

- [ ] Added a changelog entry

### Authors

- @jonathajones 
- @minhthenguyen


Sources/PaymentsCore/Models/OrderData.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public struct OrderData {
+    let orderID: String
+    let status: OrderStatus?
```
minhthenguyen: currently `OrderData` only has orderID and status, but the actual network response for `/confirm-payment-source` also includes `payment-source`:
```
{
    "id": "6NV18671YP407052Y",
    "status": "APPROVED",
    "payment_source": {
        "card": {
            "last_digits": "7321",
            "brand": "VISA",
            "type": "CREDIT"
        }
    }
}
```
Should this be addressed and included in this PR?
Or leave as is for simplicity's sake and we'll consider adding this when more payment methods are implemented?

sshropshire: We came across this too. I believe we flattened the response into a Kotlin `data class` in a `CardResponse` object to decouple ourselves from the RESTful response's JSON schema.

Sources/PaymentsCore/Networking/Models/Requests/ConfirmPaymentSourceRequest.swift
```diff
@@ -0,0 +1,48 @@
+import Foundation
+
+/// Describes request to confirm a payment source (approve an order)
+public struct ConfirmPaymentSourceRequest: APIRequest {
```
sshropshire: qq: Should this be visible to the merchant?

minhthenguyen: good call, I agree this shouldn't be public!!!
But we had to since this request lives in `PaymentsCore`, but `Card` module (and others in the future) also need it :(

sshropshire: Ah yeah it'd be cool if there was a way to have it live in `Card` but sometimes the modules line up that way.

scannillo: Why does this need to live in Core? It should only be relevant to Card integrations.

Sources/Card/CardClient.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    /// Initialize a CardClient to process card payment
+    /// - Parameter config: The CoreConfig object
+    public init(config: CoreConfig) {
+        self.config = config
+        self.apiClient = APIClient(environment: config.environment)
+    }
+
+    /// For internal use for testing/mocking purpose
+    init(config: CoreConfig, apiClient: APIClient) {
```
scannillo, line 19:
```suggestion
    internal init(config: CoreConfig, apiClient: APIClient) {
```

Sources/Card/CardClient.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import PaymentsCore
+
+public class CardClient {
+
+    let apiClient: APIClient
+    let config: CoreConfig
```
scannillo:
```suggestion
    private let apiClient: APIClient
    private let config: CoreConfig
```

Sources/Card/Models/CardExpiry.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public struct CardExpiry: Codable {
```
scannillo:
```suggestion
public struct ExpirationDate: Codable {
```

We can remove the `Card` prefix since it lives in the Card module. Also `ExpirationDate` is more standard, at least in the docs I've seen for BT and PP.

Sources/Card/CardClient.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.apiClient = apiClient
+    }
+
+    /// Approve an order with a card, which validates buyer's card, and if valid, attaches the card as the payment source to the order.
+    /// After the order has been successfully approved, you will need to handle capturing/authorizing the order in your server.
+    /// - Parameters:
+    ///   - orderID: The ID of the order to be approved
+    ///   - card: The card to be charged for this order
+    ///   - completion: Completion handler for approveOrder, which contains data of the order if success, or an error if failure
+    public func approveOrder(orderID: String, card: Card, completion: @escaping (Result<OrderData, Error>) -> Void) {
```
scannillo, line 30: 👏  👏  👏

Sources/PaymentsCore/Networking/Models/Requests/ConfirmPaymentSourceRequest.swift
minhthenguyen: since we can do `confirm-payment-source` with payment sources other than card, I think this should live in `Core` so other payment methods we add in the future can use it :smile:

Sources/Card/CardClient.swift
minhthenguyen: I believe `internal` is by default

minhthenguyen:
@scannillo @jaxdesmarais Can you help me with the CI failures? I can't seem to figure out why `Build/CocoaPods` + `Build/SPM` failed? :(

Sources/Card/CardClient.swift
jaxdesmarais: Swift defaults to internal if you don't specify an access control level. I think typically I'd lean toward not specifying internal but if we feel like it's necessary and provides clarity, happy to be overruled.

Sources/Card/CardClient.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import PaymentsCore
+
+public class CardClient {
```
jaxdesmarais, line 6: General question - does this class need to be public or can it be gated to only the Card module?

Sources/Card/Models/ExpirationDate.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public struct ExpirationDate: Codable {
+
+    var expiryString: String
```
jaxdesmarais: `expiryString` > `expirationDateString`

Sources/Card/Models/ExpirationDate.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    /// Initialize a card expiry object
+    /// - Parameters:
+    ///   - month: The card expiration month (1 to 12)
+    ///   - year: The card expiration year in 4-digit format, `YYYY`
+    public init(month: Int, year: Int) {
+        let isSingleDigit = (1...9).contains(month)
+        let monthString = isSingleDigit ? "0\(month)" : "\(month)"
+        expiryString = "\(year)-\(monthString)"
```
jaxdesmarais: It [looks like Android](https://github.com/paypal/Android-SDK/blob/8b86a133ff5aa2cc3027da8d8ba8c70bdffd8e61/Demo/src/main/java/com/paypal/android/ui/card/validation/DateFormatter.kt) used `/` between the month and year and formats as `month/year`. Cool if we update to match?

Sources/Card/CardClient.swift
jaxdesmarais: Just saw Minh made the same comment 4 hours ago 🤦‍♂️  probably should have refreshed the PR

Sources/Card/CardClient.swift
minhthenguyen: ah this needs to be public so merchants can use it :smile:

Sources/Card/Models/ExpirationDate.swift
minhthenguyen: I think that was for their demo app. In here we'd need to format it to `YYYY-MM` as expected by our server

Sources/Card/CardClient.swift
sshropshire: Yeah I think for the most part `CardClient` and anything in its public interface ended up needing to be `public` for Android as well. Everything else we were able to make internal, with the exception of a few classes in our `Core` module.

Sources/Card/Models/ExpirationDate.swift
jaxdesmarais: Ah, you're right, my bad! That makes sense, sorry missed that that was a UI component.

Sources/Card/Models/ExpirationDate.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public struct ExpirationDate: Codable {
+
+    var expirationDateString: String
```
sshropshire: We could technically make this class internal and make `expirationDateString` available to merchants via a single property on `Card`. This would improve the onboarding experience a good bit since they'll probably make this assignment using a`TextField`'s `text` value in most cases.

Sources/PaymentsCore/Models/OrderData.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+public struct OrderData {
```
sshropshire, line 5: This is something Android can align on. Having a more generic `OrderData` class does make sense. On the other hand, there is some additional information from the `confirm-payment-source` endpoint response like `lastDigits` and `cardBrand` do we feel like the merchant will need this info?

Sources/PaymentsCore/Networking/Models/Requests/ConfirmPaymentSourceRequest.swift
scannillo: 🤔  At this point, it is only being used for card checkouts - so it should live in the Card Module. We want to avoid the PaymentsCore being a catch-all for everything and utilize / lean into the modularity we have. If we come across other payment methods needing `confirm-payment-source` we can cross that bridge later. How's that sound?

Sources/Card/Models/Card.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+
+public struct Card: Codable {
+
+    enum CodingKeys: String, CodingKey {
+        case number
+        case expirationDate = "expiry"
+        case securityCode = "security_code"
```
scannillo: Do we need to do all of this? Can we just do `JSONDecoder.KeyDecodingStrategy.convertFromSnakeCase`? I think we can assume things will be snake case from `orders/v2` for now. If that changes, we can look at workarounds if needed.

This would simplify our enum to then just be like this. We also get the init by default.

```
struct Card: Codable {
    case number
    case expirationDate
    case securityCode
}
```

It would also simplify things for PaymentsCore.

Sources/PaymentsCore/Models/OrderData.swift
scannillo: I think after some discussions with David we landed on a parent class living in Core, and then subclass result types living in each payment module. `CardResult`, `PayPalResult` etch to contain the more specific info.

Not sure about `OrderData` naming

Sources/PaymentsCore/Networking/Models/Responses/ConfirmPaymentSourceResponse.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+public struct ConfirmPaymentSourceResponse: Codable {
```
scannillo: Can move into Card module and make private

Tests/UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -0,0 +1,122 @@
...
+    lazy var cardClient: CardClient = {
+        let config = CoreConfig(clientID: "", environment: .sandbox)
+        let apiClient = APIClient(
+            urlSession: URLSession(urlProtocol: URLProtocolMock.self),
+            environment: .sandbox
+        )
+        return CardClient(config: config, apiClient: apiClient)
+    }()
+
+    func testApproveCardOrder_withCardSuccessMockResponse_returnsValidOrderData() throws {
```
scannillo, line 16: 👏

Tests/UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -0,0 +1,122 @@
...
+        """
+
+        let card = Card(number: "", expirationDate: ExpirationDate(month: 1, year: 2021), securityCode: "")
+        try setupConfirmCardPaymentSourceMock(card: card, mockResponse: mockFailureResponse, statusCode: 404)
+
+        cardClient.approveOrder(orderID: "", card: card) { result in
+            guard case .failure(NetworkingError.unknown) = result else {
+                XCTFail("Expect NetworkingError.unknown for status code 404")
+                return
+            }
```
scannillo, line 74: Should we add a TODO for adding error handling granularity for this case?

Tests/UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -0,0 +1,122 @@
...
+
+            expect.fulfill()
+        }
+
+        waitForExpectations(timeout: 1)
+    }
+
+    func testApproveCardOrder_withInvalidFormatMockResponse_returnsError() throws {
```
scannillo:
```suggestion
    func testApproveCardOrder_withInvalidFormatMockResponse_returnsParseError() throws {
```

Tests/UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,56 @@
...
+class APIClientTests: XCTestCase {
+
+    lazy var apiClient: APIClient = {
+        APIClient(
+            urlSession: URLSession(urlProtocol: URLProtocolMock.self),
+            environment: .sandbox
+        )
+    }()
+
+    func testAccessTokenMockRequest_Success() {
```
scannillo: Naming nitpick - the SUT here is `apiClient.fetch` method

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
```diff
@@ -0,0 +1,78 @@
+import XCTest
+@testable import PaymentsCore
+
+class APIClientTests: XCTestCase {
```
scannillo, line 4: We already have `APIClient_Tests` here (https://github.com/paypal/iOS-SDK/blob/main/Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift). I don't think we need this additional file.

Tests/UnitTests/PaymentsCoreTests/APIClientTests.swift
```diff
@@ -0,0 +1,56 @@
+import XCTest
+@testable import PaymentsCore
+@testable import Card
+
+class APIClientTests: XCTestCase {
```
scannillo: We already have `APIClient_Tests` here (https://github.com/paypal/iOS-SDK/blob/main/Tests/UnitTests/PaymentsCoreTests/NetworkingTests/APIClient_Tests.swift). I don't think we need this additional file.

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
scannillo: 🤔  We now have 3 files for APIClientTests at different nest levels / with different names. We should clear this up.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -24,55 +24,48 @@ public final class APIClient {
         }
 
         let task = urlSession.dataTask(with: request) { data, response, error in
-            let finish: (Result<T.ResponseType, NetworkingError>) -> Void = { result in
```
scannillo, line 24: Did some cleanup here to make this easier to read. 

_If this is too much/confusing for this PR, we can make a separate PR._

Sources/Card/Models/Card.swift
scannillo: Steven & I updated to simplify this.

Sources/PaymentsCore/Networking/Models/Responses/ConfirmPaymentSourceResponse.swift
scannillo: Moved into Card module.

Sources/Card/Models/Card.swift
minhthenguyen: thank you :smile: !!!!

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
scannillo: The testing structure got a little wonky after the networking PR. Also note, the file structure in XCode doesn't always accurately reflect the file structure in the repo, sometimes it gets out of sync and needs to be manually fixed.

Screen Shot 2021-10-05 at 9 55 05 AM

Sources/PaymentsCore/Networking/APIClient.swift
minhthenguyen: thanks Sammy!! I don't think it's too much.
Although probably I'd move `correlationID` to the top since no matter whether it's error or not, as long as we have it, we should pass it to merchants

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
scannillo: I think for this PR, let's just consolidate the APIClient_Tests into one place. In a follow up PR, let's fix this structure.

Sources/PaymentsCore/Networking/Models/Requests/ConfirmPaymentSourceRequest.swift
minhthenguyen: yea that works for me!!

minhthenguyen: actually thinking a bit about it, I think `confirm-payment-source` makes more sense living in Core. a few reasons:
- it's basically just an API wrapper around an Orders v2 endpoint (and we know all endpoints work for all payment methods), plus I believe handling network requests should be Core's responsibility.
- Why do we want to move it to Card when we know for sure it works for other payment methods and we'll have to move it back to Core in the future?

Please lemme know your thought :smile:

UnitTests/PaymentsCoreTests/NetworkingTests/APIClientTests.swift
minhthenguyen: Ope this is everything I see.

Screen Shot 2021-10-05 at 11 18 43 AM

Must've been my bad fixing merge conflicts in project file, I'll get this cleanup now :D

scannillo: You're looking in the XCode navigator ^

minhthenguyen: all `APIClientTests` file has been removed :smile:

Sources/Card/Models/Card.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+
+/// Initialize a card object
+/// - Parameters:
+///   - number: The card number
+///   - expiry: The card's expiration date
+///   - securityCode: The card's security code (CVV, CVC, CVN, CVE, or CID)
```
jaxdesmarais: Thoughts on adding something like "expected format" and the format to all of these? I assume there is a way that we need this data to be formatted to avoid errors.

scannillo: Great feedback. Will make this update!

Sources/Card/CardClient.swift
```diff
@@ -0,0 +1,52 @@
...
+    ///   - completion: Completion handler for approveOrder, which contains data of the order if success, or an error if failure
+    public func approveOrder(orderID: String, card: Card, completion: @escaping (Result<OrderData, Error>) -> Void) {
+        do {
+            let confirmPaymentRequest = try ConfirmPaymentSourceRequest(
+                card: card,
+                orderID: orderID,
+                clientID: config.clientID
+            )
+
+            apiClient.fetch(endpoint: confirmPaymentRequest) { result, _ in
```
scannillo, line 38: Do we need the `correlationID` in the `APIClient.fetch`? Here we are ignoring it.

scannillo (approved):

Sources/Card/CardClient.swift
minhthenguyen: yep we'll need it. 
I figure it will be somewhere in the `Result` we vend to merchants, either in `OrderData` or `Error` or both. And we have plan to discuss both `OrderData` and `Error` in both platforms so I left it unused here for now :smile:

jaxdesmarais (approved):
Thanks for addressing feedback! ✅ 🚀

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -0,0 +1,55 @@
...
+            .authorization: "Basic \(encodedClientID)"
+        ]
+    }
+
+    /// Creates a request to attach a payment source to a specific order.
+    /// In order to use this initializer, the `paymentSource` parameter has to
+    /// contain the entire dictionary as it exists underneath the `payment_source` key.
+    ///
+    /// For more information on the expected body structure, see PayPal's internal
+    /// [API reference](https://ppaas/api/3719176155270329#apiReference)
```
sshropshire:
```suggestion
```

sshropshire (approved):

PR #14, opened 2021-09-28
Demo App Basic Structure and Settings Menu

jaxdesmarais:
### Reason for changes

We want our demo app to have the abilities of all payment methods. This lays the structure for the demo app by adding the basic views needed as well as a settings menu to toggle between payment method types as they are added.

There are a lot of TODOs in here that will be covered in future stories

### Summary of changes

**Settings**
- Added `InAppSettingsKit` to handle settings changes and an easy interface for these changes
- Added `Settings.bundle` a required part of `InAppSettingsKit`
- Added environment, intent, and demoType to settings menu

**Views**
- `CardDemoViewController` - basic structure for our simple card view with TODO for future stories
- `PayPalDemoViewController` - nothing here yet besides a purple background - this will hold our PayPal view logic eventually
- `FeatureBaseViewController` - the base view for all of our views

**Other**
- Enums: Environment, Intent, DemoType
- `UIVIewController+Extensions` - convenience extension for our labels and buttons - can add other things as needed

### Checklist

@scannillo and @jaxdesmarais


Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        let alertController = UIAlertController(title: title, message: message, preferredStyle: .alert)
+        let okayAction = UIAlertAction(title: "Okay", style: .default) { _ in
+            self.navigationController?.popViewController(animated: true)
+        }
+
+        alertController.addAction(okayAction)
+        present(alertController, animated: true, completion: nil)
+    }
+}
```
scannillo: So for this base/parent class that all of the `<PaymentMethodType>Feature` classes will inherit from, I think we'll want it to do a few things, which will be relevant to all payment methods:
- Fetch an orderID
- Have some sort of status/view label that displays the state of the flow
- Have a button that allows us to `process` order (capture/authorize) after the flow completes

I'm not sure an alert is the most helpful here, since we will likely want to show multiple types of status updates such as "orderID fetched: \(...)", "cardApproval in process" etc. We also probably want a button nearby this status bar/label for "Process Order" since that is something we'll want to test for each payment flow.

Let me know if this makes sense!

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,91 @@
...
+    }()
+
+    let titleLabel: UILabel = {
+        let label = UILabel()
+        label.textAlignment = .center
+        label.numberOfLines = 0
+        label.lineBreakMode = .byWordWrapping
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
```
scannillo: This is neat! Cool way of consolidating all view customization logic for this property in one place.

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,91 @@
+import UIKit
+
+class FeatureBaseViewController: UIViewController {
+
+    let containerView: UIView = {
+        let view = UIView()
+        view.backgroundColor = .white
```
scannillo:
```suggestion
        view.backgroundColor = .systemBackground
```

We might as well use system colors when we can, since it will automatically support light & dark modes.

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,91 @@
...
+        super.viewDidLoad()
+        configureBottomView()
+    }
+
+    func configureBottomView() {
+        view.addSubview(containerView)
+        containerView.addSubview(titleLabel)
+
+        NSLayoutConstraint.activate([
+            containerView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
```
scannillo: Not sure why we need this containerView?

Demo/Demo/ViewController.swift
```diff
@@ -1,31 +1,40 @@
...
         super.viewDidLoad()
         view.backgroundColor = .systemBackground
 
-        // TODO: Update Demo UI with proper Card & PayPal checkout demos
-        fetchOrderID()
+        if #available(iOS 15.0, *) {
+            navigationController?.navigationBar.scrollEdgeAppearance = navigationController?.navigationBar.standardAppearance
+        }
+
+        let settingButton = UIBarButtonItem(title: "Settings", style: .plain, target: self, action: #selector(settingsTapped))
```
scannillo: Nitpick
```suggestion
        let settingsButton = UIBarButtonItem(title: "Settings", style: .plain, target: self, action: #selector(settingsTapped))
```

Demo/Demo/Features/FeatureBaseViewController.swift
jaxdesmarais: Typically, I've seen using a container for these types of labels to make adding things to them in the future easier as well as alignment since the text of the label will need to be dynamic. That said, we can certainly remove it and if we get to that point where it makes sense, re-add it.

Demo/Demo/Features/FeatureBaseViewController.swift
jaxdesmarais: I think that makes sense. Let me think on this a little more and get an idea implemented. I think initially, I liked the idea of some details (like success/failure) being in an alert to keep the bottom status bar from getting to busy, but I also understand what you're saying!

jaxdesmarais: Curious to hear your thoughts on the updated implementation of this @scannillo! Maybe we can sync up and chat about it if I'm still not quite following. 😄

Demo/Demo/Enums/DemoSettings.swift
```diff
@@ -0,0 +1,23 @@
+import UIKit
+
+enum DemoSettings {
```
scannillo:
```suggestion
struct DemoSettings {
```

I don't think enum makes the most sense here

Demo/Demo/Extensions/UIVIewController+Extensions.swift
```diff
@@ -0,0 +1,32 @@
+import UIKit
+
+extension UIViewController {
```
scannillo: Also I don't think this should live as an extension to `UIViewController`. It's a bit cleaner to have individual subclasses of UIButton, UITextField, etc such as `CustomLargeButton`, that customize these certain settings/properties.

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+            payButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 16),
+            payButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -16),
+            payButton.heightAnchor.constraint(equalToConstant: 40)
+        ])
+    }
+
+    @objc func didTapPayButton() {
+        // TODO: Pass card / order details into fetchOrderID
+        processOrder(paymentMethod: .card, state: .orderProcessing)
+        fetchOrderID(paymentMethod: .card)
```
scannillo: Fetch orderID should happen before they click the button, when the view loads. I think this PR has them flipped!

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,103 @@
...
+        switch state {
+        case .notStarted:
+            message = "Enter your \(paymentMethod) details to begin"
+        case .orderProcessing:
+            message = "Processing your \(paymentMethod) order"
+        case .orderCreated:
+            message = "Order ID: \(orderID ?? "")"
+        case .orderFailed:
+            message = "Your order has failed, please try again"
+        }
```
scannillo: Not sure if all these enum cases apply - since the first method above `fetchOrder` is what actually creates this order. This method should submit that order for authorizing or capturing, only needing an orderID string and intent.

scannillo: The state when we call this method could be like `Submitted order for authorize/capture ... `. This method should call `func processOrder(processOrderParams: ProcessOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void))` in our Demo API Client.

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,105 @@
...
+
+class CardDemoViewController: FeatureBaseViewController {
+    // TODO: Validate card fields are filled out
+    // TODO: Add proper spacing / styling / max characters to card fields
+    // TODO: Animate pay button to indicate loading
+    // TODO: Disable pay button until fields are filled out
+
+    private enum Constants {
+        static let stackViewSpacing: CGFloat = 16
+    }
```
scannillo: Do we need this enum, or can we just have a single property?

Demo/Demo/Enums/DemoSettings.swift
jaxdesmarais: Yeah so initially had this as a struct but was getting the following linter warning: `Types used for hosting only static members should be implemented as a caseless enum to avoid instantiation. (convenience_type)`. I did a bit more reading on caseless enums (I liked [this article](https://www.swiftbysundell.com/articles/powerful-ways-to-use-swift-enums/) that mentions it a bit down but there are some others) and I think it makes a lot of sense for the usecase of the settings as we use them today.

Demo/Demo/Features/CardDemoViewController.swift
jaxdesmarais: Ope, this was my misunderstanding! Will get  this updated.

Demo/Demo/Features/FeatureBaseViewController.swift
jaxdesmarais: Updated - removed the enum and switch altogether for now since the labels may be in flux for a bit. We can always add it back in and pass the state into the func in the future if we want.

Demo/Demo/Features/CardDemoViewController.swift
jaxdesmarais: We had a few other properties being repeated so added some additional properties to this enum

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    // TODO: Animate pay button to indicate loading
+    // TODO: Disable pay button until fields are filled out
+
+    private enum Constants {
+        static let stackViewSpacing: CGFloat = 16
+        static let layoutSpacing: CGFloat = 16
+        static let textFieldHeight: CGFloat = 40
+    }
+
+    let mainStackView: UIStackView = {
```
scannillo: `mainStackView` and `secondaryStackView` are a bit confusing - maybe better to rename them based on what they contain? Like `cardFormStackView` and `expirationCVVStackView` or something.

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,100 @@
+import UIKit
+
+class FeatureBaseViewController: UIViewController {
+
+    private enum Constants {
+        static let stackViewSpacing: CGFloat = 16
+        static let layoutSpacing: CGFloat = 16
+    }
+
+    lazy var stackView: UIStackView = {
```
scannillo: Nitpick - same naming comment as above

Demo/Demo/Features/FeatureBaseViewController.swift
```diff
@@ -0,0 +1,100 @@
...
+    }()
+
+    lazy var createOrderButton: CustomButton = {
+        let createOrderButton = CustomButton(title: "Create Order")
+        createOrderButton.addTarget(self, action: #selector(createOrderTapped), for: .touchUpInside)
+        createOrderButton.tintColor = .systemPink
+        return createOrderButton
+    }()
+
+    lazy var titleLabel: UILabel = {
```
scannillo: Nitpick - naming. Title sounds like it's the top of the screen. Maybe `bottomStatusLabel` or something

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -105,7 +136,9 @@
 			children = (
 				806F1E3826B85367007A60E6 /* AppDelegate.swift */,
 				806F1E4126B85369007A60E6 /* Assets.xcassets */,
-				80F33CF226F8EA50006811B1 /* DemoSettings.swift */,
+				BECD849E27036D95007CCAE4 /* Enums */,
+				BEE0EB6326FD10D90021AD42 /* Extensions */,
```
scannillo: `Extensions` --> `CustomViews` now, or something like that.

Demo/Demo/Extensions/CustomButton.swift
```diff
@@ -0,0 +1,21 @@
...
+
+class CustomButton: UIButton {
+    convenience init(title: String) {
+        self.init(type: .system)
+
+        setTitle(title, for: .normal)
+        translatesAutoresizingMaskIntoConstraints = false
+        titleLabel?.font = .preferredFont(forTextStyle: .body)
+        titleLabel?.numberOfLines = 0
+        titleLabel?.textAlignment = .center
```
minhthenguyen: Can this just be an extension on `UIButton` instead of a new class?

Demo/Demo/Extensions/CustomTextField.swift
```diff
@@ -0,0 +1,16 @@
...
+
+        translatesAutoresizingMaskIntoConstraints = false
+        borderStyle = .roundedRect
+        placeholder = placeholderText
+        text = defaultValue
+        clearButtonMode = .never
+        if let textDelegate = self as? UITextFieldDelegate {
+            delegate = textDelegate
+        }
+    }
```
minhthenguyen: Same comment in here on custom init in extension

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        payButton.tintColor = .white
+        return payButton
+    }()
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        view.backgroundColor = .systemBackground
+
+        setupUI()
```
minhthenguyen: very nit: I think we can either change `setupUI()` -> `addSubviews()`, or keep `setupUI()` but move `backgroundColor` in there. What do you think?

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    func setupConstraints() {
+        NSLayoutConstraint.activate([
+            mainStackView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            mainStackView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+
+            secondaryStackView.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+            secondaryStackView.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
+
+            cardNumberTextField.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+            cardNumberTextField.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
```
minhthenguyen: since `cardNumberTextField` is an arranged subview inside `mainStackView`, I think it'll be sized properly by the stackview so we may not need `leading/trailingAnchor` for this.

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        secondaryStackView.addArrangedSubview(cvvTextField)
+    }
+
+    func setupConstraints() {
+        NSLayoutConstraint.activate([
+            mainStackView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            mainStackView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+
+            secondaryStackView.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+            secondaryStackView.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
```
minhthenguyen: similar comment for `secondaryStackView` since it lives inside `mainStackView`

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+        mainStackView.addArrangedSubview(payButton)
+
+        secondaryStackView.addArrangedSubview(expirationTextField)
+        secondaryStackView.addArrangedSubview(cvvTextField)
+    }
+
+    func setupConstraints() {
+        NSLayoutConstraint.activate([
+            mainStackView.centerYAnchor.constraint(equalTo: view.centerYAnchor),
+            mainStackView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
```
minhthenguyen: instead of centerX, I think we can do leading and trailing :smile:

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+            cardNumberTextField.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+            cardNumberTextField.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
+            cardNumberTextField.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight),
+
+            expirationTextField.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight),
+            cvvTextField.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight),
+
+            payButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+            payButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
```
minhthenguyen: same comment here as with `secondaryStackView` and `cardNumberTextField`
Normally for stackview, we can set whatever layout we want for the outter stackview's constraints + insets, and the arranged subview will be sized accordingly :smile:

Demo/Demo/Features/CardDemoViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+
+class CardDemoViewController: FeatureBaseViewController {
+    // TODO: Validate card fields are filled out
+    // TODO: Add proper spacing / styling / max characters to card fields
+    // TODO: Animate pay button to indicate loading
+    // TODO: Disable pay button until fields are filled out
+
+    private enum Constants {
+        static let stackViewSpacing: CGFloat = 16
+        static let layoutSpacing: CGFloat = 16
```
minhthenguyen: can we remove these two since they exists in `FeatureBaseViewController`? 
And we can make the `Constants` in `FeatureBaseViewController` `internal` so subclasses can use them.

Demo/Demo/Features/FeatureBaseViewController.swift
scannillo: Nope, this looks good now!

minhthenguyen (commented):
I just built the demo app and it looks awesome :rocket:
Just some very minor comments

Demo/Demo/Features/CardDemoViewController.swift
minhthenguyen: instead we can have constraints for `mainStackView` https://github.com/paypal/iOS-SDK/pull/14#discussion_r722453787 and it will be applied to all arranged subviews :smile:

Demo/Demo/Extensions/CustomButton.swift
scannillo: Extensions should be used carefully. I think it's easy for them to become a catch-all. I think a question to ask before making an extension is "Does the entire class need to know". So the entire UIButton class doesn't really need to know about our custom settings, so in that case a subclass makes more sense.

Also another use-case for extensions is to add functionality to a class, which this doesn't really fall into that category. Does that make sense?

jaxdesmarais: My only hesitancy with making this an extension is if we ever want to have a button or text field that doesn't conform to these things. Now we have the flexibility to use custom buttons / text fields that looks the same while still being able to create a standard button or text field if needed. These were initially functions on an extension of UIViewController but were moved to be a subclass based on @scannillo suggestion. Sammy curious to hear your thoughts!

Demo/Demo/Extensions/CustomTextField.swift
jaxdesmarais: See https://github.com/paypal/iOS-SDK/pull/14#discussion_r722604924

Demo/Demo/Features/CardDemoViewController.swift
jaxdesmarais: I think changing to `addSubviews` makes sense - I'll make this update

Demo/Demo/Extensions/CustomButton.swift
minhthenguyen: Imo it's just shorter, no subclass needed, and if we don't use the new `init` in the extension then it's just gonna be a normal usual UIButton, so no harm risked :smile:
but it's just a minor suggestion, not a blocker, I'll approve once the other comments are addressed !!!! 🚀

minhthenguyen (approved):
this looks good to me!!! 🥳

Demo/Demo/Extensions/CustomButton.swift
scannillo: We shouldn't use an extension just because it is shorter. Extending UIViewController to expose custom buttons and views doesn't make much sense to me in terms of organization of logic. How often do you access buttons or UI elements via functions on UIViewController? Does that make sense? It doesn't feel like a native pattern. What if we want this custom button on a `UIView`, not on `ViewController`? We can't - the pattern just feels odd.

Also, it is common practice to create custom classes for custom views / what you see in tutorials often.

This is a demo app, so the stakes are super low. But, just in general thought it might help to explain where I am coming from.

We can now instantiate a `CustomButton` type in the same way we would instantiate a regular `UIButton`.

scannillo: Let me know if this makes sense / helps at all!

scannillo (approved):
🤘🏼

Demo/Demo/Extensions/CustomButton.swift
minhthenguyen: What you said makes 100% sense :smile: 

I just meant extending `UIButton`, not `UIViewController`, something like this:
```
extension UIButton {
    convenience init(title: String) {
        self.init(type: .system)
        ...
    }
}
```

But as said I think either extension or subclassing approach works fine :smile:
If we meant extending `UIViewController` for this then I'd also say no :smile:

PR #16, opened 2021-10-07
Simplify UnitTesting setup for APIClient

scannillo:
### Reason for changes 

Previously, our testing setup for `APIClient` was a bit hard to understand and sift through, at least to me. We had several files to mock out `URLSession` as well as several different `MockRequest` and `Response` types that were hard to differentiate.

The goal of this PR is to simplify our testing setup for `APIClient` while increasing test readability.

### Summary of changes 

- Create a `URLSessionProtocol`
  - Used to abstract `URLSession`'s `dataTask.resume()`
  - This makes our mocking of `URLSession` a lot easier
- Use dependency injection via an internal initializer in `APIClient` to inject a `MockURLSession` using the protocol above
- Remove `ResponseMock`, `RequestResponseMocks`, and `BasicRequestMocks` files in favor of single `FakeRequests`
- Update `APIClient_Test` and `CardClient_Test`
  - Use new `MockURLSession` and streamlined `FakeRequest` types
  - Rename & refactor tests to be a little more clear
- Remove `EmptyResponse`
  - Currently none of our API requests require this.
  - For now, let's remove it, adding it in if we need later on.

### Next Steps 

- We need to be able to assert on Error types. The current `guard case .success` style of testing doesn't give us helpful details if a test fails. 
  - Where I could, I update to use the `switch` within our tests.
  - Also, merchants will need this ability as well.
- We have several TODOs for more granular error types/ handling.

### Feedback 

Let me know what you guys think of this!

### Authors 
@scannillo


Sources/Card/CardClient.swift
```diff
@@ -46,6 +46,11 @@ public class CardClient {
                 }
             }
         } catch {
+            // TODO: There is no way to test this scenario
+            // since we cannot test the `try` of ConfirmPaymentSourceRequest
+            // failing.
+            // This deserves a re-write. Also not sure if the try
+            // pattern there is necessary at all.
```
minhthenguyen, line 53: I agree we should remove this `try catch`, we'd never fall into the `catch` here anyway I believe.
How about in `ConfirmPaymentSourceRequest` we do `body = try? jsonEncoder.encode(paymentSource)` and remove `throw`? :smile:

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -3,15 +3,24 @@ import Foundation
 public final class APIClient {
     public typealias CorrelationID = String
 
-    public var urlSession: URLSession
-    public var environment: Environment
+    private var urlSession: URLSessionProtocol
+    private var environment: Environment
```
minhthenguyen, line 7: what do you think of leaving `environment` public? 
we can also do `public private(set)` if we don't want merchant to re-set this after init :smile:

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -69,7 +69,16 @@ public final class APIClient {
...
+}
+
+protocol URLSessionProtocol {
+    func performRequest(with urlRequest: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void)
+}
 
+extension URLSession: URLSessionProtocol {
+    func performRequest(with urlRequest: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) {
+        let task = dataTask(with: urlRequest, completionHandler: completionHandler)
         task.resume()
```
minhthenguyen, line 82: niceeeee :rocket:

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -44,17 +53,8 @@ public final class APIClient {
             switch response.statusCode {
             case 200..<300:
                 do {
-                    // TODO: Get rid of this empty case, relevant tests, & files.
-                    if let emptyResponse = EmptyResponse() as? T.ResponseType {
-                        completion(.success(emptyResponse), correlationID)
-                        return
-                    } else {
```
minhthenguyen, line 51: the empty response was added to account for patch order in Orders v2, which returns an empty JSON body when succeeded
I agree we can re-add it when the time comes, just wanna explain a bit on why we had this in the first place :smile:

UnitTests/TestShared/MockURLSession.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+@testable import PaymentsCore
+
+class MockURLSession: URLSessionProtocol {
+    
+    var cannedError: Error?
+    var cannedURLResponse: URLResponse?
+    var cannedData: Data?
```
minhthenguyen: can we add `var cannedJsonResponse: String?` and let this be
```
private var cannedData: Data? { cannedJsonResponse.data(using: String.Encoding.utf8)! }
```
so we don't have to convert string -> data in every test? :smile:

minhthenguyen (commented):
all are very minor comments. 
Thanks for simplifying this, this looks good to me :smile:

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Gotcha - yeh I figured it came from something that was used in the Native XO SDK, originally.

UnitTests/TestShared/MockURLSession.swift
scannillo: I don't mind leaving as-is ...

```
let jsonResponse = "some-string"
mock.cannedData = jsonResponse.data(using: String.Encoding.utf8)!
```

in each test. I think setting the `cannedData` explicitly in the test makes things pretty clear.

Open to folk's input though if y'all find this confusing / feel strongly.

Sources/Card/CardClient.swift
scannillo: Yeh, maybe we can tackle that in a separate PR? I can leave this TODO in.

minhthenguyen: I wouldn't mind having it now rather than opening a one line PR that we can't get enough test coverage for :smile:

scannillo: I meant more along the lines that it isn't really in the scope of this PR - refactoring the CardClient.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: The `APIClient` is meant for merchant consumption, only the `CardClient` is. Kind of stinks we have to make it `public` so the other modules can access it. In ObjC, we were able to create `internal headers` and share them with modules (slightly hacky) but it allowed us to really lock down our public API and make sure it was only what we actually want merchants to use.

Not sure of a similar option for Swift. I think this is worth investigating.

But - basically the `APIClient` isn't meant for merchant consumption. Let's keep the `env` access level to private, and then only change if we need in the future. I think this a good practice too, defaulting to the strictest level of access first, then re-assessing as needs change.

minhthenguyen: ope my bad, for a moment I confused `APIClient` with `CoreConfig` 😆  
Yep this should be private, please disregard my comment!!!

sshropshire (approved):
🌊  clean!

Sources/PaymentsCore/Networking/APIClient.swift
minhthenguyen: and I agree we should investigate making this internal but shared between module! if that's possible that'd be awesome

UnitTests/TestShared/MockURLSession.swift
sshropshire: A `String` accessor would allow a [heredoc](https://stackoverflow.com/a/56877844) assignment on the response body to increase test readability. I think it's fair since HTTP is a plain-text protocol.

Sources/PaymentsCore/Networking/APIClient.swift
sshropshire: It'd be cool if we could infer `Environment` from an auth token and make it internal. Plaid does something similar.

UnitTests/TestShared/MockURLSession.swift
minhthenguyen: definitely not confusing or anything, just that I think passing in the json string is clear enough
and it'd make our tests a bit shorter without having to repeat the same string -> data conversion everytime (which doesn't help make the tests any clearer) :smile: 
but not a blocker

minhthenguyen (approved):

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: Since this already exists here, I'm not sure what the value of removing it right now is. I don't see it as being tied to a specific network request that we're not making, but rather supporting any request that does not vend a response body. those endpoints exist and I think there's a decent chance we'll have to use them - so removing this kind of feels like creating more work for ourselves in the future if we do end up needing to hit one of those enpoints

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -4,20 +4,42 @@ import XCTest
...
+    
+    // MARK: - Test lifecycle
+    
+    override func setUp() {
+        super.setUp()
+        
+        mockURLSession = MockURLSession()
+        mockURLSession.cannedError = nil
+        mockURLSession.cannedURLResponse = nil
+        mockURLSession.cannedData = nil
```
jonathajones, line 31: optional properties will be nil by default

jonathajones: optional properties will be nil by default
```suggestion
```

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -73,107 +116,94 @@ class APIClient_Tests: XCTestCase {
...
 
         apiClient.fetch(endpoint: noURLRequest) { result, _ in
-            guard case .failure(.noURLRequest) = result else {
-                XCTFail("Expected bad URLResponse error")
-                return
+            switch result {
+            case .success(_):
+                XCTFail()
+            case .failure(let error):
+                XCTAssertNotNil(error)
```
jonathajones, line 201: we should assert this as a `noURLRequest` error instead of asserting non-nil

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -73,107 +116,94 @@ class APIClient_Tests: XCTestCase {
...
             expect.fulfill()
         }
         waitForExpectations(timeout: 1)
     }
 
-    func testFetch_withNoResponseMock_vendsNoReponseError() {
-        let expect = expectation(description: "Should receive bad URLResponse error")
-        let noReponseRequest = NoReponseRequestMock()
+    func testFetch_whenPayPalDebugHeader_returnsCorrelationID() {
+        let expect = expectation(description: "Should receive noURLRequest error")
```
jonathajones:
```suggestion
    func testFetch_whenPayPalDebugHeader_returnsCorrelationID() {
        let expect = expectation(description: "Should receive correlation ID")
```

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -73,107 +116,94 @@ class APIClient_Tests: XCTestCase {
...
         waitForExpectations(timeout: 1)
     }
 
-    func testFetch_withRequestExpectingEmptyResponse_vendsSuccessResult() {
-        let expect = expectation(description: "Get empty response type for mock request")
-        let emptyRequest = EmptyRequestResponseMock()
+    // TODO: Get more granual here. Also, do we want to move this check for
+    // non-success status code above data parsing in our source code?
+    func testFetch_whenBadStatusCode_returnsUnknownError() {
+        let expect = expectation(description: "Get mock response for access token request")
```
jonathajones:
```suggestion
    func testFetch_whenBadStatusCode_returnsUnknownError() {
        let expect = expectation(description: "Return unknown error for bad status code")
```

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -73,107 +116,94 @@ class APIClient_Tests: XCTestCase {
...
-        apiClient.fetch(endpoint: emptyRequest) { result, _ in
-            guard case .success = result else {
-                XCTFail("Expected successful empty response")
-                return
-            }
-            expect.fulfill()
-        }
-        waitForExpectations(timeout: 1)
-    }
+        mockURLSession.cannedData = jsonResponse.data(using: String.Encoding.utf8)!
```
jonathajones: we can avoid force unwrapping in tests is to make the func throwing and then use `XCTUnwrap` -
```swift
func testFetch_whenBadStatusCode_returnsUnknownError() throws {
...
mockURLSession.cannedData = try XCTUnwrap(jsonResponse.data(using: String.Encoding.utf8))
...
```
Not that this would ever realistically fail, but a crash on a force unwrap would abort the test suite from running rather than fail the test, so as we get a lot more tests / code to test, defaulting to this would be a good way to avoid unwrapping and minimize that risk

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -24,47 +42,72 @@ class APIClient_Tests: XCTestCase {
...
-    func testFetch_withAccessTokenFailureMockResponse_vendsUnknownError() {
+    func testFetch_whenNoURLResponse_returnsInvalidURLResponseError() {
         let expect = expectation(description: "Get mock response for access token request")
 
-        let mockFailureResponse: String = """
-        {
-            "error": "unsupported_grant_type",
-            "error_description": "unsupported grant_type"
+        mockURLSession.cannedURLResponse = nil
+
```
jonathajones: since we're using a shared `mockURLSession` var on this test class, we'll need to update all of its' state on every test to avoid accidentally reusing properties that aren't set

in this test `mockURLSession` will have the `cannedError` and `cannedData` from whatever test from this file was run immediately before this.

Maybe we recreate a new `apiClient` in each test case? since we're not initializing a new `URLSession` object, this wouldn't be expensive and would let us avoid inadvertently sharing state between tests

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -24,47 +42,72 @@ class APIClient_Tests: XCTestCase {
...
+        waitForExpectations(timeout: 1)
+    }
 
-        let mockRequestResponse = RequestResponseMock(
-            request: AccessTokenRequest(clientID: ""),
-            statusCode: 404,
-            responseString: mockFailureResponse
-        )
+    func testFetch_whenNoResponseData_returnsMissingDataError() {
+        let expect = expectation(description: "Get mock response for access token request")
```
jonathajones:
```suggestion
    func testFetch_whenNoResponseData_returnsMissingDataError() {
        let expect = expectation(description: "Return missing data error")
```

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -3,15 +3,24 @@ import Foundation
...
-    public var environment: Environment
+    private var urlSession: URLSessionProtocol
+    private var environment: Environment
 
-    private let decoder = JSONDecoder()
+    private var decoder: JSONDecoder {
+        let decoder = JSONDecoder()
+        decoder.keyDecodingStrategy = .convertFromSnakeCase
+        return decoder
+    }
```
jonathajones: Instead of being computed, this should be an instance var - 
```suggestion
    private var decoder: JSONDecoder = {
        let decoder = JSONDecoder()
        decoder.keyDecodingStrategy = .convertFromSnakeCase
        return decoder
    }()
```

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -69,7 +69,16 @@ public final class APIClient {
...
+protocol URLSessionProtocol {
+    func performRequest(with urlRequest: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void)
+}
 
+extension URLSession: URLSessionProtocol {
+    func performRequest(with urlRequest: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void) {
+        let task = dataTask(with: urlRequest, completionHandler: completionHandler)
         task.resume()
     }
 }
```
jonathajones, line 84: I think this protocol and the extension probably deserve their own file separate from the `APIClient` - what do you think ?

UnitTests/CardTests/CardClient_Tests.swift
scannillo: We're doing this in the `setUp` to wipe these settings out in between test runs. The alternative would be a new instance of `MockURLSession` per test.

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Ah - so in the `setUp` which gets run before each test, we're wiping out these settings. Is that clear? If not, yeh we could create new instances per test.

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Yeh, agreed! We have a big TODO for that for a future PR, I mention it a bit in the PR description. Current limitation to our testing setup is that our Error types don't conform to `Equtable` so we can't assert on them. We definitely need to fix this.

Also this syntax `guard case .failure(.unknown) = result else {` in the test suite needs to be updated alongside those future changes. When a test with that syntax fails, we get no useful info.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Good call - will update!

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Definitely, that is how all of these tests should be written.

Check out the convo I wrote in the end of the PR description / next steps. This is a big TODO for our SDK that we have listed in the source code. Our Errors don't conform to `Equatable` and we can't assert on them. We definitely need to update this.

Also this syntax ...
```
guard case .failure(.noResponseData) = result else {
    XCTFail()
   return
}
```

Needs to be changed in that future PR when we make that Error type update. The current setup here gives us no helpful info when a test fails.

I think that work should be a separate PR, though!

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Gotcha - yeh that's a great callout. There are a few other comments on this PR to move the string -> data conversion anyways, so I'll address the force unwrap in that slight refactor.

UnitTests/TestShared/MockURLSession.swift
scannillo: Sure - I refactored the string -> data conversion logic to now live in the MockURLSession. Let me know how that looks to you.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: No worries, Minh! And Steven, agreed. Ideally we can infer the environment. We currently do that in the BT SDK too. Merchant doesn't need to specify the env, and the SDK can just detect it from the auth credential (or even via a config request and we can remove CoreConfig). I think convo for that is still ongoing?

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: I thought about this - I think that this protocol & extension are solely for use in `APIClient` and that moving them into separate files would just make things more confusing. Their functionality should be super limited and honestly coupled to the `APIClient`, which is why I thought it was easier to read to leave it as one. Thoughts?

UnitTests/CardTests/CardClient_Tests.swift
jonathajones: oh yep gotcha!

Sources/PaymentsCore/Networking/APIClient.swift
sshropshire: It seems like `EmptyResponse` serves as a [Null object](https://en.wikipedia.org/wiki/Null_object_pattern) here. In the future could we return `nil` instead?

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
jonathajones: ah yeah gotcha, yep this makes sense! I confused the instance `setUp` func with the class `setUp` func. my bad!

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: I'm also just sort of confused by `EmptyResponse() as? T.ResponseType`. 

Since we're associating each Request type with a Response type [(here)](https://github.com/paypal/iOS-SDK/blob/55dc95000daf81da83959e667d9056fd1fb70fe8/Sources/PaymentsCore/Networking/Protocols/APIRequest.swift#L4), we already know if the associated Response type for a given APIRequest should be `EmptyResponse` or not. So, we shouldn't need to do a runtime check/ cast.

jonathajones: an empty data response cannot `decode` into anything - even an empty type. The cast works to see if the request expects an empty data response, and if it does, return that to the caller without trying to decode anything from the request (since that would fail and trigger the `catch` block). 

Maybe a better approach would be to implement a custom decoder on `EmptyResponse`, so that we don't have to include the cast here?

sshropshire: On Android we have an`HttpResponse` type that contains a `status` and an optional `body`.

This might be something to investigate in another PR. Technically, we could do the decoding one layer up in the client that's making the request. It wouldn't be as DRY but it would enable modularity across the codebase.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: I'm going to mark this comment "Resolved" since it isn't blocking this PR at all. We can discuss this again.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: > Maybe a better approach would be to implement a custom decoder on EmptyResponse, so that we don't have to include the cast here?

I like this idea, definitely something we should explore!

Another thing here, I think we should stick to [YAGNI](https://martinfowler.com/bliki/Yagni.html) as much possible. I haven't run into a scenario when supporting this use case is necessary and am not confident it's likely to happen. I think adding stuff we don't need at this point adds to tech debt / confusion / cruft. We can iterate when requirements evolve.

At this point, if something falls into that block (AccessTokenRequest or ConfirmPaymentRequest) something went wrong anyways.

Thoughts?

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Resolving this, since we have a TODO in the source code for it 👍

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: that works! can always move it out later if it needs to grow

PR #18, opened 2021-10-08
Update Card UI

jaxdesmarais:
### Reason for changes

Resolve a number of the TODOs for the card UI!

### Summary of changes

- Validate card fields are filled out.
- Add proper spacing / styling / max characters to card fields. Note: this includes some _very basic_ luhn and card validation. We can certainly add more to the card type enum.
- Animate pay button to indicate loading. Note: we will need to stop the animation when we add process card.
- Disable pay button until fields are filled out correctly

#### Visuals
Simulator Screen Recording - iPod touch (7th generation) - 2021-10-08 at 14 51 35

### Checklist

- @jaxdesmarais


Demo/Demo/Views/CustomButton.swift
```diff
@@ -1,9 +1,15 @@
 import UIKit
 
 class CustomButton: UIButton {
+    var titleHolder: String?
+    // swiftlint:disable:next implicitly_unwrapped_optional
```
scannillo: Should we turn off the linter for our Demo app, only have it run on the `PayPal.xcodeproj`?

Demo/Demo/Card/CardFormatter.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    private func getCardType(_ cardNumber: String) -> CardType {
+        if cardNumber.starts(with: "34") || cardNumber.starts(with: "37") {
+            return .americanExpress
+        } else if cardNumber.starts(with: "4") {
+            return .visa
+        } else {
+            return .unknown
+        }
+    }
```
scannillo: We can pull this logic into the `CardType` enum. Also might as well as the regexes for all 4 of the major card types while we're at it? (discover + mastercard)

Demo/Demo/Card/CardType.swift
```diff
@@ -0,0 +1,25 @@
+enum CardType {
+    case americanExpress, visa, unknown
+
+    var cardNumberIndices: [Int] {
```
scannillo: This variable name is a bit confusing - is this where the space goes between the digits?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -88,6 +90,25 @@ class CardDemoViewController: FeatureBaseViewController {
 
     @objc func didTapPayButton() {
         updateTitle("Processing order...")
-        // TODO: Call `processOrder`
+        payButton.startAnimating()
+        // TODO: Call `processOrder` and `payButton.stopAnimating()` once process order is complete
+        }
```
scannillo:
```suggestion
    }
```

Demo/Demo/Views/CustomButton.swift
```diff
@@ -18,4 +24,40 @@ class CustomButton: UIButton {
...
+        let activityIndicator = UIActivityIndicatorView()
+        activityIndicator.hidesWhenStopped = true
+        activityIndicator.color = UIColor.white
+        return activityIndicator
+    }
+
+    private func showSpinning() {
+        activityIndicator.translatesAutoresizingMaskIntoConstraints = false
+        self.addSubview(activityIndicator)
+        setupConstraints()
```
scannillo: These setup-type lines should probably live outside of the `showSpinning()`. Maybe in one of the inits?

Demo/Demo/Views/CustomButton.swift
```diff
@@ -1,9 +1,15 @@
...
 
 class CustomButton: UIButton {
+    var titleHolder: String?
+    // swiftlint:disable:next implicitly_unwrapped_optional
+    var activityIndicator: UIActivityIndicatorView!
+
     convenience init(title: String) {
         self.init(type: .system)
 
+        titleHolder = title
```
scannillo: A little confused what we need the `titleHolder` for / if we need it?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -88,6 +90,25 @@ class CardDemoViewController: FeatureBaseViewController {
...
+            let expirationDateCount = expirationTextField.text?.count,
+            let cvvCount = cvvTextField.text?.count,
+            cardNumberCount >= 18 && cardNumberCount <= 19,
+            expirationDateCount >= 7 && expirationDateCount <= 9,
+            cvvCount >= 3 && cvvCount <= 4
+        else {
+            payButton.isEnabled = false
+            return
+        }
+        payButton.isEnabled = true
```
scannillo: 👏  Great idea making the button only clickable with valid card input.

Demo/Demo/Views/CustomButton.swift
jaxdesmarais: I like the idea of running it on both just to keep style consistent across the SDK and demo. But I'll update this so that I don't need to force unwrap and can remove this linter comment. 😄

Demo/Demo/Card/CardType.swift
jaxdesmarais: Yeah, totally open to renaming this, but I couldn't think of a better name so just used the same as Android used.

scannillo: `spaceDelimiterIndeces`?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -88,6 +90,25 @@ class CardDemoViewController: FeatureBaseViewController {
 
     @objc func didTapPayButton() {
         updateTitle("Processing order...")
-        // TODO: Call `processOrder`
+        payButton.startAnimating()
+        // TODO: Call `processOrder` and `payButton.stopAnimating()` once process order is complete
+    }
+
+    @objc private func editingChanged() {
```
scannillo: Nitpick - naming of `editingChanged` is vague. What does this method represent?

Demo/Demo/Views/CustomButton.swift
```diff
@@ -18,4 +30,29 @@ class CustomButton: UIButton {
...
+    func startAnimating() {
+        self.setTitle("", for: .normal)
+
+        showSpinning()
+    }
+
+    func stopAnimating() {
+        self.setTitle(titleHolder, for: .normal)
+        activityIndicator.stopAnimating()
+    }
```
minhthenguyen: one concern I have is that we keep adding subview + setting constraints for the `activityIndicator` everytime the button is tapped.
What do you think of doing `addSubview()` and `setupConstraints()` in `init()`, then toggle either `alpha` or `isHidden` of `activityIndicator` in these functions? :smile:

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -88,6 +90,25 @@ class CardDemoViewController: FeatureBaseViewController {
...
+        cardNumberTextField.text = cardFormatter.formatCardNumber(cardNumberTextField.text ?? "")
+        expirationTextField.text = cardFormatter.formatExpirationDate(expirationTextField.text ?? "")
+
+        guard
+            let cardNumberCount = cardNumberTextField.text?.count,
+            let expirationDateCount = expirationTextField.text?.count,
+            let cvvCount = cvvTextField.text?.count,
+            cardNumberCount >= 18 && cardNumberCount <= 19,
```
minhthenguyen: should we do `18` here, or should it be `15`? 
I'm not sure what the required length for each card type is, but I do see you have `americanExpress` or `visa` which can be 15 or 16 :smile:

minhthenguyen: we can also make sure `cardNumberCount` here is `>= 4`, then we can use `CardType` to `getCardType()` then we can assert the exact length / or length range of the specific card type.

But this is just a suggestion, not a blocker, I'm fine either way :smile:

Demo/Demo/Card/CardFormatter.swift
```diff
@@ -0,0 +1,47 @@
...
+            return newCardNumber
+        }
+
+        var cardString: String = newCardNumber.replacingOccurrences(of: " ", with: "")
+
+        let cardType = CardType.unknown.getCardType(cardString)
+
+        if previousCardNumber.count == cardType.maxLength + cardType.cardNumberIndices.count {
+            return previousCardNumber
+        } else {
```
minhthenguyen: this is getting into the nitpicking/edge case domain, but this might not work if user copy/paste in a new string with this length (we'll skip formatting the new string in this case :smile:)
maybe we can remove this and just leave the `else` part. What do you think?

Demo/Demo/Card/CardFormatter.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+    func formatExpirationDate( _ newExpirationDate: String, previousExpirationDate: String = "") -> String? {
+        var currentExpirationDate = newExpirationDate
+
+        if currentExpirationDate.count == 4 {
+            currentExpirationDate = String(currentExpirationDate.prefix(2))
+            return currentExpirationDate
+        }
```
minhthenguyen: this is not very clear at least to me, can you explain what this is used for?

minhthenguyen (commented):
This looks great! I just have a few comments + clarification questions :smile:

Demo/Demo/Card/CardFormatter.swift
scannillo: That's a good call - I'm sure we'll need to copy paste card #s at some point!

Demo/Demo/Views/CustomButton.swift
jaxdesmarais: Yep, I think it makes sense to move add subview and setup constraints in the init. For the alpha, I think it makes sense to use the `start/stopAnimating` functions built into activity indicator as that will always give us the expected behavior, but that can be simplified once we move the other portion to the init.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: Since we format the card numbers to add spaces, we need to add the spaces into the count for total length. 😄   But you are right, this is incorrect since I added the additional card types. I'm going to refactor this to hopefully be a bit more clear.

For getting the CardType and setting max characters, this should be possible, though we will probably use `shouldChangeCharactersIn` in the textfield's delegate vs having it within this function. Let me do some more digging on this!

Demo/Demo/Card/CardFormatter.swift
jaxdesmarais: Sure! I think that makes sense!

Demo/Demo/Card/CardFormatter.swift
jaxdesmarais: Yep, so this is used when we delete the expiration date field after typing. If you comment this out, you will see that it only deletes up to the `/` or if you add a breakpoint here you will hit it when deleting to re-enter text like if you got the month of the expiration date wrong. 😄

Demo/Demo/Views/CustomButton.swift
scannillo: These are marked as `pending` - @jaxdesmarais wanted to make sure you saw these comments?

Demo/Demo/Card/CardFormatter.swift
minhthenguyen: gotcha, I think adding some comments here may help ppl follow the code easier :smile:

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
minhthenguyen: I agree on `shouldChangeCharactersIn`, I know it's often used to format text field. It'd be nice if we use it.
Still, I'll approve since this is already working well.
But lemme know if you wanna dig into it and decide to use it, I don't mind doing another pass :smile:

minhthenguyen (approved):
this looks good to me :smile:

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: This was already updated to use `shouldChangeCharactersIn` to get the `cardType` and `maxLength` and only allow input up to the max length for that card type. I also added a similar limit to that delegate for CVV and expiration date. 😄 

I'm not sure it makes sense to have the logic to disable the button, format the test fields, and limit character length all in 1 function, but can certainly dig into that a bit more to see if there is a clean way to do this. It looks like `shouldChangeCharactersIn` is more often used for this type of logic when there is just 1 field to validate vs we have 3 fields we need to be filled out.

minhthenguyen: I think we should we able to do it without `editingChanged()`. I just tried `shouldChangeCharactersIn` and was able format + limit length + enable button for date.
But maybe that's too much to be done in one function? We can pair and discuss later to see if you like it? :smile:

scannillo:
There's a few unresolved comments I made, they're marked as `Pending` which I'm not sure what that means - just wanted to make sure they're publically visible?

Demo/Demo/Views/CustomButton.swift
jaxdesmarais: These were moved to the init already. Looks like for some reason these comments didn't post until yesterday, so some have already been addressed.

Demo/Demo/Views/CustomButton.swift
jaxdesmarais: We need some sort of holder so we can set the title to nothing when we are animating the spinner and then re-add the title when we are done with the animation. Having some sort of holder was the best way I found to achieve this and to avoid having the title and spinner overlay one another, but certainly open to other options if there is a better way to do this!

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: This function was removed and the logic moved to `shouldChangeCharactersIn`.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -86,8 +89,88 @@ class CardDemoViewController: FeatureBaseViewController {
         ])
     }
 
+    func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
+        /// Ensures that only numbers are entered into our text fields
+        guard CharacterSet(charactersIn: "0123456789").isSuperset(of: CharacterSet(charactersIn: string)) else {
+            return false
+        }
```
scannillo, line 96: Could we instead only show a numeric keyboard for these text fields?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -86,8 +89,88 @@ class CardDemoViewController: FeatureBaseViewController {
...
+        /// Holders for the cursor positions
+        let positionOriginal = textField.beginningOfDocument
+        let cursorLocation = textField.position(from: positionOriginal, offset: (range.location + NSString(string: string).length))
+
+        if let cursorLocation = cursorLocation {
+            textField.selectedTextRange = textField.textRange(from: cursorLocation, to: cursorLocation)
+        }
+    }
+
+    private func toggleButton(cardNumber: String, expirationDate: String, cvv: String) {
```
scannillo: Naming nitpick - `toggle button` is a bit confusing. Maybe `enableButton` or `enableButtonForClick`?

Demo/Demo/Views/CustomButton.swift
scannillo: Gotcha - thanks for explaining that, makes sense.

We could do something like `setTitleColor(.clear, for: .normal)` and then set it back to `.white` or `.label`

jaxdesmarais: 😱  I didn't know `setTitleColor` even existed. Will update to use that instead!

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: The keyboard type is set as `.numberPad` but if someone was testing in simulator instead they would still be able to enter letters. Maybe that doesn't matter and this can be removed, but seemed easy enough to just handle.

scannillo (approved):
🚀

PR #19, opened 2021-10-12
Add optional params for Card

minhthenguyen:
### Reason for changes

We need to add optional params for Card

### Summary of changes

Add card params:
- `name`
- `billing_address`

Notes:
- `Address` is now living in `Card` module since that's the only place it's used for now. But may need to move to `PaymentsCore` in the future.
- Will need to add `attributes` but per https://paypal.slack.com/archives/C01UE344JD7/p1634057993450900 the API team is planning to move this property to somewhere else. Will pick this back up once there are updates from API team.
- Didn't add `card_type`, `type`, `brand` since our API can infer those. Passing them in doesn't do any difference.
- Didn't add `id` and `authentication_results` as those are not needed for now.

### Checklist

- [ ] Added a changelog entry

- @minhthenguyen


Sources/Card/Models/Address.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+
+/// The portable international postal address.
+/// Maps to [AddressValidationMetadata](https://github.com/googlei18n/libaddressinput/wiki/AddressValidationMetadata)
+/// and HTML 5.1 [Autofilling form controls: the autocomplete attribute](https://www.w3.org/TR/html51/sec-forms.html#autofilling-form-controls-the-autocomplete-attribute).
```
jaxdesmarais: This comment is a little confusing to me especially the reference to HTML 5.1 and autofilling form controls. Do we need this part? Also, I wonder if we can change `The portable international postal address.` to just `The postal address of the customer.` or something like that to clarify what this is.

Sources/Card/Models/Address.swift
```diff
@@ -0,0 +1,59 @@
...
+        case addressLine1
+        case addressLine2
+        case locality = "admin_area_2"
+        case region = "admin_area_1"
+        case postalCode
+        case countryCode
+    }
+
+    /// The first line of the address. For example, number or street. For example, `173 Drury Lane`.
+    /// Required for data entry and compliance and risk checks. Must contain the full address.
```
jaxdesmarais: Looks like we use "for example" twice. Maybe this can be `The first line of the address, number and/or street. For example, 173 Drury Lane.` Similar for the second line as above, do merchants need to know why this is required or can that me removed? If it _is_ required maybe we can update it to read as "required when using PayPal's fraud tools" or whatever the merchant facing term is for when we run those checks.

Sources/Card/Models/Address.swift
```diff
@@ -0,0 +1,59 @@
...
+    public var addressLine1: String?
+
+    /// The second line of the address. For example, suite or apartment number.
+    public var addressLine2: String?
+
+    /// A city, town, or village. Smaller than `admin_area_level_1`.
```
jaxdesmarais: From the merchant side, they likely would know what "Smaller than `admin_area_level_1`" means since that's in our coding keys and obfuscated from them. Maybe we should update this to say smaller than region instead?

Sources/Card/Models/Address.swift
```diff
@@ -0,0 +1,59 @@
...
+    public var locality: String?
+
+    /// The highest level sub-division in a country, which is usually a province, state, or ISO-3166-2 subdivision. Format for postal delivery. For example, `CA` and not `California`. Value, by country, is:
```
jaxdesmarais: I think we should either link to something for `ISO-3166-2` or remove it

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
-    public var number: String
-    public var expirationMonth: String
-    public var expirationYear: String
-    public var securityCode: String
-
     enum CodingKeys: String, CodingKey {
         case number
         case expiry
         case securityCode
+        case cardHolderName = "name"
```
jaxdesmarais: `cardHolderName` > `cardholderName` to match android

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+        case billingAddress
     }
 
+    /// The primary account number (PAN) for the payment card.
+    public var number: String
+
+    /// The card expiration month
```
jaxdesmarais: Can we add the `in MM format` back into this comment?

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+    public var expirationMonth: String
+
+    /// The card expiration year
```
jaxdesmarais: Same as above to add `in YYYY format` back into this

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+    public var expirationYear: String
+
+    /// The three- or four-digit security code of the card. Also known as the CVV, CVC, CVN, CVE, or CID.
+    public var securityCode: String
```
jaxdesmarais, line 23: Looks like Android made this optional, can we do the same?

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+
+    /// The card holder's name as it appears on the card.
+    public var cardHolderName: String?
```
jaxdesmarais: `cardHolderName` > `cardholderName`

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+
+    /// The portable international postal address. Maps to [AddressValidationMetadata](https://github.com/googlei18n/libaddressinput/wiki/AddressValidationMetadata) and HTML 5.1 [Autofilling form controls: the autocomplete attribute](https://www.w3.org/TR/html51/sec-forms.html#autofilling-form-controls-the-autocomplete-attribute).
```
jaxdesmarais: Same comment as above about the links to validation and autofilling. I don't know that these are needed.

jaxdesmarais (commented):
Mostly just questions around documentation, but otherwise this looks good! 😄

Sources/Card/Models/Address.swift
scannillo: Agreed - I think we can make this a little simpler

Sources/Card/Models/Card.swift
```diff
@@ -1,32 +1,60 @@
...
+    public var billingAddress: Address?
```
scannillo, line 29: I don't think `Card.swift` should be a top-level class to represent all potential params for the `Card` checkout flow. I think it makes more sense that a `Card` only has card number, expiration, and CVV. We're probably going to have a similar situation for all of the payment methods - needing a request object that is sort of a bucket for everything.

I think we should have a `CardRequest` with a `Card` attached to it, alongside other optional properties.

I think this will help us establish a consistent pattern across all payment methods our SDK will provide, and makes more logical sense to contain actual `Card` details into `Card`.

Sources/Card/Models/Card.swift
minhthenguyen: I asked this in api-unification-team channel once before: https://paypal.slack.com/archives/C01UHMGB4DS/p1632412932043000
And for our use case this should always be required :smile:

Sources/Card/Models/Address.swift
minhthenguyen: ope I forgot to edit this!!!! :smile:

Sources/Card/Models/Address.swift
minhthenguyen: yea I think we won't need that, I'll shorten this to just be `The first line of the address`

Sources/Card/Models/Card.swift
minhthenguyen: Hmmm so if in the future we have some payment method having no optional params, we still need to have a `<PaymentMethod>Request` object to just encapsulate `<PaymentMethod>` object to keep a consistent pattern?

I think a benefit of keeping it like what we have in our API is that when merchants look at our code, then look at our API docs, they can know exactly which params/objects in client-side match with which params/objects in server-side.
Plus I think the card holder name and the billing address makes sense to be in `Card` object 🤔

jaxdesmarais (approved):
Thanks for addressing comments on documentation! This looks great! 🚀

Sources/Card/Models/Card.swift
scannillo: As David re-iterated again today in our sync, we want the SDK to take liberties to improve merchant integration. We do not need to mirror or tie to orders API naming or interface.

I agree billingAddress and maybe some other things do make sense on Card ... but the list is of request params is likely to grow, (ex email). Another reason, a `Request` object allows us to be consistent across payment methods. Ex: `CardRequest`, `PayPalRequest`, `VenmoRequest`, `LocalPaymentsRequest`. This is helpful because the request params for other payment methods won't be as straightforward as `Card` (ex: shouldVault, displayMerchantName, etc).

Sources/Card/Models/Address.swift
```diff
@@ -0,0 +1,56 @@
...
+    public var region: String?
+
+    /// The postal code, which is the zip code or equivalent. Typically required for countries with a postal code or an equivalent.
+    /// See [postal code](https://en.wikipedia.org/wiki/Postal_code).
+    public var postalCode: String?
+
+    /// The [two-character ISO 3166-1 code](/docs/integration/direct/rest/country-codes/) that identifies the country or region.
+    /// Note:
+    /// - The country code for Great Britain is GB and not UK as used in the top-level domain names for that country. Use the `C2` country code for China worldwide for comparable uncontrolled price (CUP) method, bank card, and cross-border transactions.
+    public var countryCode: String
```
scannillo, line 39: Is this correct? All fields optional except for `countryCode`?

Also let's add the word `Optional.` to the start of each docstring so that is even more clear to merchants

Sources/Card/Models/Card.swift
minhthenguyen: understood! should we leave it for another PR after we've added other payment methods tho?
As of the params we're having now + considering Android is also having these in `Card`, I feel better leaving it like this for now. Lemme know your thoughts :smile:

Sources/Card/Models/Address.swift
minhthenguyen: yes only `countryCode` is required. I'll add `Optional` to the docs :smile:

Sources/Card/Models/Card.swift
scannillo: Sure! We can do something like this in a follow up PR 👍 - also gives us a chance to align with Android too on it

Sources/Card/Models/Address.swift
scannillo: Sweet - yeh everything looks good to go to me after we update the optional docstrings in Address and in Card

scannillo (approved):
🎃

PR #20, opened 2021-10-15
Advanced Error Handling

scannillo:
### Reason for changes

Previously we couldn't assert on our Error types and they didn't all return valuable information to merchants. 

### Summary of changes

- Create custom `PayPalSDKError`
  - Include merchant-friendly `localizedDescription`
  - Include domain & errorCode for an easy way to pinpoint exact site of failure for future inbounds and merchant debugging.
- Create `APIClientError` defining networking specific codes, domain, and descriptions.
- Create `CardClientError`

### Checklist

- ~Added a changelog entry~

### Notes

Commit history is wonky on this one, apologies. Diff should be readable and history resolved w/ squash.

### Authors
@jonathajones @scannillo


Sources/PaymentsCore/Networking/Models/ErrorData.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+struct ErrorData: Codable {
```
scannillo: Maybe we include `Response` somewhere in this name to make it clear that it technically is a `response` type that we will decode, and follow the pattern of the others.

Hecim1984 (approved):
Samo jako...

Hecim1984 (approved):
Nja

Hecim1984 (approved):
[🎉](🚀)

Hecim1984 (commented):
**** 💸

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -120,35 +150,31 @@ class APIClient_Tests: XCTestCase {
...
 
-        let jsonResponse = """
-        {
-            "test": "wrong response format"
-        }
-        """
-
         mockURLSession.cannedURLResponse = successURLResponse
-        mockURLSession.cannedJSONData = jsonResponse
+        mockURLSession.cannedJSONData = "{ \"test\" : \"bad-format\" }"
```
sshropshire, line 157:
```suggestion
        mockURLSession.cannedJSONData = """{ "test" : "bad-format" }"""
```
It seems like heredoc has been the goto format for JSON in these tests. Maybe one line heredoc here?

Sources/PaymentsCore/Networking/APIClientError.swift
```diff
@@ -0,0 +1,75 @@
+import Foundation
+
+enum APIClientError {
+
+    static let domain = "APIClientErrorDomain"
+
+    enum Code: Int {
```
sshropshire, line 7: qq: Can `Code` be made private here?

Sources/PaymentsCore/Networking/Models/ErrorData.swift
sshropshire: Yeah I agree with this too. At the moment all of the Response types follow this naming convention.

sshropshire (approved):
🌊 &nbsp; couple nits overall LGTM 🔥

Sources/PaymentsCore/Networking/APIClientError.swift
```diff
@@ -0,0 +1,75 @@
...
+    )
+
+    static let invalidURLRequestError = PayPalSDKError(
+        code: Code.invalidURLRequest.rawValue,
+        domain: domain,
+        errorDescription: "An error occured constructing an HTTP request. Contact developer.paypal.com/support."
+    )
+
+    static let serverResponseError: (String) -> PayPalSDKError = { description in
+        return PayPalSDKError(
```
jaxdesmarais: Thoughts in using an implicit return here?

Sources/PaymentsCore/Networking/Models/ErrorData.swift
jaxdesmarais: I agree with adding `Response` to the name to make this clearer

Sources/PaymentsCore/Networking/Models/ErrorData.swift
```diff
@@ -0,0 +1,10 @@
...
+    let name: String
+    let message: String
+
+    var readableDescription: String {
+        return name + ": " + message
```
jaxdesmarais: Same comment about instead using an implicit return here

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -52,7 +52,7 @@ final class CardClient_Tests: XCTestCase {
             }
         }
         """
-      
+
```
jaxdesmarais: Is this extra space needed?

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -85,11 +84,15 @@ final class CardClient_Tests: XCTestCase {
         mockURLSession.cannedJSONData = jsonResponse
 
         cardClient.approveOrder(orderID: "", card: card) { result in
-            guard case .failure(NetworkingError.parsingError) = result else {
-                XCTFail("Expect NetworkingError.parsingError for invalid response")
-                return
+            switch result {
+            case .success(_):
+                XCTFail()
```
jaxdesmarais, line 89: Can we add a message to XCTFail here? May be helpful in debugging if we ever do hit the success case.

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -85,11 +84,15 @@ final class CardClient_Tests: XCTestCase {
...
+            case .failure(let error):
+                XCTAssertEqual(error.domain, APIClientError.domain)
+                XCTAssertEqual(error.code, APIClientError.Code.dataParsingError.rawValue)
+                XCTAssertEqual(error.localizedDescription, "An error occured parsing HTTP response data. Contact developer.paypal.com/support.")
             }
-
+
```
jaxdesmarais: Is this needed?

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -64,21 +63,46 @@ class APIClient_Tests: XCTestCase {
...
+    func testFetch_withNoURLRequest_returnsInvalidURLRequestError() {
+        let expect = expectation(description: "Callback invoked.")
+
+        // Mock request whose API object does not vend a URLRequest
+        let noURLRequest = FakeRequestNoURL()
+
+        apiClient.fetch(endpoint: noURLRequest) { result, _ in
+            switch result {
+            case .success(_):
+                XCTFail()
```
jaxdesmarais, line 76: Add a message to the fail to potentially help with debugging

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -64,21 +63,46 @@ class APIClient_Tests: XCTestCase {
...
 
         mockURLSession.cannedError = serverError
         
         apiClient.fetch(endpoint: fakeRequest) { result, _ in
-            guard case .failure(.connectionIssue) = result else {
-                XCTFail("")
-                return
+            switch result {
+            case .success(_):
+                XCTFail()
```
jaxdesmarais, line 101: Add message to XCTFail to help with debugging potentially

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: You can't write the `"""` style on one line, it will complain with `Multi-line string literal closing delimiter must begin on a new line`.

Sources/PaymentsCore/Networking/APIClientError.swift
scannillo: What do you mean implicit return?

Sources/PaymentsCore/Networking/APIClientError.swift
scannillo: We could! Question is whether or not we we think it will be helpful for merchants to see these codes?

Sources/PaymentsCore/Networking/Models/ErrorData.swift
scannillo: DONE!

UnitTests/CardTests/CardClient_Tests.swift
scannillo: Fixed

UnitTests/CardTests/CardClient_Tests.swift
scannillo: I personally think adding description strings for XCTFail() cases is never really worth it. The test name already is super explicit so seeing it fail in CI is telling enough. I always thought maintaining the description strings across (lots) of tests is annoying and ended up never being really helpful. Just my experience, though!

Sources/PaymentsCore/Networking/APIClientError.swift
jaxdesmarais: So for 1 line functions/variables/etc, the [`return` keyword isn't needed](https://www.swiftbysundell.com/questions/when-can-the-return-keyword-be-omitted/). So you can do like:
```
    static let urlSessionError: (String) -> PayPalSDKError = { description in
        PayPalSDKError(
            code: Code.urlSessionError.rawValue,
            domain: domain,
            errorDescription: description
        )
    }
```
Not a dealbreaker either way, I think it's just a cool language feature 😄

Sources/PaymentsCore/Networking/APIClientError.swift
```diff
@@ -0,0 +1,75 @@
+import Foundation
+
+enum APIClientError {
```
minhthenguyen, line 3: question: do you think this needs to be `public`?

Sources/PaymentsCore/Networking/APIClientError.swift
```diff
@@ -0,0 +1,75 @@
...
+            code: Code.urlSessionError.rawValue,
+            domain: domain,
+            errorDescription: description
+        )
+    }
+
+    static let dataParsingError = PayPalSDKError(
+        code: Code.dataParsingError.rawValue,
+        domain: domain,
+        errorDescription: "An error occured parsing HTTP response data. Contact developer.paypal.com/support."
```
minhthenguyen, line 47: niceeee 🚀

Sources/PaymentsCore/Networking/Models/ErrorResponse.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+struct ErrorResponse: Codable {
+    let name: String
+    let message: String
```
minhthenguyen: I see our API has a `details` field which contains very useful info about the error. Should we add it here?

minhthenguyen: and I think `message` here can be optional? I see if I put in an invalid access token for a request we won't have `message` :smile:

Sources/PaymentsCore/Networking/APIClientError.swift
sshropshire: Ah right yeah that's a good point. In that case we'd want the docs to be public I agree.

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
sshropshire: Word yeah, the escapes on the quotations are always jarring IMO. It's a nit also not a deal-breaker for me.

scannillo: I honestly just did it here to reduce the # of lines. I can revert if y'all feel it's confusing

sshropshire: It's a valid string I'm cool with that format.

UnitTests/CardTests/CardClient_Tests.swift
jaxdesmarais: That's fair! Yeah, maybe I've just seen that done previously because test names were not as clear! 😄 🙃

Sources/PaymentsCore/Networking/APIClientError.swift
scannillo: I mean the code is really meant for us mostly - but we can leave public in case merchants want that extra detail

Sources/PaymentsCore/Networking/APIClientError.swift
scannillo: There is no need for this to be public! This file is a way for us internally to consolidate all error creation per module / domain. Having them all in one place helps us make sure we're consistent in our instantiation, wording, etc per error.

But this file doesn't need to be public since a merchant will get back the relevant `PayPalSDKError` in the completion of whatever checkout flow they kicked off (ex: `cardClient.approveOrder(orderID: String, card: Card, completion: @escaping (Result<OrderData, PayPalSDKError>) -> Void)`)

scannillo: Let me know if that makes sense

Sources/PaymentsCore/Networking/APIClientError.swift
scannillo: Good eye - lets do it!

Sources/PaymentsCore/Networking/APIClientError.swift
minhthenguyen: ah yep that makes sense :smile:

Sources/PaymentsCore/Networking/Models/ErrorResponse.swift
scannillo: Good call - we can make this optional.

scannillo: In terms of `details` we decided that wasn't super useful info for merchants. We did some manual testing / triggering different error scenarios and for now are surfacing the name & message, after we determined these to be most helpful.

minhthenguyen (approved):
this looks good to me pending resolving merge conflicts :smile:

Sources/PaymentsCore/Networking/Models/ErrorResponse.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+struct ErrorResponse: Codable {
+    let name: String
```
jaxdesmarais, line 4: nit: should we add a space before name to match other files?

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -91,11 +115,14 @@ class APIClient_Tests: XCTestCase {
...
+            switch result {
+            case .success(_):
                 XCTFail()
-                return
+            case .failure(let error):
+                XCTAssertEqual(error.domain, APIClientError.domain)
+                XCTAssertEqual(error.code, APIClientError.Code.invalidURLResponse.rawValue)
+                XCTAssertEqual(error.localizedDescription, "An error occured due to an invalid HTTP response. Contact developer.paypal.com/support.")
             }
-
```
jaxdesmarais, line 98: nit: it looks like in some places we include a space before `expect.fulfill()` and other places we've removed them. Should this be the same everywhere?

jaxdesmarais (approved):
Couple of small non-blocking nits, but approved otherwise. This new error handling is gonna make things a lot more clear for merchants! 🥳

Sources/PaymentsCore/Networking/Models/ErrorResponse.swift
scannillo: Good eye - Jon and I noticed a discrepancy on this throughout the project as a whole - we think a follow up PR should be make to make things more consistent everywhere.

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Good call - updated!

Sources/PaymentsCore/Networking/Models/ErrorResponse.swift
jaxdesmarais: Follow up PR sounds great! Maybe something to look into would be a linter rule for this so we get a warning when we don't include a space (I'm not sure that's possible but may be helpful if it is).

PR #21, opened 2021-10-15
HTTPClient

sshropshire:
### Reason for changes

This PR adds an `HTTPClient` to allow for a single `HTTPResponse` type used to simplify the response of a `(URLResponse?, Data?, Error?)` result from `URLSession` data tasks. 

### Summary of changes

- Create `API` and `HTTP` protocols
- Create `HTTPClient` and `APIClient` to use `HTTPClient`
- Move response parsing up a level in `CardClient`
- Create `MockAPIClient` and `MockHTTPClient` to use in unit tests

### Checklist


Sources/Card/Models/Card.swift
```diff
@@ -6,7 +6,7 @@ import Foundation
 ///   - expirationMonth: The card's expiration month in MM format.
 ///   - expirationYear: The card's expiration year in YYYY format.
 ///   - securityCode: The card's security code (CVV, CVC, CVN, CVE, or CID).
-public struct Card: Encodable {
+public struct Card: Encodable, Equatable {
```
sshropshire, line 3: Added `Equatable` conformance to make test assertions easier but this should probably be removed.

jonathajones:
Could you talk more about the motivation for this PR? I'm kind of confused since it seems to be a re-implementation of the networking layer. is there a desire to align it more to Android's implementation ?

sshropshire:
> Could you talk more about the motivation for this PR? I'm kind of confused since it seems to be a re-implementation of the networking layer. is there a desire to align it more to Android's implementation?

There isn't a strict requirement to align the two, but they are both similar. There are some patterns on the Android side that worked really well for us from a testing perspective. This PR aims to re-imagine the iOS networking layer with those same patterns to see if it's worth it.

I'll add some inline comments to highlight the benefits of a refactor.

Sources/Card/CardClient.swift
```diff
@@ -5,20 +5,18 @@ import PaymentsCore
 
 public class CardClient {
 
-    private let apiClient: APIClient
-    private let config: CoreConfig
+    private let api: API
```
sshropshire, line 8: Having a separate `API` protocol allows for easy mock implementation / injection here.

Sources/Card/CardClient.swift
```diff
@@ -27,31 +25,31 @@ public class CardClient {
...
-                clientID: config.clientID
-            )
+    public func approveOrder(
+        orderID: String,
+        card: Card,
+        completion: @escaping (Result<OrderData, NetworkingError>, CorrelationID?) -> Void
+    ) {
+        let path = "/v2/checkout/orders/\(orderID)/confirm-payment-source"
+        var request = APIRequest(method: .post, path: path)
+        request.body = ConfirmPaymentSourceBody(card: card)
```
sshropshire, line 35: Here an `APIRequest` object can be used to provide both the path, HTTP method, and request body; similar to a `URLRequest`.

Sources/Card/CardClient.swift
```diff
@@ -27,31 +25,31 @@ public class CardClient {
...
-                switch result {
-                case .success(let response):
-                    let orderData = OrderData(orderID: response.id, status: OrderStatus(rawValue: response.status))
-                    completion(.success(orderData))
+        api.send(request) { response in
+            let correlationID = response.headers["Paypal-Debug-Id"] as? String
 
-                case .failure(let error):
-                    completion(.failure(error))
+            if response.status == 404 {
```
sshropshire: Having full access to an `HTTPResponse` object allows us to take more advantage of the `HTTP` protocol. From each feature client, we can access the status code, headers, etc.

Sources/Card/CardClient.swift
```diff
@@ -27,31 +25,31 @@ public class CardClient {
...
+                completion(.failure(NetworkingError.unknown), correlationID)
+            } else if let body = response.body {
+                do {
+                    let result = try self.decoder.decode(ConfirmPaymentSourceResponse.self, from: body)
```
sshropshire, line 42: Lifting the responsibility of decoding the response object from `APIClient` to the feature client allows us to keep response objects within their respective modules without having to use generics.

Sources/Card/ConfirmPaymentSourceBody.swift
```diff
@@ -0,0 +1,17 @@
+import UIKit
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+struct ConfirmPaymentSourceBody: APIRequestBody {
```
sshropshire, line 6: Request body objects are fully encodable and can be directly set as a property of `APIRequest`.

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -1,52 +0,0 @@
...
-        // TODO: Remove clientSecret when this endpoint is updated and can be used with low-scoped token
-        // For now, include your clientSecret here and your clientID in CoreConfig to test this request
-        let clientSecret = ""
-        let encodedClientID = "\(clientID):\(clientSecret)".data(using: .utf8)?.base64EncodedString() ?? ""
-
-        return [
-            .contentType: "application/json",
-            .acceptLanguage: "en_US",
-            .authorization: "Basic \(encodedClientID)"
-        ]
```
sshropshire, line 31: The `APIRequest` pattern allows us to remove duplication with `Authorization`, `Content-Type`, and `Language` headers.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -1,84 +1,51 @@
...
+        // For now, include your clientSecret here and your clientID in CoreConfig to test this request
+        let clientSecret = ""
+        let encodedClientID = "\(config.clientID):\(clientSecret)".data(using: .utf8)?.base64EncodedString() ?? ""
 
-protocol URLSessionProtocol {
-    func performRequest(with urlRequest: URLRequest, completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void)
-}
+        request.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        request.addValue("en_US", forHTTPHeaderField: "Accept-Language")
+        request.addValue("Basic \(encodedClientID)", forHTTPHeaderField: "Authorization")
```
sshropshire, line 47: The `APIClient` is `CoreConfig` aware, and it is capable of turning an `APIRequest` into a `URLRequest`.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -1,84 +1,51 @@
...
-    public func fetch<T: APIRequest>(
-        endpoint: T,
-        completion: @escaping (Result<T.ResponseType, NetworkingError>, CorrelationID?) -> Void
-    ) {
-        guard let request = endpoint.toURLRequest(environment: environment) else {
-            completion(.failure(.noURLRequest), nil)
-            return
+    public func send(_ apiRequest: APIRequest, completion: @escaping (HTTPResponse) -> Void) {
+        if let urlRequest = urlRequest(from: apiRequest) {
+            http.send(urlRequest, completion: completion)
```
sshropshire, line 22: `APIClient` doesn't need to know about the underlying `URLSessionDataTask` to do its job, and receives a single `HTTPResponse` instead of the usual `(Data?, URLResponse?, Error?)` tuple.

Sources/PaymentsCore/Networking/APIRequestBody.swift
```diff
@@ -0,0 +1,14 @@
...
+public protocol APIRequestBody: Encodable {
+    func encoded() -> Data?
+}
+
+extension APIRequestBody {
+
+    public func encoded() -> Data? {
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        return try? encoder.encode(self)
```
sshropshire, line 12: We can keep encoding DRY by providing a public extension in the core module to produce a `Data` encoded `APIRequestBody`.

Sources/PaymentsCore/Networking/HTTPClient.swift
```diff
@@ -0,0 +1,31 @@
...
+    // MARK: - Properties
+    private let urlSession: URLSessionProtocol
+
+    // MARK: - Initializers
+    public init(urlSession: URLSessionProtocol = URLSession.shared) {
+        self.urlSession = urlSession
+    }
+
+    // MARK: - HTTP
+    func send(_ urlRequest: URLRequest, completion: @escaping (HTTPResponse) -> Void) {
```
sshropshire, line 19: Having an `HTTPClient` allows for easing stubbing of network calls. `HTTPClientMock` allows us to set stubs for specific incoming `URLRequest`s.

Sources/PaymentsCore/Networking/Models/Requests/AccessTokenRequest.swift
```diff
@@ -1,25 +0,0 @@
-import Foundation
-
-/// Describes request to get a low scoped access token (LSAT) given some Client ID
-public struct AccessTokenRequest: APIRequest {
-    public typealias ResponseType = AccessTokenResponse
-
-    public var clientID: String
```
sshropshire, line 7: We technically aren't using this for the time being. It makes sense to keep the codebase clean until we need it. We can always revert this request in a separate PR.

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -3,206 +3,112 @@ import XCTest
...
+        let httpResponse = HTTPResponse(status: 123)
+        http.stubSend(with: httpResponse)
+        
+        let sut = APIClient(config: sandboxConfig, http: http)
+        
+        let apiRequest = APIRequest(method: .post, path: "/sample/url")
+        
+        let expect = expectation(description: "forwards http response to completion handler")
+        sut.send(apiRequest) { response in
+            XCTAssertEqual(123, response.status)
```
sshropshire, line 109: `APIClient` tests become simpler since the networking logic has been moved into `HTTPClient`.

Sources/Card/CardClient.swift
jonathajones: evaluating the network response feels like it is appropriately a responsibility of the networking layer and can be abstracted away from the Client types for the specific modules. by bubbling this up to the `CardClient` (and every other module), we would have to largely re-implement our response parsing for each module / function that makes a network request

Sources/Card/CardClient.swift
jonathajones: Why would we want to avoid generics?

Sources/Card/ConfirmPaymentSourceRequest.swift
jonathajones: this is assuming that all of our requests will have the same set of basic headers. If this is true - we can accomplish the same thing by moving the headers into the existing `APIClient`, or into the existing `APIRequest` protocol extension. I don't see the benefit (may also be misunderstanding here -  please correct me !)

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -0,0 +1,6 @@
+import UIKit
+
+protocol HTTP {
+
+    func send(_ urlRequest: URLRequest, completion: @escaping (HTTPResponse) -> Void)
+}
```
jonathajones, line 6: This gives us two protocols (`HTTP` and `URLSessionProtocol`) that seem to largely overlap in their functionality - both existing to enable us to more easily write tests and mock responses. this seems confusing to me - do we need both?

UnitTests/TestShared/API/MockAPIClient.swift
```diff
@@ -0,0 +1,28 @@
+
+import UIKit
+@testable import PaymentsCore
+
+class MockAPIClient: API {
+    
+    private(set) var lastSentRequest: APIRequest?
```
jonathajones, line 7: not sure why we would need to track this

Sources/PaymentsCore/Networking/APIRequestBody.swift
jonathajones: I think the error thrown by the decoder would need to be handled, instead of silenced by the `?`. if the encoding fails, we wouldn't want to wait til the network request fails to notify the merchant that something has gone wrong

Sources/PaymentsCore/Networking/HTTPClient.swift
jonathajones: can you expand on this? currently, network mocking would look like the following: 

```swift
        mockURLSession = MockURLSession()

        mockURLSession.cannedError = <error to vend>
        mockURLSession.cannedURLResponse = <response to vend>
        mockURLSession.cannedJSONData = <data to vend>
        
        apiClient = APIClient(urlSession: mockURLSession, environment: .sandbox)

        apiClient.fetch(endpoint: <someEndpoint>) { result, _ in 
            /// the `result` vended here reflects the mock data just specified
        }
```

with your changes, a similar mock would look like this: 
```swift
        let httpResponse = HTTPResponse(status: 123) // specify the mock expected response here
        http.stubSend(with: httpResponse) // give response to the urlsession mock

        let apiClient = APIClient(config: sandboxConfig, http: http)

        let apiRequest = APIRequest(method: .post, path: "/sample/url") // create the request

        apiClient.send(apiRequest) { response in
            /// `response` represents the mock data
        }
```
The only meaningful difference that I see between these approaches is how we give the mock expected response to the mock URLSession (which is what I understand the `MockHTTPClient` to be). In the current implementation, the response data is set on the mock session directly, and in your example the mock data is captured first in the `HTTPResponse` type before being handed to the mock session. 

I understand it's hard to compare the comparative ease of response testing between two different types of networking layers, but i don't i understand the benefit yet

jonathajones (changes requested):
I think at the root of my questions about this PR is that I'm not really convinced of the benefit. 

two main things I would push back on : 
1. Moving the HTTP response parsing up from the networking layer into the card module would mean that every module (and every top level function that makes a network request) would need to do some HTTP response parsing on its own. that would almost entirely be duplicate logic, wouldn't it? i think one of the benefits of a centralized networking layer is that the response parsing can be entirely abstracted away from the business logic that would occur in the modules

2. This PR moves away from having HTTP requests modeled by individual objects, and towards having requests created ad hoc from a generic root type (`APIRequest`). This makes reuse more difficult, and it also makes it harder to know if any particular network request that you would want to make already exists somewhere in the codebase. I don't see the benefit of moving away from generics

Sources/PaymentsCore/Networking/HTTP.swift
jonathajones: I'd include the `API` protocol in this list as well - the only difference between `API` and `HTTP` is that one takes a `URLRequest` and the other an `APIRequest` (which itself describes most of what is required to make a URLRequest). I feel like there's some potential simplification here that would be make this a bit more straightforward

Sources/Card/CardClient.swift
sshropshire: I understand the case for DRY code. The response parsing could be moved to a parser potentially. Having access to the HTTP status would help an `EmptyResponse` scenario though. A [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) will communicate to the SDK that the body response is empty.

Sources/Card/CardClient.swift
sshropshire: Mostly to make the code simpler in the networking layer. Generics tend to decrease flexibility / readability of the code. They are a good tool for producing DRY code, but an alternative solution using protocols could allow more flexibility. I saw some [terse commentary on Generics usage](https://www.quora.com/When-should-you-use-generics-in-Swift/answer/Josh-Osborne-6?ch=10&oid=41601058&share=54b5630c&target_type=answer) a while back that came to mind.

Sources/Card/ConfirmPaymentSourceRequest.swift
sshropshire: Yeah I agree moving this into `APIClient` would also make the code more DRY. At the moment any calls to a `ConfirmPaymentSourceRequest` will depend on having access to CoreConfig to obtain the `clientID`. Moving authentication into the networking layer will make `APIRequest` objects easier to create.

Sources/PaymentsCore/Networking/APIRequestBody.swift
sshropshire: Yeah honestly this was the slickest way to do it lol. This should probably throw.

Sources/PaymentsCore/Networking/HTTPClient.swift
sshropshire: Right they are essentially the same. IMO `HttpResponse` is a more cohesive unit than the three-param tuple returned by the `URLSessionDataTask`. We do also have the option to implement `http.stubSend(for urlRequest: URLRequest, with response: HttpResponse)` if we have multi request scenarios in the future.

Sources/PaymentsCore/Networking/HTTP.swift
sshropshire: `HTTPRequest` and `HTTPResponse` are meant to encapsulate the networking layer into our domain. Since they're plain value objects, they are good assertion vehicles when unit testing. The same goes for `APIRequest`. The only awkward part (on Android too) is there is no _APIResponse_ which breaks the symmetry, but having access to the HTTPResponse was sufficient enough.

UnitTests/TestShared/API/MockAPIClient.swift
sshropshire: For assertions we use it [here](https://github.com/paypal/iOS-SDK/blob/19ebdeb7523568ca50ce6e02ca20cdcce425ccf0/UnitTests/CardTests/CardClient_Tests.swift#L19) e.g.

sshropshire:
@jonathajones added some inline responses. I would also like to note that Android/Kotlin does not have a `Codable` counterpart, and `URLSessionDataTask` is a much cleaner API than [HTTPURLConnection](https://github.com/paypal/Android-SDK/blob/main/Core/src/main/java/com/paypal/android/core/Http.kt).

A layered architecture does appear to have benefits on both iOS and Android. We use a similar pattern on BT Android and it has scaled well. The usage of generics isn't necessarily a deal-breaker, but the need for an `EmptyResponse` type was the primary catalyst for this refactoring exercise.

sshropshire:
> it also makes it harder to know if any particular network request that you would want to make already exists somewhere in the codebase.

This is a good callout I think as the codebase grows this will be a fundamental risk and may depend on project structure / onboarding for new team members.

Sources/Card/CardClient.swift
minhthenguyen: I agree with Jon here about this maybe more suitable being APIClient's responsibility.
In here we only handle 404. How about other error cases where we do have a response body? Should they all fall into `parsingError` while in fact we know they're not? Otherwise I imagine we'd need to handle all other status codes in here (and in all other similar API wrapper functions we'll add in the future). imo it feels a bit redundant while in fact we can handle all error status codes the same way just once in APIClient, which helps maintainability.

sshropshire: Yeah that is a good callout here. I mean to add an `isSuccessful` convenience boolean to check whether the request is considered a success. I didn't get heavy into error handling I know Sammy and Jon have a PR up for that.

Sources/Card/CardClient.swift
minhthenguyen: As far as I know one request (endpoint) expects exactly one success response type 🤔 
so maybe generics is acceptable? do we need flexibility for this use case? 

On the concern about API evolving causing breaking changes that you and David have, how does this approach help? we would need to convert raw response -> merchant facing data objects no matter which approach we end up using (the only difference I see is previously we do it in APIClient, now we do it in CardClient).
If an API change will cause breaking changes to our public interface, I think it applies to both approaches. please correct me if I'm wrong

scannillo: I can see why moving the parsing logic into each module makes sense/ is helpful. Especially for weirder scenarios like the EmptyResponse case, or once we had a specific 4XX code come back from a PayPal endpoint that actually had success/ result data we needed to parse to continue the flow (weird - but happened).

minhthenguyen: hmm I think if status code is 4xx, it's correct for the APIClient to returns an error. If some modules/clients wanna treat it as a success, then it can still get whatever error responses that APIClient returns and do what it wants with that 

About `EmptyResponse`, I see Alamofire has [the same thing](https://alamofire.github.io/Alamofire/Structs/Empty.html) for the exact same reason, so I can't see why it's a concern 🤔  if we wanna eliminate the weirdness in how we handle empty response in APIClient, it can just simply be a custom decoder for `EmptyResponse` 🤔

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: Yeh, I agree with Jon these three protocols are a bit confusing. I imagine Steven you added them to be able to create test mocks at each layer?

Maybe we can think of a special naming convention for protocols who sole purpose is to enable testing? Also, I think putting them into their own file also makes it harder to sift through / more confusing. Can we have the protocol definitions which are tightly coupled to a specific implementation file live there?

sshropshire: Yeah the protocols are specifically for injecting mocks. The naming isn't final `HTTPClientProtocol` and `APIClientProtocol` felt kind of verbose.

Sources/PaymentsCore/Networking/HTTPClient.swift
scannillo: >  IMO HttpResponse is a more cohesive unit than the three-param tuple returned by the URLSessionDataTask

I think that 3-param tuple is just super commonplace for iOS networking layers & testing. But it looks like the real benefit of the `HTTPResponse` that you added comes in when unit testing feature clients like `CardClient`. I personally don't mind the 3-tuple because I'm so used to it, but I think the `HTTPResponse` idea is neat and makes the tests (and feature clients) easy to read.

sshropshire: Yeah having to cast `URLResponse` to `HTTPURLResponse` is another thing that can be done internally. We could also add some generic error handling in `HTTPClient` to make things easier on `APIClient` e.g. connection timeouts, etc. Since it's within our domain we can return `NetworkingError`s via callback and `APIClient` can focus on the body of the result.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: In our previous implementation we were using a protocol extension on `APIRequest` to provide a default implementation for constructing the URLRequest. Though I don't have experience using it - it feels like a pattern we should avoid. 

It isn't common practice that protocols implement any behavior. And when you look into the object’s implementation you don't know about it’s extended functionality. It felt like a workaround to avoid the extra code that comes with subclassing.

Sources/Card/ConfirmPaymentSourceBody.swift
scannillo: I think subclassing `APIRequest` is a little simpler / easier to understand than the generics approach from prior.

Sources/Card/CardClient.swift
scannillo: > @minhthenguyen If some modules/clients wanna treat it as a success, then it can still get whatever error responses that APIClient returns and do what it wants with that

Currently, it can't. The APIClient isn't surfacing those specific details to the feature clients.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Then again - I haven't worked with them before. Curious what you and @jonathajones think on that?

Sources/Card/CardClient.swift
minhthenguyen: is it very difficult to do that with current implementation? 🤔

sshropshire: It seems like Alamo may have introduced this [out of necessity](https://github.com/Alamofire/Alamofire/issues/2466) for non-standard API responses. They originally inferred an empty response with the `204 No Content` HTTP status code. I'd expect the Orders API holds up a lot of these HTTP best practices.

minhthenguyen: I see [they use generics](https://github.com/Alamofire/Alamofire/blob/effb907b0e27bbdff1d773c39346e1bac319fc11/Source/ResponseSerialization.swift#L961) for response types -> check if data exists -> if yes try decoding the generic type from data, if no (which by default is for 204 205 for them, or if the generic type is `EmptyResponse`) return an instance of the `EmptyResponse`, correct me if I'm wrong, but it sounds pretty much like the logic we have

Do you see there is any specific downside from having/handling `EmptyResponse`?

sshropshire: I see. I think it makes sense for Alamofire conceptually they have to support any restful API a developer wishes to connect to. In our case, we're only talking to orders at the moment. Having an additional type for this scenario is also adding additional complexity.

Sources/PaymentsCore/Networking/APIClient.swift
jonathajones: I would say that protocol extensions aren't unusual - but they should be used sparingly, since they can lead to unusual behavior depending on how the concrete object is being cast when using a protocol function with a default implementation.

Our previous implementation uses a protocol to describe a url request, and i think that's still the right way to go. There isn't a reason to use a reference type (class / subsclass) to describe a network request - an api request should only really hold the data needed to to help the network layer create a network request, which can be done by a value type more safely (we won't have to worry about mutability or shared access to the reference type). Since structs can't inherit from classes, i think that using a protocol interface is still preferable to subclassing / inheritance. 

practically, since this is not something that we will expose to merchants, either approach would work - but i do think using a protocol interface with conforming structs is a safer way to get the job done. 

The protocol extension itself was a way of pulling shared logic out of the individual types, but still provide a way for the concrete types to create the url requests a different way (by implementing the `toURLRequest(environment:)` protocol function on the conforming struct

Sources/Card/CardClient.swift
jonathajones: The way that alamofire is using generics matches what we are using generics for though, doesn't it? in both cases, the goal is to provide a typed response to whoever is making a network request. 

we may have to interact with an API where we need to handle the url response in an unconventional way - and in that case i think we could make use of another fetch function (was trying to get at that [on the original networking PR here](https://github.com/paypal/iOS-SDK/pull/9/files#r713174796)). but i don't think we should go out of our way to support that until we know that we will need to . 

seems we may be coming down to a difference of opinion though - i don't see the complexity added by an `EmptyResponse` type to be very great. and i think that, when it's used, it's pretty semantically clear - 

```swift
public struct SomeAPIRequest: APIRequest {
    public typealias ResponseType = EmptyResponse // no response is expected from the server
    ...
}
```

jonathajones:
the main thing that I'm getting kind of hung up on right now is that it's still not super clear to me what is being fixed between this solution and the one on main. is the current implementation overly confusing / hard to test? i think if i had a better understanding of what the goals of the PR are, it would be easier to evaluate against that. I definitely think there are some good ideas in here that we should incorporate - but i have a baby/bathwater feeling about dropping typed responses / and having the modules handle response parsing right now, without really understanding what we'd be gaining here yet that we don't already have support for

maybe we can schedule some time to discuss as a group? i feel like we might be reaching the limits of what we can communicate in a text review :)

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Jon, thanks for this - very helpful. 

Agreed that structs are the preferred type to be used in Networking.

So in this refactor, `APIRequest` is a struct and `APIRequestBody` is a protocol. Which part are you referring to where classes / inheritance is being used?

jonathajones: Ah it was in response to 
> It felt like a workaround to avoid the extra code that comes with subclassing.

scannillo: Ah makes sense! Thanks for clarifying that.

scannillo:
In prep for our meeting on Thursday, here is a bulleted list of the things tackled in this PR:

- Increased parity with Android
- Simplification of APIRequest construction & logic
- Moves response parsing into feature module
- Separates HTTP layer out of APIClient

sshropshire:
Closing for now. We may use this as a reference as the architecture evolves.

PR #22, opened 2021-10-19
Add Card Client to Demo App

jaxdesmarais:
### Reason for changes

Currently, we don't utilize the card module in the demo app, this update uses it 😄 

### Summary of changes

- Integrate with card client in the demo app
- Call `createOrder` `authorizeOrder/captureOrder` with our sample server
- Add implementation for `processOrder` in `FeatureBaseViewController`
- Add init to `CoreConfig`

### Checklist

- @minhthenguyen @scannillo @jaxdesmarais


Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -163,6 +235,12 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
     }
 
     private func enableButton(cardNumber: String, expirationDate: String, cvv: String) {
```
scannillo: Naming nitpick - which button is this enabling? `enablePayButton` or something

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -91,11 +90,31 @@ class FeatureBaseViewController: UIViewController {
                 self.updateTitle("Your order has failed, please try again")
                 print("❌ failed to fetch orderID: \(error)")
             }
+            completion()
```
scannillo: Would it be more common practice if we passed the `orderID` as a param in this completion?

Demo/Demo/DemoSettings/DemoSettings.swift
```diff
@@ -20,4 +20,14 @@ enum DemoSettings {
         UserDefaults.standard.string(forKey: DemoTypeDefaultsKey)
             .flatMap { DemoType(rawValue: $0) } ?? .card
     }
+
+    static var clientID: String {
+        switch environment {
+        case .sandbox:
+            return "AdFeO5qt2rodLjSgxSj66sjlBylwpacFtov1VgnptN_VclH1ZNhIcgS1px02mNIOUSEO54B-5drJ7F85"
```
jaxdesmarais: How do we want to store these credentials (client ID / secret). I assume we don't actually want them exposed publicly in our demo app? Seems like us having these values just in our app could lead to merchants doing something nefarious (I know it's sandbox, but still probably not a great idea).

scannillo: We _could_ fetch the clientID from the sample server and store the credential on Heroku.

But I honestly think it is fine to leave here - highly doubt a merchant will try to tamper with it or use it. Even if they did access it - there isn't much they could do with it.

sshropshire (approved):

Demo/Demo/DemoSettings/Environment.swift
```diff
@@ -1,3 +1,5 @@
+import PaymentsCore
+
```
jonathajones, line 2: We don't need this import

jonathajones: oh nevermind, we do 🙃

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -1,7 +1,22 @@
...
+    // MARK: - PayPal SDK Setup
+
+    let coreConfig = CoreConfig(
+        clientID: DemoSettings.clientID,
+        environment: DemoSettings.environment.paypalSDKEnvironment
+    )
+
+    lazy var cardClient: CardClient = {
+        CardClient(config: coreConfig)
+    }()
```
jonathajones: `coreConfig` can't be a constant here - otherwise the `cardClient` will only ever reflect the initial `Environment` value. 
This could be solved either by making both the config and card client computed, or by resetting both their values whenever the environment is changed

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -89,6 +106,67 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         ])
     }
 
+    // MARK: - FeatureBaseViewController Override
+
+    /// Allows us to only enable the pay button once all fields are filled out and we have an order ID
+    override func createOrderTapped(completion: @escaping (String?) -> Void = { _ in }) {
+        super.createOrderTapped { _ in
```
jonathajones: we can get rid of `super` here - we want to call this class' implementation of `createOrderTapped`, even if it is inherited from a superclass

jonathajones: wait i may have misunderstood.  this line is confusing - why do we need the `= { _ in }` in the function definition?

jonathajones: oh it's the default parameter value. that syntax is kind of confusing.

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -24,7 +25,7 @@ class FeatureBaseViewController: UIViewController {
 
     lazy var createOrderButton: CustomButton = {
         let createOrderButton = CustomButton(title: "Create Order")
-        createOrderButton.addTarget(self, action: #selector(createOrderTapped), for: .touchUpInside)
+        createOrderButton.addTarget(self, action: #selector(createOrderTapped(completion:)), for: .touchUpInside)
```
jonathajones: it feels strange to have a button selector also have a completion handler - what happens to the vended orderID? We should separate this into two functions imo - one `createOrderTapped`, and then a separate function that calls our SDK and eventually gets the order ID. what do you think ?

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -87,15 +86,37 @@ class FeatureBaseViewController: UIViewController {
...
+                completion(nil)
             }
         }
     }
 
-    func processOrder() {
-        // TODO: Pass in card details once card client is merged and call processOrder from DemoMerchantAPI
+    func processOrder(orderData: OrderData, completion: @escaping () -> Void = {}) {
+        var intent = DemoSettings.intent.rawValue.capitalized
+        intent.removeLast()
```
jonathajones: why are we calling `removeLast` on an `Intent` enum? is it just to trim the string? 

i feel like we're saving some lines of code from a `switch` statement here that is costing us clarity, since it's not intuitive that we're treating the intent `String` like and `Array` here by dropping the last character. What do you think?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -89,6 +106,67 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
...
+            DispatchQueue.main.async {
```
jonathajones: Since we control when the closure is fired, maybe we always call this completion handler from the main thread, to avoid having to wrap all our invocations in `DispatchQueue.main.async`

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -107,30 +185,24 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
...
             return false
         } else if textField == cvvTextField {
             /// Limit cvv character count to be 4 characters max
             guard new.count <= 4 else { return false }
             textField.text = new
-            enableButton(cardNumber: cardNumber, expirationDate: expirationDate, cvv: textField.text ?? "")
+            enablePayButton(cardNumber: cardNumber, expirationDate: expirationDate, cvv: textField.text ?? "")
             return false
         }
         return true
```
jonathajones: we can swift this up a bit : 
```swift
      switch textField {
      case cardNumberTextField:
        ...
      case expirationTextField:
        ...
      case cvvTextField:
        ...
      default:
        return true
      }
```

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
minhthenguyen: I like it. I think it makes sense to have `completion` in a function which actually does the create order request rather than this buttonTapped function :smile:

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
minhthenguyen: yea we did this to build the title text `Authorizing/Capturing order ...` below.
I agree a switch statement will be clearer here!!! :smile:

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -89,6 +105,65 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
...
+    /// Allows us to only enable the pay button once all fields are filled out and we have an order ID
+    override func createOrder(completion: @escaping (String?) -> Void = { _ in }) {
+        super.createOrder { _ in
+            self.enablePayButton(
+                cardNumber: self.cardNumberTextField.text ?? "",
+                expirationDate: self.expirationTextField.text ?? "",
+                cvv: self.cvvTextField.text ?? ""
+            )
+        }
+    }
```
jonathajones: This function takes a completion handler, but if you call this function (on `CardDemoViewController`), the completion handler you provide will be ignored (since you're manually specifying the call to `super` with a hardcoded completion handler inline). 

This doesn't seem like it should be an overriding function - whoever invokes this should just be calling `createOrder` on the superclass and provide this completion block there. what do you think?

minhthenguyen: ah yep. What do you think of we calling `completion` before line 113 to make sure we don't override the `completion` in the function argument?

jonathajones: I think that whoever calls this function should instead call 
```swift
     createOrder { _ in 
            self.enablePayButton(
                cardNumber: self.cardNumberTextField.text ?? "",
                expirationDate: self.expirationTextField.text ?? "",
                cvv: self.cvvTextField.text ?? ""
            )
     }
```
and remove this overriding function altogether

jonathajones: Instead of overriding `createOrder`, what if we override the `@objc` selector? ie -

```swift
    @objc override func createOrderTapped() {
        updateTitle("Creating order...")
        createOrder { _ in
            self.enablePayButton(
              cardNumber: self.cardNumberTextField.text ?? "",
              expirationDate: self.expirationTextField.text ?? "",
              cvv: self.cvvTextField.text ?? ""
            )
        }
    }
```
and remove the override on `createOrder`

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -89,6 +105,65 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
...
+        createOrder { _ in
+            self.enablePayButton(
+                cardNumber: self.cardNumberTextField.text ?? "",
+                expirationDate: self.expirationTextField.text ?? "",
+                cvv: self.cvvTextField.text ?? ""
+            )
+        }
+    }
+
+    // MARK: - Card Module Integration
```
scannillo, line 121: Good call to have a `MARK` for this. Makes identifying the SDK integration easy!

scannillo (approved):

jonathajones (approved):

PR #23, opened 2021-10-20
Adding a CardResult to the response provided by the CardClient

jonathajones:
These changes are steps towards aligning on the result types vended to merchant: https://engineering.paypalcorp.com/jira/browse/DTNOR-203

### Reason for changes
We replaced the `OrderData` response data with a card specific `CardResult` type. This will help us provide more module specific information back to the merchant, since the data we want to vend might be different for the Card module vs the PayPal module. We considered having these results backed by a shared protocol in core, but didn't think it added much value. 

### Summary of changes
- Removing OrderData from Core, replacing it with Card specific `CardResult` type. 

- Updated the expected `ResponseType` conformance from `Codable` to `Decodable`

### Checklist

~ [ ] Added a changelog entry ~

### Authors
- @scannillo @jonathajones


Sources/Card/ConfirmPaymentSourceResponse.swift
```diff
@@ -1,6 +1,21 @@
...
+import PaymentsCore
+#endif
 
-struct ConfirmPaymentSourceResponse: Codable {
-    var id: String
-    var status: String
+struct ConfirmPaymentSourceResponse: Decodable {
+    let id: String
+    let status: OrderStatus
+
```
sshropshire:
```suggestion
```

Sources/Card/Models/CardResult.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+/// The result of a card payment flow.
+public struct CardResult {
+
+    /// The order ID associated with your transaction
```
sshropshire:
```suggestion
    /// The order ID associated with the transaction
```

sshropshire (approved):
🙌

Sources/PaymentsCore/Enums/OrderStatus.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 /// The status of an order
-public enum OrderStatus: String {
+public enum OrderStatus: String, Codable {
```
minhthenguyen, line 4: ope I didn't notice this wasn't `Codable` before 😮 
I wonder how could it work before, are String enums automatically codable without having to conform somehow? :-ss

Sources/Card/Models/CardResult.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+/// The result of a card payment flow.
+public struct CardResult {
```
minhthenguyen, line 7: I just realized if we pass header field `Prefer: return=representation`, we'll be able to get `intent` in the response.
Do you think we should add it to the request, then add `intent` here so merchant won't have to keep track of `intent` by themselves (plus make it more consistent with paypal where we have `ApprovalData.intent`) ? :smile:

minhthenguyen (approved):
I just have two questions, but everything looks good to me :smile:

Sources/Card/Models/CardResult.swift
scannillo: This is a good callout! I think we should check in w/ Android folks on this, see if it's something we want to add. We can add it in additional PR if so. How does that sound to you?

minhthenguyen: ah yep that makes sense!!!! :smile:

PR #24, opened 2021-10-22
Swiftlint Custom Rule - Spacing After Class/Struct

jaxdesmarais:
### Reason for changes

We had decided previously to include a space after class/struct/etc. There is not an out of the box way to do this with Swiftlint so a custom rule has been written to do this check as part of our linter.

### Summary of changes

- Add custom rule
- Fix UnitTest path as tests were not being run through the linter as expected
- Resolve all current linter warnings after above changes were made

### Checklist

- @jaxdesmarais


.swiftlint.yml
```diff
@@ -15,6 +15,7 @@ disabled_rules:
   - todo
   - unused_capture_list
   - type_name # tests will have have the format <SUT>_Tests
+  - xctfail_message
```
jaxdesmarais, line 18: I have currently disabled this rule based on [previous PR comment discussion](https://github.com/paypal/iOS-SDK/pull/20#discussion_r731997857). I have disabled it for now as we were getting warnings once the `UnitTest` path was corrected and we were properly running the linter on tests. If someone feels strongly that we should keep this rule, lets use this comment as a jumping off point!

.swiftlint.yml
```diff
@@ -63,6 +64,11 @@ custom_rules:
     capture_group: 1
     message: "Use explicit type annotation when initializing empty arrays and dictionaries"
     severity: warning
+  space_after_main_type:
```
jaxdesmarais, line 67: If someone has a better name for this, we can certainly rename, I had a hard time coming up with what to call this.

.swiftlint.yml
```diff
@@ -63,6 +64,11 @@ custom_rules:
...
+    name: "No space after main type"
+    regex: '(class|struct|extension)((?-s)\s.*\{$\n)(?!^\s*$)'
```
jaxdesmarais, line 69: Currently I only included class, struct, and extension to this rule. Do we also want to include enum in this?

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -1,13 +1,16 @@
 struct CreateOrderParams: Codable {
+
     let intent: String
     let purchaseUnits: [PurchaseUnit]
 }
 
 struct PurchaseUnit: Codable {
+
     let amount: Amount
 }
```
jonathajones: I think this type is a good example of when this rule might lead to weird looking types. The primary use of adding a line after the opening type declaration is to add visual clarity, right? Feel like it might be working against us for short data objects

.swiftlint.yml
```diff
@@ -15,6 +15,8 @@ disabled_rules:
   - todo
   - unused_capture_list
   - type_name # tests will have have the format <SUT>_Tests
+  - xctfail_message
+  - empty_enum_arguments
```
jaxdesmarais: Disabled this rule as it is failing CI. In our unit tests we use this format in several switch cases which triggers this rule:
```
case .failure(_):
```
We can keep this by potentially enabling this rule and passing the argument into the XCTFail message if we feel strongly about keeping these two rules as they go hand in hand a bit.

jonathajones: Changing the `case .failure(_):` to `case .failure:` would satisfy the linter, right? is there a downside to removing the discarded failure type?

Demo/Demo/Models/CreateOrderParams.swift
jaxdesmarais: Potentially in the short term, yes, but I expect [purchase unit for example](https://developer.paypal.com/docs/api/orders/v2/#definition-purchase_unit_request) to get much larger in the future. I agree that it could look weird in some cases for now, but maybe that's a time when we could disable the rule until we feel we need it?

.swiftlint.yml
jaxdesmarais: Nope, that should work as well - I didn't know if there was a reason we had decided to go with `case .failure/.success(_):` since we have that format ~10 times throughout our tests, but if we see no value, I also think it makes sense to remove altogether and revert this rule 😄

.swiftlint.yml
scannillo: Sounds good to me!

Demo/Demo/Models/CreateOrderParams.swift
jonathajones: It feels weird to me to be prescriptive about our code styling in a way that knowingly makes some of our types look wonky. what do y'all think? @scannillo @minhthenguyen @jaxdesmarais

.swiftlint.yml
jonathajones: yeah I think we should follow the linter's advice here and update `.failure(_):` and `.success(_)` to `.failure:` and `.success:`

jaxdesmarais: Yep, already updated! 😄

Demo/Demo/Models/CreateOrderParams.swift
jaxdesmarais: I think, at least for me personally, I'd lean towards having a linter rule for this to catch the places we _should_ have this type of spacing currently and don't if that's the styling we agreed to go with. 

Previously we did have a rule that would check the inverse of this but removed that rule, so seems like a good idea to have a rule for what we actually want? This is also in the demo app, so feels slightly less weird to just disable this rule if we don't like how this looks. 🤷‍♂️ 

But again, that's just my opinion! If everyone feels strongly otherwise, I'd be happy to just close out this PR and open a small one to address the other issues.

scannillo: Hm ... I think the space helps when the file is larger. For Codable structs where are pretty straightforward, it's less helpful.

I don't feel too strongly to be honest. Maybe we could leave the rule as optional, and then just use it when the file is large?

jaxdesmarais: I can also update this rule to not apply to codable/decodable structs if that's the main sticking point. 🤷‍♂️

jaxdesmarais: Actually looking at my above comment again, it looks like most of our other Codable/Decodable structs (such as Address and Card) already use this spacing (really the demo app is the biggest place this may "look weird"). I think I still am not fully understanding why we wouldn't want a linter rule for this but want a linter rule for other similar styling agreements that we have come to? Would love to make a decision one way or the other on this though. @jonathajones @minhthenguyen @scannillo

jaxdesmarais: I am not married to either having/not having this, fwiw. I had some free time during open dev that I wanted to spend understanding more regex and know this styling inconsistency was mentioned in another PR as an item to address, so wanted to try to tackle it since we would need a custom regex rule for it. 😄 More-so just trying to better understand when a rule is needed and when it's not and we are fine with inconsistencies even though we may lean towards a particular style preference.

jonathajones: >I think I still am not fully understanding why we wouldn't want a linter rule for this but want a linter rule for other similar styling agreements that we have come to?

I think for me the difference is that I don't think we would necessarily want to follow this for a data container that is >10 lines long (at least in my opinion). For long files, I have no issue with it - but it visually breaks up really small types in a way that looks weird and makes me think we should maybe not have a rule for it (or at least not one that is enforced for small files. Is swiftlint smart enough to enforce only for longer files?)

jaxdesmarais: That makes sense, sorry I wasn't following so just wanted to clarify. I am not sure if that's possible! Let me dig into it and see what I can find.

scripts/swiftlint.sh
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+if test -d "/opt/homebrew/bin/"; then
+  PATH="/opt/homebrew/bin/:${PATH}"
+fi
+
+export PATH
+
```
jaxdesmarais, line 8: Swiftlint was not playing nice on the M1 that I'm working on because my main laptop 💥 . I found the above solution here: https://www.anotheriosdevblog.com/installing-swiftlint-on-a-m1/.

jonathajones (approved):

Demo/Demo/Models/CreateOrderParams.swift
jaxdesmarais: Synced with Jon yesterday (and Sammy the day before) and it seems like customizing the regex may be more effort than it's worth and potentially difficult to maintain for other edge cases. Jon and I discussed leaving this rule as is for now and disabling it in places where things look weird. This has been updated, but we both agreed that we should probably revisit this in a few months to see how we are feeling about this rule.

minhthenguyen (approved):
this looks good to me :smile:

PR #25, opened 2021-10-25
Paypal integration + demo

minhthenguyen:
### Reason for changes

We need to integrate with PayPalCheckout and provide support for PayPal paysheet to handle PayPal transaction

### Summary of changes

- Add `PayPalCheckout` as dependency for `PayPal` module
- Add `PayPalError` and `PayPalResult`
- Add `PayPalPaysheet` and `PayPalCompletionState`

#### Sample integration
```swift
class MerchantViewController: UIViewController {
    let config = CoreConfig(clientID: "clientID", environment: .sandbox)
    let paypalClient = PayPalClient(config: coreConfig, returnUrl: "returnUrl")

    func checkoutWithPayPal(orderID: String) {
        paypalClient.start(presentingViewController: self, orderID: orderID) { state in
            switch state {
                case .success(let result):
                    print("Order is successfully approved and ready to be captured/authorized")
                case .failure(let error):
                    print("There's an error: \(error)")
                case .cancellation:
                    print("Buyer has canceled checkout")
            }
        }
    }
}
```

### Checklist

- [ ] Added a changelog entry

### Authors
- @scannillo 
- @jonathajones 
- @jaxdesmarais 
- @minhthenguyen


scannillo:
I think we want to follow the `Client` naming pattern, as we've discussed in our docs-driven dev discussions in person (also see [here](https://github.com/paypal/paypal-sdk-spec/pull/48/files#r719542037) for a similar thread).

So that would look like `PayPalClient(config: MyConfig)` to align with the Card integration point, and act as the main point of entry into the PayPal module.

Sources/PayPal/UI Components/PayPalUIDelegate.swift
```diff
@@ -0,0 +1,17 @@
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+public protocol PayPalUI: AnyObject {
+    var delegate: PayPalUIDelegate? { get set }
+}
+
+/// Methods for managing approval, error and cancellation from PayPal UI components
+public protocol PayPalUIDelegate: AnyObject {
```
scannillo: There was debate/discussion on using a delegate vs callback pattern here. I know Jon, David, and I had discussed it a lot together.

I thought where we left off after chatting with David was to use the callback pattern with the main take-home point being it allows our integrations across payment methods to be the most consistent / easier to document & then more nuanced hooks into UI flows would happen via an optional delegate. @jonathajones - was that also your understanding?

Sources/PaymentsCore/CoreConfig.swift
```diff
@@ -3,6 +3,16 @@ import Foundation
 /// The configuration object containing information required by every payment method.
 /// It is used to initialize all Client objects.
 public struct CoreConfig {
+
     public let clientID: String
     public let environment: Environment
+
+    /// Return URL required for PayPal integration
+    public let returnUrl: String?
```
scannillo: Do y'all have any updates on the investigation to remove this requirement?

scannillo: Also, if this is only required for the PayPal checkout, would make more sense to be a param to initialize a `PayPalClient`

Package.swift
```diff
@@ -20,7 +20,7 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        // .package(url: /* package url */, from: "1.0.0"),
+        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .exact("0.73.0"))
```
scannillo: I think we can do _up to next major version_ instead of exact version inclusion here.

Sources/PaymentsCore/CoreConfig.swift
minhthenguyen: good call, I'll move this to `PayPalClient`.

> Do y'all have any updates on the investigation to remove this requirement?

I remember we won't be able to remove this for now but I can't remember the reason. Maybe Tim or Jon will have better context than me on this.

Package.swift
minhthenguyen: I think dependencies may have breaking changes across versions causing our integration working in one but fail in another version, so maybe an exact version might be more beneficial? What do you think?

scannillo: The Native XO SDK doesn't follow semantic versioning?

minhthenguyen: our major version is still at 0, and I know we've introduced breaking changes in minor versions before :smile:

jonathajones: Yeah just to clarify - since the native SDK is technically prerelease (currently major version 0), any API changes would result in a minor version bump. I think it's probably good to pin to exact versions for now, until the NXO sdk goes to 1.0. Not sure when that will be - would need to discuss with the NXO team.

Sources/PayPal/UI Components/PayPalUIDelegate.swift
jonathajones: I think that convo tapered off more than completely settled, but i think the general feeling was to use a single callback for approve / error / cancel, and pass back an enum with `success`, `failure`, and `cancellation` cases. I think we can go with this (the enum parameter in a callback) approach and then in the future (when we add support for shipping callbacks / creating orders on the client side), we can re-evaluate how the interface feels then

Sources/PayPal/Enums/PayPalError.swift
```diff
@@ -0,0 +1,46 @@
...
+extension ErrorInfo {
+
+    func toPayPalSDKError() -> PayPalSDKError {
+        PayPalSDKError(
+            code: PayPalError.Code.payPalCheckoutError.rawValue,
+            domain: PayPalError.domain,
+            errorDescription: reason
+        )
+    }
+}
```
jonathajones: instead of extending a native checkout type, could we instead follow a similar pattern that we use for card errors and add this as a static closure that vends a `PayPalSDKError`? 
```swift
    static let payPalCheckoutError: (ErrorInfo) -> PayPalSDKError = { errorInfo in
        PayPalSDKError(
            code: PayPalError.Code.payPalCheckoutError.rawValue,
            domain: PayPalError.domain,
            errorDescription: errorInfo.reason
        )
    }
```

it's helpful i think to have them all declared the same way like this, just for consistency' sake

Sources/PayPal/Models/PayPalResult.swift
```diff
@@ -0,0 +1,21 @@
...
+import Foundation
+
+/// The result of a PayPal payment flow.
+public struct PayPalResult {
+
+    /// The order ID associated with the transaction.
+    public let orderID: String
+
+    /// Intent, this will be `CAPTURE` or `AUTHORIZE` if the order token was generated using the `v2/` API.
+    public let intent: String
```
jonathajones: We specify possible values for the v2 api - should we specify what the values would be if a merchant is using v1/payments?

PayPal.xcodeproj/project.pbxproj
```diff
@@ -181,6 +199,31 @@
 			path = UnitTests;
 			sourceTree = "<group>";
 		};
+		0649326C2727014600FA83A6 /* UI Components */ = {
```
jonathajones: Could we remove the space here? to keep it consistent with the rest of our directory names

Package.swift
scannillo: Gotcha, thanks for clarifying where that stands!

I also think there is a way to specify "up to the next minor version release", which sounds like would work while NXO is on v0.

Sources/PayPal/Models/PayPalResult.swift
scannillo: I _think_ we can ignore v1 payments cases for now

Sources/PayPal/UI Components/PayPalUIDelegate.swift
minhthenguyen: ah I see. Thank you both for providing context and clarifying this!!! 
I'll make this change :smile:

Sources/PaymentsCore/CoreConfig.swift
jonathajones: I don't think i have any additional context - the conversation was brought up with the Identity team, and it's a requirement to enable the auth SDK to perform its function. So we have to live with it this way for the time being - though maybe it's possible to abstract this away from the merchant somehow? worth a discussion at some point.

Sources/PayPal/Enums/PayPalError.swift
```diff
@@ -0,0 +1,34 @@
...
+
+        /// 1. Error returned from the PayPal Checkout SDK
+        case payPalCheckoutError
+    }
+
+    static let unknown = PayPalSDKError(
+        code: Code.unknown.rawValue,
+        domain: domain,
+        errorDescription: "An unknown error has occured."
+    )
```
scannillo: We can delete this, as it isn't being used.

Sources/PayPal/Extensions/ApprovalData+Extension.swift
```diff
@@ -0,0 +1,11 @@
...
+import PayPalCheckout
+#endif
+
+extension ApprovalData {
+
+    /// Convert `PayPalCheckout.ApprovalData` to `PayPalResult`
+    func toPayPalResult() -> PayPalResult {
+        PayPalResult(orderID: ecToken, intent: intent, payerID: payerID)
+    }
+}
```
jonathajones: I think extending external types is a pattern we maybe want to generally avoid - i think it makes it less clear where methods are coming from (when i read where this is used, i would assume that we are invoking a function internal to the paypal checkout module if i didn't know about this extension already)

I feel like instead, having a convenience initializer on our internal type that takes an instance of `ApprovalData` would be clearer. what do yall think? want to hear other opinions on this

Sources/PayPal/UIComponents/PayPalClient/PayPalClient.swift
```diff
@@ -0,0 +1,56 @@
...
+
+/// PayPal Paysheet to handle PayPal transaction
+public class PayPalClient {
+
+    private let config: CoreConfig
+    private let returnURL: String
+
+    lazy var paypalCheckoutConfig: CheckoutConfig = {
+        CheckoutConfig(clientID: config.clientID, returnUrl: returnURL)
+    }()
```
jonathajones: do we need to account for the environment here as well?

Sources/PayPal/UIComponents/PayPalClient/PayPalClient.swift
```diff
@@ -0,0 +1,56 @@
...
+    public init(config: CoreConfig, returnURL: String) {
+        self.config = config
+        self.returnURL = returnURL
+    }
+
+    /// Present PayPal Paysheet and start a PayPal transaction
+    /// - Parameters:
+    ///   - presentingViewController: the ViewController to present PayPalPaysheet on
+    ///   - orderID: the order ID for the transaction
+    public func start(presentingViewController: UIViewController, orderID: String, completion: PayPalCheckoutCompletion? = nil) {
```
jonathajones: I believe the `UIViewController` param can be optional, since if it doesn't exist, the NXO SDK will crawl through the key window to find a VC to present on.

Sources/PayPal/UIComponents/PayPalCompletionState.swift
```diff
@@ -0,0 +1,16 @@
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+/// The completion state of a PayPal checkout experience
+public enum PayPalCompletionState {
```
jonathajones:
```suggestion
public enum PayPalCheckoutResult {
```
what do you think?

Sources/PayPal/Extensions/ApprovalData+Extension.swift
minhthenguyen: ah yep, I wasn't thinking about `init`.
I do like that better than `extension` :smile:

Sources/PayPal/Enums/PayPalError.swift
```diff
@@ -0,0 +1,28 @@
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+#if canImport(PayPalCheckout)
+import PayPalCheckout
+#endif
```
jonathajones: If we want to create all our own types for our public interface, we might use `@_implementationOnly` imports (https://forums.swift.org/t/update-on-implementation-only-imports/26996/6) 

This would prevent us from exposing any NXO types in our public interface, which could be a helpful form of enforcement

scannillo: I love this!

scannillo: Now that I think about it - CocoaPods & SPM don't use that `import` line, our Demo app does / Carthage people would if we add Carthage support.

scannillo: I still think it is worth keeping the `@_implementationOnly` import, but just wanted to note that CocoaPods & SPM are doing their own thing in terms of potentially exposing NXO details. We could double check with each of those package managers to see if we have any customizability on that front, though I haven't come across it before.

Sources/PayPal/Enums/PayPalError.swift
```diff
@@ -0,0 +1,28 @@
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+#if canImport(PayPalCheckout)
+@_implementationOnly import PayPalCheckout
```
scannillo: Another slightly side note - we _could_ put these all under the `#if canImport(PaymentsCore)` level if we wanted to make it a little easier to read. 

```
#if canImport(PaymentsCore)
import PaymentsCore
@_implementationOnly import PayPalCheckout
#endif
```

Sources/PayPal/UIComponents/PayPalClient/PayPalClient.swift
```diff
@@ -0,0 +1,61 @@
...
+
+    lazy var paypalCheckoutConfig: CheckoutConfig = {
+        CheckoutConfig(
+            clientID: config.clientID,
+            returnUrl: returnURL,
+            environment: config.environment.toPayPalCheckoutEnvironment()
+        )
+    }()
+
+    public init(config: CoreConfig, returnURL: String) {
```
scannillo: We should add a docstring for this initializer, with specific info on what the returnURL is.

scannillo: This is what we had written for when the BT SDK was going to pull in NXO, if we want to pull from this language:

```
    /**
     Initializes a PayPal Checkout request.
     - Parameter payPalReturnURL: The return URL provided to the PayPal Native UI experience.
     Used as part of the authentication process to identify your application. This value should match the one set in the `Return URLs` section of your application's dashboard on your [PayPal developer account](https://developer.paypal.com)
     - Returns: A PayPal Checkout request.
     */
```

Sources/PayPal/UIComponents/PayPalClient/PayPalClient.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.config = config
+        self.returnURL = returnURL
+    }
+
+    /// Present PayPal Paysheet and start a PayPal transaction
+    /// - Parameters:
+    ///   - orderID: the order ID for the transaction
+    ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
+    ///   - completion: Completion block to handle buyer's approval, cancellation, and error.
+    public func start(orderID: String, presentingViewController: UIViewController? = nil, completion: PayPalCheckoutCompletion? = nil) {
```
scannillo: 👏

Sources/PayPal/UIComponents/PayPalClient/PayPalClient.swift
```diff
@@ -0,0 +1,61 @@
...
+        setupPayPalCheckoutConfig(orderID: orderID, completion: completion)
+        Checkout.start(presentingViewController: presentingViewController)
+    }
+
+    func setupPayPalCheckoutConfig(orderID: String, completion: PayPalCheckoutCompletion? = nil) {
```
scannillo: I think we can make this private. Also I notice we have unit tests for this method. I think in general, we don't want unit tests for private methods directly - especially if we don't have to. We should instead test the public interface/methods, which then we can assume that our private methods are working correctly.

So in this case, we should test on the `start` method directly.

There's a decent convo about this under `Motivation` in [this article.](https://anthonysciamanna.com/2016/02/14/should-private-methods-be-tested.html)

UnitTests/PayPalTests/PayPalEnvironment_Tests.swift
```diff
@@ -0,0 +1,14 @@
+import XCTest
+@testable import PaymentsCore
+@testable import PayPal
+
+final class PayPalEnvironment_Tests: XCTestCase {
```
scannillo: Do we need `final`? I think we're probably fine without it for our test suite. But if we _really did_ want to add it, we should make it consistent across all test files.

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -0,0 +1,31 @@
+import XCTest
+import PayPalCheckout
+@testable import PaymentsCore
+@testable import PayPal
+
+final class PayPalClient_Tests: XCTestCase {
+
+    func testPayPalClient_whenPayPalCheckoutOnCancelInvoked_completionCalledWithCancellationState() throws {
```
scannillo: This should be a test for the `start` method instead, not the internal `setupPayPalCheckoutConfig` method.

Also since the file is already for `PayPalClient` the test method name can be more specific to the method we're testing. Example: `testStart_when....`

Package.swift
scannillo:
```suggestion
        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .upToNextMinor("0.73.0"))
```

Reference - https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html

scannillo:
General comment - it's a little hard to decipher `PayPalCheckout` references as to the NXO binary, or to the `PayPalCheckout` public API that our SDK is exposing. We could consider calling the NXO binary related functionality something different internally to make this distinction more clear. Maybe something like `CheckoutSDK`? Thoughts?

Sources/PayPal/Enums/PayPalError.swift
jonathajones: Yeah let's test this! If nothing else, it will help enforce a separation between NXO interfaces and Northstar interfaces, since it will throw a compile time error if you try to expose an NXO type while using an `_implementationOnly` import

UnitTests/PayPalTests/PayPalEnvironment_Tests.swift
jonathajones: yeah +1 on removing it - `final` is only helpful when you never want anyone to be able to subclass your type.

Sources/PayPal/Models/PayPalResult.swift
```diff
@@ -0,0 +1,24 @@
...
+
+import Foundation
+
+/// The result of a PayPal payment flow.
+public struct PayPalResult {
+
+    /// The order ID associated with the transaction.
+    public let orderID: String
+
+    /// Intent, this will be `CAPTURE` or `AUTHORIZE` if the order token was generated using the `v2/` API.
```
scannillo, line 13:
```suggestion
    /// Intent, this will be `CAPTURE` or `AUTHORIZE`.
```

Sources/PayPal/Models/PayPalResult.swift
```diff
@@ -0,0 +1,24 @@
...
+    public let intent: String
+
+    /// Payer ID is also the user id associated with the transaction.
```
scannillo:
```suggestion
    /// The Payer ID (or user id) associated with the transaction.
```

Sources/PayPal/Models/PayPalResult.swift
scannillo: I don't think we need INTENT. We _could_ have a separate convo about it ... but we aren't including it on the `CardResult` (https://github.com/paypal/iOS-SDK/blob/main/Sources/Card/Models/CardResult.swift). A merchant will have access to the intent themselves since they create the order on their own.

Sources/PayPal/Models/PayPalResult.swift
```diff
@@ -0,0 +1,24 @@
+#if canImport(PayPalCheckout)
+@_implementationOnly import PayPalCheckout
+#endif
+
+import Foundation
+
+/// The result of a PayPal payment flow.
+public struct PayPalResult {
```
scannillo, line 6: We want to add `orderStatus` to `PayPalResult`:

```
    /// The current status of the transaction
    public let status: OrderStatus
```

Sources/PaymentsCore/CoreConfig.swift
scannillo: Sure - we can put a pin in this. We should definitely investigate trying to hide this from the merchant - we would need to work with the team behind the dashboard (auth/identity?) to figure out if we can maybe use 1 URL for every merchant instead.

scannillo: I also just don't really understand how / why this URL is used on the backend side. Maybe we can do a pow-wow on it soon and figure out who to talk to next.

jonathajones: Yeah this is used by identity to verify that the url provided matches what exists in the dev portal to avoid malicious redirects. This doesn't really apply to an iOS environment - and we could absolutely use a single `returnURL` for every merchant (we would need to set this value from the BE when the user is integrating with / opting into the NXO SDK). 

On Android, this would be more difficult, since the chrome custom tab they are using actually redirects to the specified returnURL (rather than just dismissing, which is what the auth sdk does natively on iOS) - so the return url has to be their app identifier (@tdchow could articulate this better)

Package.swift
scannillo: @jaxdesmarais gave me a good summary about the state of NXO's testing setup and needing to build for 1st & 3rd party. Sometimes builds for 1st party breaks 3rd party integrations, and that using `exact` will be a safer bet at this point.

Sources/PaymentsCore/CoreConfig.swift
jonathajones: I'm all for obscuring this from the merchant however we can. It's been the source of a lot of integration issues for merchants.

Sources/PayPal/Models/PayPalResult.swift
minhthenguyen: currently the PayPalCheckout SDK uses our internal GQL layer which doesn't have OrderStatus, but has this:
```
enum CartState {
CREATED
APPROVED
FAILED
SUBMITTED
}
```
Currently we're not fetching this in PayPalCheckout so it's not possible to get this for now.
But I believe our GQL approve query + `CartState` here are probably wrapper around Orders v2 + `OrderStatus`. I can investigate this and make this addition, but maybe it should be a separate ticket? :smile:

UnitTests/PayPalTests/PayPalClient_Tests.swift
minhthenguyen: About this and https://github.com/paypal/iOS-SDK/pull/25#discussion_r740329241
I'm not sure what is a good way to unit test `start`, since `Checkout.start()` is an UI experience + the flow involves network requests + user actions required.
So I decided to split the callbacks into another function `setupPayPalCheckoutConfig`. imo from our standpoint we just wanna test our callbacks being invoked when PayPalCheckout's callbacks are invoked.

Sources/PayPal/Models/PayPalResult.swift
minhthenguyen: oh just saw you removed `OrderStatus`. Good idea :smile:

Sources/PayPal/Models/PayPalResult.swift
```diff
@@ -0,0 +1,15 @@
...
+
+    /// The order ID associated with the transaction.
+    public let orderID: String
+
+    /// The Payer ID (or user id) associated with the transaction.
+    public let payerID: String
+}
```
jcnoriega, line 13: @minhthenguyen what has been decided on the other fields that `ApprovalData` provides? like `Buyer`, `Payee`, etc etc, becasue i added them on Android

minhthenguyen: just my opinion but I think for now we'll just have `orderID` and `payerID`.
@scannillo @jaxdesmarais @jonathajones is there any other param you think we should add here?

UnitTests/PayPalTests/PayPalClient_Tests.swift
sshropshire: I agree with @scannillo it'd be more coverage complete if we can test the `start` method.

Sources/PayPal/Models/PayPalResult.swift
jaxdesmarais: I don't think anything else is needed currently, imo. We can always add things as we see fit, but I think it'll be better to keep the interface as simple as possible for now and iterate on it as merchants ask for certain elements or we see a need. On the NXO side I think Android today offers more data in the ApprovalData object than we do on the iOS side.

jcnoriega: yes, we have been asked by merchants on the Android github to include more fields. So maybe here we can diverge? idk, but it feels odd to reduce the info we send to the merchant, when we are going to get those complaints again.

sshropshire: I think it's better to reduce the scope of the Android PR. The iOS PR is much smaller than the Android one. At this point a reduction in scope is a positive.

scannillo: +1 - we also confirmed this with David that we want to keep this pretty streamlined and basic right now across iOS and Android and then iterate later on once we get actual merchant feedback. 

Jose, I understand your concern and it is good to keep merchant feedback in mind - but the Northstar SDK right now shouldn't be a "kitchen sink" and offer everything that the NXO SDK already does. We also talked with David & Rahul on this recently. The current goal is to offer what the BT SDK already does, and then go from there after merchant feedback. The BT SDK doesn't give back much more advanced details. And if we need them, then we definitely would need them on iOS and Android.

So in summary, we should go ahead with reducing the Android PR 👍 .

Sources/PayPal/PayPalClient.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    func setupPayPalCheckoutConfig(orderID: String, completion: PayPalCheckoutCompletion? = nil) {
+        paypalCheckoutConfig.createOrder = { action in
+            action.set(orderId: orderID)
+        }
+
+        paypalCheckoutConfig.onApprove = { approval in
+            let payPalResult = PayPalResult(orderID: approval.data.ecToken, payerID: approval.data.payerID)
+            completion?(.success(result: payPalResult))
+        }
```
sshropshire: qq: Will `onShippingChange` use the same completion handler?

minhthenguyen: I think it's gonna be a separate delegate function or listener, since the flow hasn't completed when user changes shipping so probably we shouldn't treat it as a completion.

scannillo:
> From @minhthenguyen on Slack. I'm not sure I get what you mean in this comment yet: https://github.com/paypal/iOS-SDK/pull/25#issuecomment-956380295
can you explain a bit more?

Yes, so we're calling both the Northstar public facing PayPal things `PayPalCheckout` and we're also calling our internal implementation of the NXO SDK `PayPalCheckout`. As developers, it makes it harder to distinguish what is Northstar code vs what is NXO-necessary implementation code. Maybe this is harder for me since I don't work on NXO. But does that make sense?

I think if our internal variable names for `NXO` implementation were something like `nativeCheckoutSDK` this would be easier to read.

Sources/PayPal/Models/PayPalResult.swift
jcnoriega: @scannillo on it.

Sources/PayPal/Enums/PayPalError.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    enum Code: Int {
+        /// 0. An unknown error occurred.
+        case unknown
+
+        /// 1. Error returned from the PayPal Checkout SDK
+        case nativeCheckoutSDKError
+
+        /// 2. The buyer canceled the PayPal Checkout flow
+        case cancelled
```
jonathajones: I don't think I agree with this being an error scenario. Users cancel the checkout flow all the time for any number of reasons - them doing so shouldn't be regarded as an error

jonathajones: Merchants might very well record errors generated by our SDK to their own analytics platform, or make a UI change in response to a payment error that they would want to suppress if it's a checkout cancellation

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+import PayPalCheckout
+@testable import PaymentsCore
+@testable import PayPal
+
+class PayPalClient_Tests: XCTestCase {
+
+// TODO: Put up a separate PR with Mocks for NXO so we can test the below things
```
scannillo: Jax & I spent a solid amount of time trying to mock the `CheckoutConfig`, which we were eventually able to sort of get working via a protocol. Mocking the `Checkout.set` is hard to test since this is a static method on a class, and not on an instance of a class. 

So .... there's some work to be done here. Either on the NXO side making an API that is more easily testable, or spending more time trying to come up with a way to mock this out, it will be wonky though. I am not sure how we'd do it.

I think it warrants a larger convo around NXO testing strategy & setup, and how that affects the Northstar SDK.

jonathajones: Yeah NXO is not built in a way that makes it super easy to test. I spent some time hacking on this - i think this might work for us for now, to let us test / mock out the functionality of the `Checkout` type:

```swift
protocol PayPalUIFlow {

  static func set(config: CheckoutConfig)

  static func start(
    createOrder: CheckoutConfig.CreateOrderCallback?,
    onApprove: CheckoutConfig.ApprovalCallback?,
    onShippingChange: CheckoutConfig.ShippingChangeCallback?,
    onCancel: CheckoutConfig.CancelCallback?,
    onError: CheckoutConfig.ErrorCallback?
  )
}

extension Checkout: PayPalUIFlow { }

class PayPalClient2 {
  let CheckoutFlow: PayPalUIFlow.Type

  init(flow: PayPalUIFlow.Type) {
    self.CheckoutFlow = flow
  }

  func configCheckout() {
    CheckoutFlow.set(config: CheckoutConfig(clientID: "", returnUrl: ""))
  }

  func startCheckout() {
    CheckoutFlow.start(
      presentingViewController: nil,
      createOrder: nil,
      onApprove: nil,
      onShippingChange: nil,
      onCancel: nil,
      onError: nil
    )
  }
}

class CheckoutMock: PayPalUIFlow {
  static func set(config: CheckoutConfig) {
    // no-op
  }

  static func start(presentingViewController: UIViewController?, createOrder: CheckoutConfig.CreateOrderCallback?, onApprove: CheckoutConfig.ApprovalCallback?, onShippingChange: CheckoutConfig.ShippingChangeCallback?, onCancel: CheckoutConfig.CancelCallback?, onError: CheckoutConfig.ErrorCallback?) {
    // no-op
  }
}


let payPalClient = PayPalClient2(flow: Checkout.self)

let payPalMockClient = PayPalClient2(flow: CheckoutMock.self)

```

Sources/PayPal/Enums/PayPalError.swift
jaxdesmarais: I keep going back and fourth on this, to be honest. I can see an argument for it being it's own case since it's not a success or error but also can see an argument for it falling into the error bucket as it's an unsuccessful completion of checkout. I think BT handles cancellations as an error today whereas Stripe has a cancelled case as we did previously (though theirs is called "user cancelled").

UnitTests/PayPalTests/PayPalClient_Tests.swift
scannillo: Awesome, yeh this makes sense!

scannillo: We also might have to mock `CheckoutConfig` but what you have above is most of the work.

Sources/PayPal/Enums/PayPalError.swift
scannillo: Yep that is right - BT currently returns an error. I mean on one hand, merchants have been using that integration pattern for years and it hasn't been an issue. On the other hand, this doesn't mean it couldn't better, especially now that we can use Swift enums.

This is the _simpler_ way, since it doesn't require a custom Result type. Hmmmm .... I see both sides as well.

Some food for thought - what would Result types look like for other payment methods? I would imagine similar here: success, cancel, result.

Sources/PayPal/PayPalClient.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    /// Initialize a PayPalClient to process PayPal transaction
+    /// - Parameters:
+    ///   - config: The CoreConfig object
+    ///   - returnURL: The return URL provided to the PayPal Native UI experience. Used as part of the authentication process to identify your application. This value should match the one set in the `Return URLs` section of your application's dashboard on your [PayPal developer account](https://developer.paypal.com)
+    public convenience init(config: CoreConfig, returnURL: String) {
+        self.init(config: config, returnURL: returnURL, checkoutFlow: Checkout.self)
+    }
+
+    init(config: CoreConfig, returnURL: String, checkoutFlow: PayPalUIFlow.Type) {
```
scannillo: In the `APIClient.swift` we have the internal vs public initializers structured a little differently.  They both achieve the same thing, but we should stick to one pattern and keep it consistent throughout the repo.

Sources/PayPal/PayPalClient.swift
```diff
@@ -0,0 +1,79 @@
...
+        self.returnURL = returnURL
+        self.CheckoutFlow = checkoutFlow
+    }
+
+    /// Present PayPal Paysheet and start a PayPal transaction
+    /// - Parameters:
+    ///   - orderID: the order ID for the transaction
+    ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
+    ///   - completion: Completion block to handle buyer's approval, cancellation, and error.
+    public func start(
```
scannillo, line 40: I think we can come up with a better name for this method. Right now we have `CardClient.approveOrder` and `PayPalClient.start` which seem disjointed.

Maybe something like `CardClient.checkout` / `PayPalClient.checkout`? We should start thinking of some good verbage / names to use throughout the SDK.

jonathajones: I agree with this, and i think we should add some cohesion here between the modules.  the `CardClient` and the `PayPalClient` do different things - one is basically a wrapper for an API call, while the other initiates a user experience flow. The result ends up being the same, but do you think that we should make it clear that these are different sorts of functions in their names?

Sources/PayPal/Enums/PayPalError.swift
jonathajones: I think the result type would look different for other payment methods, since the idea of a user initiated cancellation doesn't really exist for flows that do not involve user interaction. For an API wrapping function, we will always have a successful response or a failure response - so having `success/failure` makes sense in taht context, but it feels like it would be sending the wrong message to the merchant to say that a user cancellation is an error. 

fwiw the JS SDK also treats cancellations separately from errors

Sources/PayPal/PayPalClient.swift
jonathajones: yeah removed the `convenience` init. May be worth revisiting in the future though to refactor that pattern to use the convenience inits though - to keep our initialization logic in a single place and avoid the chance that we leave something out of one of our initializers, since all of the public facing convenience inits would call one of our required (and tested) inits under the hood.

Sources/PayPal/PayPalUIInterfaces/PayPalUIFlow.swift
```diff
@@ -0,0 +1,52 @@
+import Foundation
+import UIKit
+@_implementationOnly import PayPalCheckout
+
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+protocol PayPalUIFlow {
```
scannillo: So these are only needed for testing purposes. For me, it's a little confusing managing both the name of the protocol, and the names of the implementation classes that it wraps / abstracts.

This is a pattern we use in `APIClientProtocol` and are sure to need / use often again in our SDK. Can we come up with a consistent naming pattern for these protocols - whose intent is just to abstract for testing?

I think `<NameOfThing>Protocol` or `<NameOfThing>Wrapper` something like that, would help with consistency and readability.

jonathajones: that's a good callout. I'll update to be `CheckoutProtocol`

Sources/PayPal/PayPalUIInterfaces/PayPalUIFlow.swift
```diff
@@ -0,0 +1,52 @@
...
+        onError: ErrorCallback?
+    )
+}
+
+extension Checkout: PayPalUIFlow {
+    static func set(config: CoreConfig, returnURL: String) {
+        let nxoConfig = CheckoutConfig(
+            clientID: config.clientID,
+            returnUrl: returnURL,
+            environment: config.environment.toNativeCheckoutSDKEnvironment()
```
scannillo: Did you mess around with removing `set` from the `PayPalUI` protocol and this extension? Ideally `config.environment.toNativeCheckoutSDKEnvironment()` (business logic) doesn't live in this extension, but rather in `PayPalClient`. 

I wouldn't be surprised if this is something you explored, but ran into issues abstracting the `CheckoutConfig` initializer. I _think_ we could achieve this by also abstracting `CheckoutConfig` in a similar way to the above - but would need to mess with it to be sure.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -1,10 +1,76 @@
...
+    // MARK: - PayPal Module Integration
+
+    @objc func didTapPayPalButton() {
+        checkoutWithPayPal(orderID: orderID ?? "")
+    }
+
+    func checkoutWithPayPal(orderID: String) {
+        let payPalClient = PayPalClient(config: config, returnURL: "")
+        payPalClient.start(orderID: orderID, presentingViewController: self) { [weak self] state in
+            guard let self = self else { return }
```
notoriousmarcos, line 63: You can avoid this guard if the below implementation will be always like that, what do you think?

Sources/PayPal/PayPalClient.swift
scannillo: That's a good question 🤔 . I'll let it marinate. -  I think we should talk with the Android team too on this.

scannillo: My gut says, to go with the simplest thing first, so using the same phrasing/names since the end result is the same.

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -0,0 +1,72 @@
+import XCTest
+import PayPalCheckout
+@testable import PaymentsCore
+@testable import PayPal
+
+class PayPalClient_Tests: XCTestCase {
```
notoriousmarcos, line 6: Do we have any reason to separate the Tests word with an underscore?

scannillo: It makes sifting through a soon-to-be very large suite of test files a lot easier visually. So, solely for readability & organization.

Sources/PayPal/Enums/PayPalError.swift
jaxdesmarais: It looks like [Android also treats cancellations separate from error or success](https://github.com/paypal/Android-SDK/blob/d2768255e848507cba91892e1a3c867f65c2bcf9/PayPalCheckout/src/main/java/com/paypal/android/checkout/PayPalCheckoutResult.kt). Anyone opposed to us going back to that?

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: Yeah we could just do `self?...` self when it's used below, but I think adding a guard here is slightly more safe. At the end of the day though I think this may just come down to preference. 🤷‍♂️

Sources/PayPal/Enums/PayPalError.swift
scannillo: Nope, that sounds good to me!

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
scannillo: Yeh - especially since this is just for the Demo app, it doesn't matter a ton. I agree the guard is safer.

Sources/PayPal/PayPalUIInterfaces/PayPalUIFlow.swift
jaxdesmarais: I was updating some other things so updated this as well! 😄

jonathajones: thank you jax!

minhthenguyen:
As discussed with @jcnoriega and Rahul:
Rahul mentioned we should add payer's info and billing/shipping address to `PayPalResult`.

Exposing these info in `NXO.ApprovalData` requires changes from NXO side
so I created a separate ticket for that and it will be addressed in another PR: https://engineering.paypalcorp.com/jira/browse/DTNOR-267

jaxdesmarais (approved):
Took another pass at this PR and it looks good!

Sources/PayPal/PayPalUIInterfaces/PayPalUIFlow.swift
jonathajones: Yeah i wanted to avoid making the protocol generic - so i had to conform the `CheckoutConfig` type to a protocol that only involved core types. I feel like the `toNativeCheckoutSDKEnvironment` conversion isn't so much business logic as it is interface conformance similar to the other conformances in the `PayPalUIInterfaces` directory. what do you think? (sorry for my delay getting back to this)

jonathajones (approved):
I think this looks good! I think we can handle the naming alignment + consistency things in a follow up PR if that works for everyone else

PR #26, opened 2021-11-16
Paypal and PayPal Credit Buttons

jaxdesmarais:
### Reason for changes

Add static PayPal and PayPal Credit Buttons to the SDK. This allows merchants to easily render the buttons while offering the ability to fully customize the experience

### Summary of changes

- Main `PaymentButton` class as a space to customize all buttons and expand on in the future as button options are added
- Add color and image assets for the standard PayPal and PayPal Credit Buttons
- Add `PayPalButton` and `PayPalCreditButton` classes to configure individual buttons and allow us to offer different button options in the future
- Add buttons to the Demo app when PayPal is selected - also provides an example for how merchants can customize the buttons and add a target to launch the PayPal Checkout flow

👀 👀 👀 

Simulator Screen Recording - iPhone 12 - 2021-11-16 at 14 20 03

### Checklist

- @jaxdesmarais @minhthenguyen


Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -0,0 +1,24 @@
...
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Deinit
+
+    deinit { }
```
scannillo:
```suggestion
```

Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+
+// Configuration for PayPal Credit button
```
scannillo:
```suggestion
// Configuration for PayPal button
```

Sources/PayPal/Buttons/PayPalCreditButton.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+
+// Configuration for PayPal Credit button
```
scannillo: This looks like a comment and not a docstring?

scannillo: Also seeing this pattern in all of the new files.

jaxdesmarais: Updated!

Sources/PayPal/Buttons/PayPalCreditButton.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit
+
+/// Configuration for PayPal Credit button
+public class PayPalCreditButton: PaymentButton {
+
+    // MARK: - Init
```
scannillo: Probably don't need this mark since there's nothing else in the file

Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -0,0 +1,20 @@
...
+
+    public init() {
+        super.init(frame: .zero)
+
+        translatesAutoresizingMaskIntoConstraints = false
+        backgroundColor = getButtonColor(for: .payPal)
+        setImage(getButtonLogo(for: .payPal), for: .normal)
+    }
+
+    @available(*, unavailable)
```
scannillo: Do we need this available check? I'm not sure what this is doing.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -6,22 +6,23 @@ class PayPalDemoViewController: FeatureBaseViewController {
...
+    let payPalButton = PayPalButton()
+
+    let payPalCreditButton = PayPalCreditButton()
 
     // MARK: - View Lifecycle & UI Setup
 
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        payPalButton.layer.cornerRadius = 4.0
```
notoriousmarcos: What do you think about to keep this view as a lazy var a put this setup inside the closure?

I think is more readable and keep every setup of the view on the same code block and you can keep viewDidLoad more clean. Example:

```
lazy var payPalButton: PayPalButton = {
         let payPalButton = PayPalButton()
         payPalButton.addTarget(self, action: #selector(paymentButtonTapped), for: .touchUpInside)
         payPalButton.layer.cornerRadius = 8
         return payPalButton
 }()
```

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -31,8 +32,13 @@ class PayPalDemoViewController: FeatureBaseViewController {
         NSLayoutConstraint.activate([
             payPalButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
             payPalButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
-            payPalButton.centerYAnchor.constraint(equalTo: view.centerYAnchor),
-            payPalButton.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight)
+            payPalButton.centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: 50),
+            payPalButton.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight),
+
+            payPalCreditButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
```
notoriousmarcos, line 46: Do we have an UIView extension to make more erase to do this kind of setup for constraints? 

Example:

```
extension UIView {
  func anchor(to view: UIView, ...) {}
}
```

Sources/PayPal/Buttons/PaymentButton.swift
```diff
@@ -0,0 +1,52 @@
...
+            return UIImage(named: "PayPalLogo", in: bundleIdentifier, compatibleWith: nil)
+        case .payPalCredit:
+            return UIImage(named: "PayPalCreditLogo", in: bundleIdentifier, compatibleWith: nil)
+        }
+    }
+
+    func setLogoImageSize() -> UIEdgeInsets {
+        let buttonSize = frame.size
+        return UIEdgeInsets(
+            top: buttonSize.height / 4,
```
scannillo: Why divided by 4 here? Maybe we can move this # into a variable for proper naming.

Sources/PayPal/Buttons/PaymentButton.swift
```diff
@@ -0,0 +1,52 @@
...
+            left: buttonSize.width / 4,
+            bottom: buttonSize.height / 4,
+            right: buttonSize.width / 4
+        )
+    }
+
+    // MARK: - Override Function
+
+    /// Once we layout our subviews, we need to update the logo with the button frame
+    public override func layoutSubviews() {
```
scannillo, line 26: This makes more sense to me to live at the top of the file, than our internal helper methods.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -6,22 +6,23 @@ class PayPalDemoViewController: FeatureBaseViewController {
...
-        payPalButton.addTarget(self, action: #selector(didTapPayPalButton), for: .touchUpInside)
-        payPalButton.layer.cornerRadius = 8
-        payPalButton.backgroundColor = .systemBlue
-        payPalButton.tintColor = .white
-        payPalButton.isEnabled = false
-        return payPalButton
-    }()
+    let payPalButton = PayPalButton()
+
+    let payPalCreditButton = PayPalCreditButton()
```
minhthenguyen: maybe we can remove the credit button from the demo for now, as it doesn't link to the credit flow? :smile:

Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -0,0 +1,20 @@
...
+public class PayPalButton: PaymentButton {
+
+    // MARK: - Init
+
+    public init() {
+        super.init(frame: .zero)
+
+        translatesAutoresizingMaskIntoConstraints = false
+        backgroundColor = getButtonColor(for: .payPal)
+        setImage(getButtonLogo(for: .payPal), for: .normal)
```
minhthenguyen: `init()` functions for `PayPalButton` and `PayPalCreditButton` look kinda similar now, what do you think of adding an init for `PaymentButton`:
```
func init(type: PayPalButtonType) {
        super.init(frame: .zero)
        translatesAutoresizingMaskIntoConstraints = false
        backgroundColor = getButtonColor(for: type)
        setImage(getButtonLogo(for: type), for: .normal)
}
```

Then call this function in PayPal and Credit button classes?

Sources/PayPal/Buttons/PaymentButton.swift
```diff
@@ -0,0 +1,52 @@
...
+    func getButtonLogo(for buttonType: PayPalButtonType) -> UIImage? {
+        switch buttonType {
+        case .payPal:
+            return UIImage(named: "PayPalLogo", in: bundleIdentifier, compatibleWith: nil)
+        case .payPalCredit:
+            return UIImage(named: "PayPalCreditLogo", in: bundleIdentifier, compatibleWith: nil)
+        }
+    }
+
+    func setLogoImageSize() -> UIEdgeInsets {
```
minhthenguyen: maybe this can be change to `getLogoImageInsets`? since we just get but don't set the insets here

Sources/PayPal/Buttons/PaymentButton.swift
minhthenguyen: for now we decided to divide by 4 to make the logo half the size of the button (as we tested that was what looks nicest to us)
but probably in the future if we get exact requirement from design/UX then we'll go back in here to update this :smile:

Sources/PayPal/Buttons/PayPalButton.swift
jaxdesmarais: I am not sure if the available check is necessary, but assume it is for something since Xcode defaults this implementation to our class if we remove this init from the class.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: Yep, that works as well as what we have here. I can update to this since this is more similar to what we had previously. 😄

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: I think that's certainly something we can explore updating, but maybe in a future PR? This PR sets the constraints in the same way we do in all of the `ViewControllers`, so I think if we want to make this change we'd want to explore it in a space where we can make this update as a whole vs in one VC?

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: I think it may be helpful to showcase the different buttons we offer, at least for now. As merchants will have full control over the flow, they also don't necessarily _have_ to launch the credit flow (though that's obviously what we'd likely want them to do).

Sources/PayPal/Buttons/PayPalButton.swift
jaxdesmarais: Hmmm, I think it feels a little confusing (at least to me) to call a function init and call it from the sub-classes? Maybe something like `createButton(with type: PayPalButtonType)` would make more sense?

minhthenguyen: hmm imo it seems normal to do so - in line 9 above we also call superclass (UIButton) 's `init` insider subclass (PayPalButton) 's `init` :smile: 
but yea I think either works fine, as long as it reduce code duplication!!!!

jaxdesmarais: It seems like calling the func `init` is not best practice as we get the following warning when doing so `Keyword 'init' cannot be used as an identifier here` since init is a protected name. I can certainly make a function to wrap the other 3 aspects, but that seems a little odd to do this in a func vs a new init.

I think it may make more sense to either make a custom `init` on `PaymentButtons` then call `super.init(type: PaymentButtonType)` on the subclasses. This will make our subclasses really tiny for now until we add more functionality/customization, but maybe that's fine? Thoughts?

jaxdesmarais: Went with the second option and extracted the color and logo into an enum instead of a func to allow more flexibility as we add additional customization options. 😄

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
scannillo: Also, keeping our demo app straightforward per-module is nice.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -31,8 +36,13 @@ class PayPalDemoViewController: FeatureBaseViewController {
         NSLayoutConstraint.activate([
             payPalButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
             payPalButton.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -Constants.layoutSpacing),
-            payPalButton.centerYAnchor.constraint(equalTo: view.centerYAnchor),
-            payPalButton.heightAnchor.constraint(equalToConstant: Constants.textFieldHeight)
+            payPalButton.centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: 50),
```
scannillo:
```suggestion
            payPalButton.centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: 50),
```

Pull out `50` into constant for readability?

Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -0,0 +1,13 @@
+import UIKit
+
+/// Configuration for PayPal button
+public class PayPalButton: PaymentButton {
```
scannillo, line 4: Thought - would it be helpful to differentiate the `PayPalCheckoutButton` from the `PayPalCreditButton`?

UnitTests/PayPalTests/PaymentButton_Tests.swift
```diff
@@ -0,0 +1,25 @@
...
+        XCTAssertEqual(payPalButton.imageView?.image, PaymentButtonImage.payPal.rawValue)
+    }
+
+    func testInit_whenPayPalCreditButtonCreated_hasUIImageFromAssets() {
+        let payPalCreditButton = PayPalCreditButton()
+        XCTAssertEqual(payPalCreditButton.imageView?.image, PaymentButtonImage.payPalCredit.rawValue)
+    }
+
+    func testInit_whenPayPalButtonCreated_hasUIColorFromAssets() {
+        let payPalButton = PayPalButton()
```
scannillo: I think we should try to follow our pattern of file-to-file unit test matching. So for example, we have a file `PayPalButton.swift` so it gets its own `PayPalButton_Tests.swift` and same for `PayPalCreditButton.swift` --> `PayPalCreditButton_Tests.swift`, for consistency.

Sources/PayPal/Buttons/PayPalButton.swift
jaxdesmarais: I wonder if that'd be confusing since merchants would still need to launch the checkout experience on their own and the button is just a static button currently. I do see what you are saying though - I feel like calling the entire SDK PayPal and having a PayPal module / buttons in it is getting kinda confusing. 🤔

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -51,7 +65,7 @@ class PayPalDemoViewController: FeatureBaseViewController {
 
     // MARK: - PayPal Module Integration
 
-    @objc func didTapPayPalButton() {
+    @objc func paymentButtonTapped() {
         checkoutWithPayPal(orderID: orderID ?? "")
```
minhthenguyen: Instead of coalescing orderID to nil, should we do a guard here and update title to be `Failed: missing orderID.` if it's nil to make the behavior more like card demo? :smile:

minhthenguyen (approved):
This looks good to me :smile: 
Just one very minor comment in demo app but I think it's fine either way and definitely not a blocker

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: That's a _great_ callout. Updated! 😄

scannillo (approved):

scannillo:
This looks great 👏

PR #27, opened 2021-12-03
Add SwiftUI Card Demo

jaxdesmarais:
### Reason for changes

I've been spending open dev time learning more about SwiftUI 🥳. I wanted to take a pass at adding a SwiftUI card demo to our demo app to spend some time getting more familiar with SwiftUI and its interoperability with UIKit and it seemed like a good task to do that with. 

I thought it would potentially also be helpful as we build out things in general to show merchants how they could use our SDK with SwiftUI. In the future, we could also test out our UI elements to make sure they are also compatible / easy to use with SwiftUI.

If we don't find this helpful though - no harm no foul. It seemed like a good tangible way for me to get more familiar with SwiftUI and was a good use of open dev time. 😄 

### Summary of changes

- Added a card demo in SwiftUI similar to our UIKit card demo
- Added a UI Framework selector to the settings menu
- Added a `BaseViewModel` to share the order ID, `updateTitle` func, and `processOrder` func across the views (this was needed as the SwiftUI view cannot inherit from our UIViewController that we use for UIKit)
- Minor updates needed to make presenting a SwiftUI view and a UIKit view happen after changing the settings

### Checklist

- @jaxdesmarais


Epreuve:
I like this.  First, because it encourages considering how the API looks in a given integration style and can prompt discussions around improvements, but also because it's a great way to showcase the flexibility.

In that same thought, would it be worth while to somehow showcase both integration styles?  My familiarity with iOS has waned in recent years, but I seem to recall that there are still some who aren't completely bought into SwiftUI yet.  Is is a reasonable assumption that one could translate a single demo example into their integration style of choice (assuming the demo isn't already in it)?

jaxdesmarais:
> I like this. First, because it encourages considering how the API looks in a given integration style and can prompt discussions around improvements, but also because it's a great way to showcase the flexibility.
> 
> In that same thought, would it be worth while to somehow showcase both integration styles? My familiarity with iOS has waned in recent years, but I seem to recall that there are still some who aren't completely bought into SwiftUI yet. Is is a reasonable assumption that one could translate a single demo example into their integration style of choice (assuming the demo isn't already in it)?

Totally agree @Epreuve - technically this demo is a combo of both integration styles, so it's:
1. Bottom of ZStack - Our FeatureBaseViewController (which is a UIViewController / UIKit) and houses the create order text field + button, as well as the bottom label that updates based on order ID / status
2. Top of ZStack - a SwiftUI card form

A bit confusing to describe, so idk if that makes sense, but if ya wanna chat more about it happy to find some time for us to hop on a quick call.

If a merchants was to have a _pure_ SwiftUI app, they would need to remove the following in `SwiftUICardDemo`:
```
FeatureBaseViewControllerRepresentable(baseViewModel: baseViewModel)
```

`FeatureBaseViewControllerRepresentable` is the abstraction layer that turns our `FeatureBaseViewController` into a [`UIViewControllerRepresentable`](https://developer.apple.com/documentation/swiftui/uiviewcontrollerrepresentable) which allows us to present that view as an object in our SwiftUI interface

Epreuve:
The more I think about it, the more I like your approach.  

Given that it shows you can use either/or & mixed integrations which I have to think is probably more true to life for most devs given the current state of Apple UI, it speaks to the potential and really sets a good foundation and minimum functional viability as we build up.

Demo/Demo/ViewController.swift
```diff
@@ -29,13 +30,30 @@ class ViewController: UIViewController {
...
+        case .uikit:
+            let demoViewController = DemoSettings.demoType.viewController
+            demoViewController.modalPresentationStyle = .fullScreen
+
+            navigationItem.title = DemoSettings.demoType.rawValue.uppercased()
+
+            addChild(demoViewController)
+            view.addSubview(demoViewController.view)
+        case .swiftui:
+            let swiftUIController = UIHostingController(rootView: SwiftUICardDemo(baseViewModel: BaseViewModel()))
```
minhthenguyen: Should we also add something like this in `DemoType.swift` so we can get corresponding swiftUI content view similar to how we get view controller for `uikit` case above:
```
var swiftUIView: Content {
   switch self {
      case .card:
          ...
      case .paypal:
          ...
   }
}
```

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -87,49 +97,18 @@ class FeatureBaseViewController: UIViewController {
         DemoMerchantAPI.sharedService.createOrder(orderParams: orderRequestParams) { result in
             switch result {
             case .success(let order):
-                self.updateTitle("Order ID: \(order.id)")
-                self.orderID = order.id
+                self.baseViewModel.updateTitle("Order ID: \(order.id)")
+                DispatchQueue.main.async {
```
minhthenguyen: why do we need to use `DispatchQueue.main.async` here?

Demo/Demo/ViewController.swift
```diff
@@ -29,13 +30,30 @@ class ViewController: UIViewController {
...
-        view.addSubview(demoViewController.view)
+        switch DemoSettings.demoUIFramework {
+        case .uikit:
+            let demoViewController = DemoSettings.demoType.viewController
+            demoViewController.modalPresentationStyle = .fullScreen
+
+            navigationItem.title = DemoSettings.demoType.rawValue.uppercased()
+
+            addChild(demoViewController)
+            view.addSubview(demoViewController.view)
```
minhthenguyen, line 43: Not relevant to this PR,
but I just realized everytime we switch demo settings, we'll keep adding childs/subviews to this view controller without ever removing.
Can we remove them before adding new child/subview?
```
children.forEach { viewController in 
  viewController.willMove(toParent: nil)
  viewController.view.removeFromSuperview()
  viewController.removeFromParent()
}
```

we can also make it a function in `extension UIViewController` so we can reuse it :smile:

Demo/Demo/Views/FeatureBaseViewControllerRepresentable.swift
```diff
@@ -0,0 +1,15 @@
+import SwiftUI
+
+@available(iOS 13, *)
+struct FeatureBaseViewControllerRepresentable: UIViewControllerRepresentable {
+
+    var baseViewModel: BaseViewModel
+
+    func makeUIViewController(context: Context) -> UIViewController {
```
minhthenguyen: I don't have much knowledge about SwiftUI so this maybe a silly question :smile: 
But I don't see we use this anywhere, what is this for? or is this something required by SwiftUI?

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
```diff
@@ -0,0 +1,114 @@
...
+                .cornerRadius(10)
+                .disabled(!(cvvText.count >= 3 && cardNumberText.count >= 17 && expirationDateText.count == 7))
+            }
+            .padding(.horizontal, 16)
+        }
+    }
+
+    // MARK: Functions
+
+    private func formatFieldWith(_ text: String, field: Fields) {
```
minhthenguyen: Can we put button enable/disable logic here to make behavior consistent with uikit card demo?

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
```diff
@@ -0,0 +1,114 @@
...
+            cardNumberText = cardFormatter.formatCardNumber(cardNumberText)
+        case .expirationDate:
+            let cleanedText = text.replacingOccurrences(of: " / ", with: "")
+            expirationDateText = String(cleanedText.prefix(4))
+            expirationDateText = expirationDateText.filter { ("0"..."9").contains($0) }
+            expirationDateText = cardFormatter.formatExpirationDate(expirationDateText)
+
+        case .cvv:
+            cvvText = String(text.prefix(4))
+            cvvText = cvvText.filter { ("0"..."9").contains($0) }
```
minhthenguyen: Can we move `formatCardNumber` and `formatExpirationDate` functions from `CardDemoViewController` to maybe `CardFormatter`, and then reuse them here instead of writing new? :smile:

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    private func createCard() -> Card? {
+        let cleanedCardText = cardNumberText.replacingOccurrences(of: " ", with: "")
+
+        let expirationComponents = expirationDateText.components(separatedBy: " / ")
+        let expirationMonth = expirationComponents[0]
+        let expirationYear = "20" + expirationComponents[1]
+
+        return Card(number: cleanedCardText, expirationMonth: expirationMonth, expirationYear: expirationYear, securityCode: cvvText)
+    }
```
minhthenguyen: hmmm I see the integration looks similar whether we use uikit or swiftUI, can we put these somewhere so we can reuse them in both places 🤔

Demo/Demo/ViewController.swift
jaxdesmarais: I think we _could_ but currently our PayPalButton(s) aren't representable in SwiftUI. We should be able to use [`UIViewRepresentable`](https://developer.apple.com/documentation/swiftui/uiviewrepresentable) in a similar manner to how this PR uses [`UIViewControllerRepresentable`](https://developer.apple.com/documentation/swiftui/uiviewcontrollerrepresentable) to create a public wrapper in our PayPal interface to allow our buttons to be represented easily in SwiftUI, but that'd be at the SDK level, so I didn't want to make that change quite yet without chatting as a team.

That to say, I think we should and then can make the above change as well as add a SwiftUI View for the PayPal case,  but for now the PayPal case just takes you to our UIKit view. Since this will require some work and a team chat, thoughts on punting on this for now - with the hopes that this can be a follow on story to make this change at the SDK level and add out buttons natively via SwiftUI to the demo?

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
jaxdesmarais: Without this since the orderID is now in a view model to be shared among our 2 views, we get the error `Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.`

Demo/Demo/Views/FeatureBaseViewControllerRepresentable.swift
jaxdesmarais: Great question - so `FeatureBaseViewControllerRepresentable` is used by `SwiftUICardDemo` when we add the `FeatureBaseViewController` to our SwiftUI view here:
```
FeatureBaseViewControllerRepresentable(baseViewModel: baseViewModel)
```
Under the hood, SwiftUI calls the `makeUIViewController` function when creating this view as part of the `UIViewControllerRepresentable` protocol. Essentially this is our abstraction layer for our UIViewController to allow us to represent this view within SwiftUI. If you add a breakpoint to this function then spin up the demo and select `SwiftUI` you will see the system calls this function automatically when needed. 😄

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
jaxdesmarais: So we _can_ but this isn't best practice in SwiftUI. SwiftUI uses the `disabled` attribute on the button, which I have implemented here:
```
.disabled(!(cvvText.count >= 3 && cardNumberText.count >= 17 && expirationDateText.count == 7))
```
We can do either but I went with this because it seems to be the preferred method with Swift UI. The button is disabled in the same way as the other card demo, but just in each UI Frameworks preferred way.

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
jaxdesmarais: That's a good call - I'll move this logic in there to share across the 2 interfaces! 😄

Demo/Demo/ViewController.swift
minhthenguyen: ah got it!
what do you think of just using an empty view or something similar for PayPal SwiftUI case for now, so we won't show card demo when PayPal + SwiftUI is selected?

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
minhthenguyen: ahhh nice, that makes sense.
Maybe we can add the check for `orderID != nil` here?

and I wonder if we can grey out the text when it's disabled, like `.foregroundColor(<bool to see if button should be enabled> ? .white : .gray)` or something similar?

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
jaxdesmarais: Moved them to a shared location so we can use this for either without duplicating. 😄

Demo/Demo/ViewController.swift
jaxdesmarais: Sure - that makes sense for now, will add.

Demo/Demo/FeatureViewControllers/SwiftUICardDemoView.swift
jaxdesmarais: Updated and made some updates to the enable pay button function so it can be shared across the two frameworks

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -98,7 +96,7 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
     /// Allows us to only enable the pay button once all fields are filled out and we have an order ID
     @objc override func createOrderTapped() {
         createOrder { _ in
-            self.enablePayButton(
+            self.payButton.isEnabled = self.baseViewModel.enablePayButton(
```
sarahkoop: nitpick: can we call this something other than payButton? It reads too closely to payPalButton to me

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -87,49 +97,18 @@ class FeatureBaseViewController: UIViewController {
...
+                    self.baseViewModel.orderID = order.id
```
sarahkoop: Should updating the order ID always be on the main thread or just in this case? If so would it make more sense to have an updateOrderID method that calls `DispatchQueue.main.async` under the hood (similar to updateTitle)?

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -82,13 +82,13 @@ class PayPalDemoViewController: FeatureBaseViewController {
...
-                self.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
+                self.baseViewModel.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
                 print("✅ Order is successfully approved and ready to be captured/authorized with result: \(result)")
             case .failure(let error):
-                self.updateTitle("\(DemoSettings.intent) failed: \(error.localizedDescription)")
+                self.baseViewModel.updateTitle("\(DemoSettings.intent) failed: \(error.localizedDescription)")
                 print("❌ There was an error: \(error)")
             case .cancellation:
-                self.updateTitle("\(DemoSettings.intent) cancelled")
+                self.baseViewModel.updateTitle("\(DemoSettings.intent) cancelled")
```
sarahkoop:
```suggestion
                self.baseViewModel.updateTitle("\(DemoSettings.intent) canceled")
```
Nitpick: for consistency let's keep single l canceled throughout

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+                self.processOrder(orderID: result.orderID) {
+                    completion()
+                }
+            case .failure(let error):
+                self.updateTitle("\(DemoSettings.intent) failed: \(error.localizedDescription)")
+            }
+        }
+    }
+
+    func enablePayButton(cardNumber: String, expirationDate: String, cvv: String) -> Bool {
```
sarahkoop: I think this might make more sense as `isCardFormValid` or `validateCardFields` or something like that since it's not actually enabling the button.

sarahkoop (approved):
Left a few nitpick comments but overall looks great to me! 🌟

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: Totally - does `checkoutButton` make more sense?

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
jaxdesmarais: I think it makes sense to always be on the main thread - SwiftUI needs the update to the `@Published var` to happen on the main thread which was why this was added. Updating this in a similar manner to the `updateTitle` function.

Demo/Demo/ViewModels/BaseViewModel.swift
jaxdesmarais: I think `isCardFormValid` makes this way more clear - thanks for this suggestion!

jaxdesmarais: I think `isCardFormValid` makes this way more clear - thanks for this suggestion!

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
jaxdesmarais: It would be fine for this to always happen on the main thread - this was originally updated because of how SwiftUI handles the `@Published var` of the order ID. I think it makes since to format this in a similar way to the `updateTitle` function.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: Totally - updated to `checkoutButton` but if there is an _even better_ name happy to update again

sshropshire (approved):
🌊

Demo/Demo/FeatureViewControllers/FeatureBaseViewController.swift
```diff
@@ -76,7 +86,7 @@ class FeatureBaseViewController: UIViewController {
     }
 
     func createOrder(completion: @escaping (String?) -> Void = { _ in }) {
```
sshropshire: random nit: could this go in the BaseViewModel too?

jaxdesmarais: Done!

PR #28, opened 2021-12-07
Card Request Parameter Object

sshropshire:
### Reason for changes

This PR is to align with a corresponding [Android PR](https://github.com/paypal/Android-SDK/pull/30) to align the Card module interface.

### Summary of changes

 - Add a `CardRequest` [parameter object](https://wiki.c2.com/?ParameterObject)
 - ~Rename `CardClient.approveOrder` to `CardClient.approve`~
 - ~General refactoring to better align with [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/)~

### Checklist

### DIFF
--- a/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
+++ b/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
@@ -125,7 +125,8 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let config = CoreConfig(clientID: DemoSettings.clientID, environment: DemoSettings.environment.paypalSDKEnvironment)
         let cardClient = CardClient(config: config)
 
-        cardClient.approveOrder(orderID: orderID, card: card) { result in
+        let cardRequest = CardRequest(orderID: orderID, card: card)
+        cardClient.approveOrder(request: cardRequest) { result in
             switch result {
             case .success(let result):
                 self.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
--- a/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
+++ b/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
@@ -128,9 +128,9 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let cardOrder = CardOrder(orderID: orderID, card: card)
         cardClient.approve(cardOrder) { result in
             switch result {
-            case .success(let result):
+            case .success(let approval):
                 self.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
-                self.processOrder(orderID: result.orderID) {
+                self.processOrder(orderID: approval.orderID) {
                     self.payButton.stopAnimating()
                 }
             case .failure(let error):
--- a/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
+++ b/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
@@ -125,8 +125,8 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let config = CoreConfig(clientID: DemoSettings.clientID, environment: DemoSettings.environment.paypalSDKEnvironment)
         let cardClient = CardClient(config: config)
 
-        let cardOrder = CardOrder(orderID: orderID, card: card)
-        cardClient.approve(cardOrder) { result in
+        let cardRequest = CardRequest(orderID: orderID, card: card)
+        cardClient.approveOrder(request: cardRequest) { result in
             switch result {
             case .success(let result):
                 self.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
--- a/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
+++ b/Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
@@ -128,9 +128,9 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let cardOrder = CardOrder(orderID: orderID, card: card)
         cardClient.approve(cardOrder) { result in
             switch result {
-            case .success(let approval):
+            case .success(let result):
                 self.updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
-                self.processOrder(orderID: approval.orderID) {
+                self.processOrder(orderID: result.orderID) {
                     self.payButton.stopAnimating()
                 }
             case .failure(let error):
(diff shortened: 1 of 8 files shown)


Sources/Card/Models/CardOrder.swift
```diff
@@ -0,0 +1,19 @@
...
+
+public struct CardOrder {
+
+    /// The ID of the order to be approved
+    public let orderID: String
+
+    /// The card to be charged for this order
+    public let card: Card
+
+    /// Creates an instance of a card request
```
scannillo:
```suggestion
    /// Creates an instance of a CardOrder
```

Should this still reference `card request`? Also, I'm curious why on Android we went with `CardRequest` and on iOS went with `CardOrder`?

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -125,11 +125,12 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let config = CoreConfig(clientID: DemoSettings.clientID, environment: DemoSettings.environment.paypalSDKEnvironment)
         let cardClient = CardClient(config: config)
 
-        cardClient.approveOrder(orderID: orderID, card: card) { result in
+        let cardOrder = CardOrder(orderID: orderID, card: card)
+        cardClient.approve(cardOrder) { result in
```
jaxdesmarais: Looking at the changes from the merchant perspective, this naming feels pretty confusing. I wonder if `CardRequest` would be a better name than `CardOrder`? To me, order implies that they would use this to create the order, but in reality they are just passing us an order ID (having created an order outside of this request) and a card object. I'm not tied to `CardRequest` either, just trying to think through if there may be a more descriptive name for this. Thoughts?

Sources/Card/Models/CardOrderApproval.swift
```diff
@@ -4,7 +4,7 @@ import PaymentsCore
 #endif
 
 /// The result of a card payment flow.
-public struct CardResult {
+public struct CardOrderApproval {
```
jaxdesmarais: I think this name also feels confusing for a first time merchant integrating. Returning an object that ends with the word `Approval` makes it seem like they are done with the order when in reality they need to take this result and either authorize or capture the order after they get this result back.

Sources/Card/Models/CardOrder.swift
sshropshire: So on Android it's still `CardRequest`. I think the API for iOS will take the lead with naming in the long run, since they have really well thought out [design guidelines](https://www.swift.org/documentation/api-design-guidelines/) for Swift. We haven't explicitly said we're going to follow these for this SDK, but it would produce an API that feels most like existing Apple APIs.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
sshropshire: Yeah I left these in a single commit so they'd be easy to revert, but I wanted to try to align naming more with [this doc](https://www.swift.org/documentation/api-design-guidelines/#omit-needless-words) to see what it looks like. The example they give in the More Detail section shows the method name as a verb and the first argument is like a noun/subject, so it reads to the merchant as "card client approve card order." To that point, "card client approve card request" would also work.

It may be out of scope for this PR but definitely wanted to try it out to see if we could get more expressiveness out of the API.

Sources/Card/Models/CardOrderApproval.swift
sshropshire: Totally agree. There's a lot of good symmetry with `CardRequest` and `CardResult` as well.

Sources/Card/Models/CardOrder.swift
sshropshire: Also TLDR the plan was to finalize on iOS and circle back to Android.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
sshropshire: I think it makes sense to revert this for now and keep it in mind down the line. For now, the `CardRequest` object should help us extend the API without causing a breaking change to the method signature.

Sources/Card/Models/CardOrder.swift
sshropshire: I went ahead and reverted. I think it'll help to re-visit once we have more merchant-facing features / sample integrations.

Sources/Card/Models/CardOrderApproval.swift
sshropshire: I went ahead and reverted to clean this up and maybe revisit down the line if it makes sense.

minhthenguyen (approved):
this looks good to me. thanks for making this change and align between Android and iOS :smile:

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -125,7 +125,8 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         let config = CoreConfig(clientID: DemoSettings.clientID, environment: DemoSettings.environment.paypalSDKEnvironment)
         let cardClient = CardClient(config: config)
 
-        cardClient.approveOrder(orderID: orderID, card: card) { result in
+        let cardRequest = CardRequest(orderID: orderID, card: card)
+        cardClient.approveOrder(request: cardRequest) { result in
```
jaxdesmarais, line 129: Just a thought based on the swift docs you sent over yesterday, but we _could_ name this function externally like is used here so that it reads as:
```
cardClient.approveOrder(with: cardRequest) { result in
```
Not a blocker by any means, but just a thought to potentially improve merchant clarity of this function. 😄

Sources/Card/CardClient.swift
```diff
@@ -27,11 +27,11 @@ public class CardClient {
     ///   - orderID: The ID of the order to be approved
     ///   - card: The card to be charged for this order
     ///   - completion: Completion handler for approveOrder, which contains data of the order if success, or an error if failure
-    public func approveOrder(orderID: String, card: Card, completion: @escaping (Result<CardResult, PayPalSDKError>) -> Void) {
+    public func approveOrder(request: CardRequest, completion: @escaping (Result<CardResult, PayPalSDKError>) -> Void) {
```
jaxdesmarais, line 30: If we wanted to do that rename to use `with` externally we could just do this here:
```
public func approveOrder(with request: CardRequest, completion: @escaping (Result<CardResult, PayPalSDKError>) -> Void) {
```

jaxdesmarais (approved):
This looks great! Thanks for making these alignment changes. Left a comment about a potential function name change, but not a blocker for this PR if we don't feel like it'd add the clarity we are looking for.

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
sshropshire: Def! Yeah that does read nicely.

Sources/Card/CardClient.swift
sshropshire: I'm up for it! I guess one thing I didn't consider was the alignment with Android. It may make sense to circle back once more of the API is done.

The SwiftUI guidelines are super in-depth. I think it makes sense for iOS to take lead on overall API structure and we can take the same blueprint and transform it for Android since named parameters are handled a bit differently in Kotlin (and are nonexistent in Java).

PR #29, opened 2021-12-20
Card Client Docs

sshropshire:
### Summary of changes

 - Add Docs for iOS Card Client Integration

 ### Checklist

### Authors

- @minhthenguyen
- @sshropshire


docs/index.md
```diff
@@ -0,0 +1,109 @@
+# Pay with Card
+
+Accept card payments in your app using the PayPal In-App Payments SDK.
+
+## How it works
+
+The In-App payments SDK allows for merchant apps to integrate easily with PayPal payment services on both the client and server.
+
+You need to have a server integration to create an order and capture the funds using [PayPal Orders v2 API](https://developer.paypal.com/docs/api/orders/v2). The Payments SDK will allow you to approve the order on the client side of your app by using the buyer's card information.
```
sarahkoop:
```suggestion
You will need a server integration to create an order and capture the funds using [PayPal Orders v2 API](https://developer.paypal.com/docs/api/orders/v2). The Payments SDK allows for client-side approval of an order with the user's card information.
```

sarahkoop: Just some potential wording suggestions

docs/index.md
```diff
@@ -0,0 +1,109 @@
+# Pay with Card
+
+Accept card payments in your app using the PayPal In-App Payments SDK.
+
+## How it works
+
+The In-App payments SDK allows for merchant apps to integrate easily with PayPal payment services on both the client and server.
```
sarahkoop: We aren't super consistent with the naming of the SDK - is In-App Payments SDK the name or just Payments SDK? Either way we should make the capitalization and naming the same across this doc.

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Steps
+
+#### 1. Create an order
+Call `v2/checkout/orders` to create an order an obtain an order ID
+
+```
+curl --location --request POST 'https://api.sandbox.paypal.com/v2/checkout/orders/' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer <access_token>' \
```
sarahkoop: Should we include docs here for how to obtain the access token or will those live somewhere else?

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+    ]
+}'
+```
+
+#### 2. Approve the order using Payments SDK:
+
+```swift
+import PaymentsSDK
+
+// Create an config object using your clientID obtained from PayPal application portal
```
sarahkoop:
```suggestion
// Create a config object using your clientID obtained from PayPal application portal
```

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+let config = CoreConfig(clientID: "<CLIENT_ID>", evironment: .sandbox)
+
+// Create a CardClient object to process a card payment
```
sarahkoop:
```suggestion
// Create a CardClient to approve an order with a card payment method
```

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+let cardClient = CardClient(config: config)
+
+// Create a card object from buyer's inputs
```
sarahkoop: Nitpick but I prefer "user" over "buyer" throughout the docs - does anyone else have on opinion on this?

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+    expirationYear: "25",
+    securityCode: "123"
+)
+
+// Create a CardRequest and approve the order using buyer's card when buyer submit their card info.
+let cardRequest = CardRequest(orderID: "<ORDER_ID>", card: card)
+cardClient.approveOrder(request: cardRequest) { result in
+    switch result {
+        case .success(let result):
+            // Order is successfully approved with card and ready to be capture/authorize.
```
sarahkoop:
```suggestion
            // Order was successfully approved with the card and is ready to be captured/authorized.
```

docs/index.md
sshropshire: These felt like placeholders I'm not sure where that would live. Honestly, a lot of this will be duplicated across web, iOS and Android repos.

docs/index.md
sshropshire: That's a good call. I went with Payments we have been saying that more often than In-App.

docs/index.md
sshropshire: I agree. I ended up omitting both user / buyer since we're statically supplying the input with hardcoded params.

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+
+- [Card Fields UI Integration](#card-fields-ui-integration)
+- [Fully Customizable Integration](#fully-customizable-integration)
+
+## Card Fields UI Integration
+
+Accept card payments using our card fields UI
+
+```
+# Placeholder: Include images of card fields
```
rsdighe76: is this meant to be when we have something like drop - in ?

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+// Create a card object
+let card = Card(
+    number: "4111111111111111",
+    expirationMonth: "01",
+    expirationYear: "25",
+    securityCode: "123"
+)
+
+// Create a CardRequest and approve the order using buyer's card when buyer submit their card info.
+let cardRequest = CardRequest(orderID: "<ORDER_ID>", card: card)
```
rsdighe76: I would create an order at this step. 
The default simple integration that we have talked about with @demerino and others was that the order creation is upon a "button" click. "Creating Order" upfront like it is here to me is an optimization not the default way to integrate particularly as not everyone has all payment method with us so they don't really want to create the order until they think it's worth while and that is only evident once the buyer clicks on a specific payment button and clicks submit.

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+        case .success(let result):
+            // Order was successfully approved with the card and is ready to be captured/authorized.
+        case .failure(let error):
+            // Encountered error when approving order.
+    }
+}
+```
+
+#### 3. Capture/authorize the order
+Call `v2/checkout/orders/<ORDER_ID>/authorize` to place funds on hold:
```
rsdighe76: this step is technically not possible if you have the intent as "capture" as in the step above ... 
so either we have 2 guides one when intent is 'authorize" and 'capture' or we find some way to articulate it.

docs/index.md
```diff
@@ -0,0 +1,109 @@
...
+import PaymentsSDK
+
+// Create a config object using your clientID obtained from PayPal application portal
+let config = CoreConfig(clientID: "<CLIENT_ID>", evironment: .sandbox)
+
+// Create a CardClient to approve an order with a card payment method
+let cardClient = CardClient(config: config)
+
+// Create a card object
+let card = Card(
```
rsdighe76: this probably is not literally how someone would code it as they will get the inputs from some UI - would it make sense to separate out the steps such that it is

1. Show your Card Fields  
2. Render "Pay now" button
3. On submit >  Create + Approve Order
4. On success > Authorize or Capture depending on intent during Order creation

Is there any example of anyone else articulating these steps

docs/index.md
rsdighe76: Also note that for "card processing" specifically we can just authorizeOrder or captureOrder without an implicit approveOrder but my recollection is that we did the below to maintain consistency with other payment methods. 

The way the api works is you can do 

1. Single Step (Create Order + Confirm Order + Authorize/Capture) - in a single call by passing card information as part of the order call 

```
"intent": "CAPTURE",
    "purchase_units": [
        {
            "amount": {
                "currency_code": "USD",
                "value": "5.00"
            }
        }
    ]
    "payment_source" : {
    	"card" : {
    		"number":"41111111111111"
    		"expiry":"2023-12"
    		"securityCode":"212"
    	}

    }
```

2. Three Steps
> Create Order
> Confirm Order
> Authorize/Capture Order

rsdighe76: Where would someone pass name and address ?

docs/index.md
tdchow: I think we could provide links to the existing documentation for getting an access token: https://developer.paypal.com/api/rest/#link-getanaccesstoken

A link to the create order documentation may help as well: https://developer.paypal.com/api/orders/v2/#orders_createe

docs/index.md
sshropshire: Correct yes. Something similar to the [DropIn docs](https://developer.paypal.com/braintree/docs/guides/drop-in/overview) we have on the BT side. It's a good way to give potential merchants a glimpse of what they're integrating before having to write a line of code.

docs/index.md
sshropshire: In the future we could link to the server SDKs. I'm hesitant to add links at the moment since these are placeholders. `curl` provides a good platform-agnostic way to illustrate steps to fetch a client token.

docs/index.md
sshropshire: Good call. I'll add those in as well.

docs/index.md
sshropshire: We are making several assumptions here since each parameter is of type `String`, we assume that the developer knows where to access the string content necessary to construct a `Card` object.

sshropshire: I agree some visual elements may be helpful, we could provide an image of a mock UI alongside the code samples to streamline the code snippets.

docs/index.md
sshropshire: I believe we still need to spec out the default simple integration but I see that we need separate auth and capture flows. Step 3 in that case would be considered _optional_. I can tweak the docs a bit to reflect this.

docs/index.md
sshropshire: @tdchow I'll reference the links though for good measure that's a good call.

docs/index.md
```diff
@@ -0,0 +1,125 @@
+# Pay with Card
+
+Accept card payments in your app using the PayPal In-App Payments SDK.
```
jaxdesmarais: Are there any card types we _don't_ support today? If so it may be worth calling that out here. Looking at [PPaaS](https://ppaas/api/3719176155270329?tab=apiReference&operation_id=post-_v2_checkout_orders_id_confirm_payment_source) for this endpoint it seems like we should cover most everything so maybe calling out the specific types isn't needed. 🤷‍♂️

docs/index.md
```diff
@@ -0,0 +1,125 @@
...
+## Card Fields UI Integration
+
+Accept card payments using our card fields UI
+
+```
+# Placeholder: Include images of card fields
+```
+
+**Requirements:**
+Card fields UI offers a low effort integration where you can use our card fields UI, and our card fields will handle ...
```
jaxdesmarais: I feel like we should move away from low effort vs high effort language as this could vary by dev experience, how their app set up, and potentially cause frustration with our SDKs. Maybe low vs high customization would be better? or something else? Thoughts?

docs/index.md
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+    ]
+}'
+```
+
+##### Additional Links
+
+- [PayPal Developer: Get Started – Get an access token](https://developer.paypal.com/api/rest/#link-getanaccesstoken)
+- [PayPal Developer: Orders – Create an order](https://developer.paypal.com/api/orders/v2/#orders_createe)
```
jaxdesmarais: This URL works but looks like `#orders_create` also points to the same thing 😄

docs/index.md
jaxdesmarais: I don't think I am fully following this. What @rsdighe76 is describing sounds like doing hook into allowing order creation client side which at one point we decided we didn't want to do currently. Has this changed? We _do_ have create order as a client side option on NXO and it added a _ton_ of models since there are a number of options. If we do decide to go this route we should probably decide what is most important to merchants from day one and potentially add other models as merchants request them to keep the SDK light.

docs/index.md
sshropshire: Yeah i'm not sure. Seeing the card logos does make it feel official though that's a good callout.

docs/index.md
sshropshire: I've seen no-code and low-code, but for traditional development I'm not sure of the marketing term lol. We could go with "drop-in" here since that's technically its purpose, but I'm up for making the language read more like an executive summary. The person evaluating our SDK may not always be the integration engineer.

docs/index.md
sshropshire: Yeah that's how the link icon copied it lol I saw that too. I can remove the extra e.

docs/index.md
sshropshire: @jaxdesmarais yeah I believe we agreed in principle to server-side order creation to narrow the scope, but we don't have a formal spec for the default simple integration AFAIK.

jaxdesmarais: Cool cool, that makes sense! Just wanted to make sure I wasn't misremembering about not doing create order client side. 😄

docs/index.md
jaxdesmarais: Hahaha, yeah maybe this is something marketing can help us with.

docs/index.md
wallportillo: @sarahkoop and I talked about it and we feel that we should move this forward because we are not offering client-side order creation

docs/Card/integration.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Add Card Payments
+
+Accept card payments with PayPal Payments SDK using your own UI.
+
+### 1. Add the Payments SDK  to your app
+
+#### Swift Package Manager
+
+To add the PayPal package to your Xcode project, select _File > Swift Packages > Add Package Dependency_ and enter https://github.com/paypal/iOS-SDK as the repository URL. Tick the checkboxes for the specific PayPal libraries you wish to include.
```
cgdibble: In attempt to avoid documenting other tools, maybe consider pointing them to the Xcode docs on adding a dependencies? https://developer.apple.com/documentation/swift_packages/adding_package_dependencies_to_your_app 

Doing that you could then get something like:
```
In Xcode, add the [package dependency](https://developer.apple.com/documentation/swift_packages/adding_package_dependencies_to_your_app) and enter https://github.com/paypal/iOS-SDK as the repository URL. Tick the checkboxes for the specific PayPal libraries you wish to include.
```

cgdibble (approved):

docs/Card/integration.md
sarahkoop: Agreed - then if Xcode has an update and the location of this changes, we won't have to change the docs

jaxdesmarais (approved):
🍏

PR #31, opened 2022-01-05
Paypal Integration Docs

jaxdesmarais:
### Summary of changes

- Add documentation for integrating PayPal via the SDK.
- [Android PR for this change](https://github.com/paypal/Android-SDK/pull/33)

### Checklist

- @sarahkoop @jaxdesmarais

### DIFF
--- a/docs/PayPal/get-started.md
+++ b/docs/PayPal/get-started.md
@@ -1,0 +1,30 @@
+---
+title: Pay with PayPal
+keywords: 
+contentType: docs
+productStatus: current
+apiVersion: TODO
+sdkVersion: TODO
+---
+
+# Pay with PayPal
+
+Accept PayPal payments in your iOS app using the PayPal Payments SDK.
+
+## How it works
+
+This diagram shows how your client, your server, and PayPal interact:
+
+// TODO - Get a diagram of the payment flow similar to [this](https://developer.paypal.com/braintree/docs/start/overview#how-it-works)
+
+## Eligibility
+
+PayPal is available as a [payment method to merchants](https://developer.paypal.com/docs/checkout/payment-methods/) in multiple [countries and currencies](https://www.paypal.com/us/webapps/mpp/country-worldwide).
+
+The PayPal iOS SDK supports a minimum deployment target of iOS 13+ and requires Xcode 13+ and macOS Big Sur 11.3+.
+iOS apps can be written in Swift 5.1+.
+This SDK can be integrated with CocoaPods or Swift Package Manager.
+
+## How to integrate
+
+- [Custom Integration](integration.md)
--- a/docs/PayPal/integration.md
+++ b/docs/PayPal/integration.md
@@ -29,7 +29,7 @@ For initial setup, the `curl` commands below can be used in place of a server SD
 
 #### Swift Package Manager
 
-To add the PayPal package to your Xcode project, select _File > Swift Packages > Add Package Dependency_ and enter https://github.com/paypal/iOS-SDK as the repository URL. Tick the checkboxes for the specific PayPal libraries you wish to include.
+In Xcode, add the [package dependency](https://developer.apple.com/documentation/swift_packages/adding_package_dependencies_to_your_app) and enter https://github.com/paypal/iOS-SDK as the repository URL. Tick the checkboxes for the specific PayPal libraries you wish to include.
 
 In your app's source code files, use the following import syntax to include PayPal's libraries:
 
--- a/docs/PayPal/integration.md
+++ b/docs/PayPal/integration.md
@@ -72,10 +72,16 @@ class MyViewController: ASWebAuthenticationPresentationContextProviding {
 
 ### 3. Create a PayPal button 
 
+#### UIKit
 Add a `PayPalButton` to your View:
 
 ```swift
 let payPalButton = PayPalButton()
+```
+
+#### SwiftUI
+```swift
+PayPalButton.Representable { }
 ```
 
 ### 4. Initiate the Payments SDK
--- a/docs/PayPal/integration.md
+++ b/docs/PayPal/integration.md
@@ -1,37 +1,31 @@
-# Pay with PayPal
+---
+title: Pay with PayPal Custom Integration
+keywords: 
+contentType: docs
+productStatus: current
+apiVersion: TODO
+sdkVersion: TODO
+---
 
-Accept PayPal payments in your iOS app using the PayPal Payments SDK.
+# Pay with PayPal Custom Integration
 
-## How it works
+Follow these steps to add Card payments.
 
-This diagram shows how your client, your server, and PayPal interact:
+1. [Know before you code](#know-before-you-code)
+1. [Add PayPal Payments](#add-paypal-payments)
+1. [Test and go live](#test-and-go-live)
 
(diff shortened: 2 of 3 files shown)


docs/paypal.md
```diff
@@ -0,0 +1,173 @@
+# Pay with PayPal
+
+Accept PayPal payments in your iOS app using the PayPal Payments SDK.
```
jaxdesmarais: We haven't landed on a name for this SDK yet so have used a few different names throughout docs/project. We should settle on a differentiated name for the SDK as well as the PayPal module. Currently we call the SDK `PayPal` and also have a `PayPal` module for the PayPal checkout flow.

docs/paypal.md
```diff
@@ -0,0 +1,173 @@
...
+Create a `CoreConfig` using your client ID from the PayPal Developer Portal:
+
+```swift
+let config = CoreConfig(clientID: "<CLIENT_ID>", environment: .sandbox)
+```
+
+Configure a return URL using your bundle identifier:
+
+```swift
+let returnUrl = "<BUNDLE_IDENTIFIER>" + "://paypalpay"
```
jaxdesmarais: The return URL is still a pain point as it's not used in the way that a return URL would typically be used in iOS. Additionally, it has to be all lowercase which causes merchant issues when integrating as their bundle ID often includes mixed cases.

docs/paypal.md
```diff
@@ -0,0 +1,173 @@
...
+@objc func paymentButtonTapped() {
+    checkoutWithPayPal(orderID: "<ORDER_ID>")
+}
+```
+
+Call `PayPalClient.start` to approve the order, and then handle results:
+
+```swift
+func checkoutWithPayPal(orderID: String) {
+    payPalClient.start(orderID: orderID, presentingViewController: self) { state in
```
jaxdesmarais: Need to align on naming of this function. On iOS it's called `start` and on Android it's `checkout`. We need to update this method to take a `PayPalRequest` and confirm a new name.

docs/paypal.md
```diff
@@ -0,0 +1,173 @@
...
+        switch state {
+        case .success(let result):
+            // capture/authorize the order using `result.orderID` (see step 6)        
+        case .failure(let error):
+            // handle the error by accessing `error.localizedDescription`
+        case .cancellation:
+            // the user canceled
```
jaxdesmarais: Need to confirm if we can move forward with adding cancellation to the failure case. This had been discussed previously.

docs/paypal.md
```diff
@@ -0,0 +1,173 @@
...
+
+## How it works
+
+This diagram shows how your client, your server, and PayPal interact:
+
+// TODO - Get a diagram of the payment flow similar to [this](https://developer.paypal.com/braintree/docs/start/overview#how-it-works)
+
+## Eligibility
+
+PayPal is available as a payment method to merchants in multiple [countries and currencies](https://developer.paypal.com/docs/checkout/payment-methods/).
```
jonathajones: This might be a better link for `countries and currencies`:  https://www.paypal.com/us/webapps/mpp/country-worldwide

The current link is kind of confusing. It doesn't seem to show the countries that PayPal is available in, but rather feature support in different countries

docs/PayPal/get-started.md
```diff
@@ -0,0 +1,30 @@
...
+sdkVersion: TODO
+---
+
+# Pay with PayPal
+
+Accept PayPal payments in your iOS app using the PayPal Payments SDK.
+
+## How it works
+
+This diagram shows how your client, your server, and PayPal interact:
```
sshropshire, line 16:
```suggestion
This diagram shows how your client, server, and PayPal interact:
```

docs/PayPal/integration.md
```diff
@@ -0,0 +1,188 @@
...
+        }
+    ]
+}'
+```
+
+The `id` field of the response contains the order ID to pass to your client.
+
+### 6. Create a request object for launching the PayPal flow
+
+Configure your `PayPalRequest` and include the order ID generated (step 5):
```
sshropshire: Maybe hyperlink to step 5 here?

sshropshire (approved):
🌊 &nbsp;LGTM! Left some take or leave nits.

docs/PayPal/integration.md
```diff
@@ -0,0 +1,188 @@
...
+1. [Add PayPal Payments](#add-paypal-payments)
+1. [Test and go live](#test-and-go-live)
+
+## Know before you code
+
+You will need to set up authorization to use the PayPal Payments SDK. 
+Follow the steps in [Get Started](https://developer.paypal.com/api/rest/#link-getstarted) to create a client ID. 
+
+You will need a server integration to create an order and capture the funds using [PayPal Orders v2 API](https://developer.paypal.com/docs/api/orders/v2). 
+For initial setup, the `curl` commands below can be used in place of a server SDK.
```
cgdibble, line 24: Just to clarify, all of the curl commands used below will allow you to get setup without doing the two prerequisites above?

cgdibble (approved):

docs/PayPal/integration.md
sarahkoop: No - the curl commands allow you not to create a server integration, but you still need to do the step on line 21. Could I phrase that somehow that makes it more clear?

cgdibble: 🤔 How about something like this: 
```
Required:
You will need to set up authorization to use the PayPal Payments SDK. 
Follow the steps in [Get Started](https://developer.paypal.com/api/rest/#link-getstarted) to create a client ID. 

Optional: 
- When creating your own server integration use [PayPal Orders v2 API](https://developer.paypal.com/docs/api/orders/v2). However, if you would like to forego this, you can use the provided `curl` commands given with each relevant step.
```

Not sure on the optional/required aspect, but wondering how to best separate them.

sarahkoop: Hmm.. Not sure about that because it isn't really optional. You have to use something to create the order (for now it's curl commands but hopefully when server SDKs exist that will be an option)

cgdibble: Ah fair enough. Maybe...
```
In lieu of a server integration that uses [PayPal Orders v2 API](https://developer.paypal.com/docs/api/orders/v2), you can use the provided curl commands to validate your integration.
```
Something like that? Then nix the optional/required bits?

docs/PayPal/integration.md
```diff
@@ -0,0 +1,194 @@
...
+sdkVersion: TODO
+---
+
+# Pay with PayPal Custom Integration
+
+Follow these steps to add Card payments.
+
+1. [Know before you code](#know-before-you-code)
+1. [Add PayPal Payments](#add-paypal-payments)
+1. [Test and go live](#test-and-go-live)
```
jaxdesmarais, line 16: Should these all be 1 or 1, 2, 3?

docs/PayPal/integration.md
```diff
@@ -0,0 +1,194 @@
...
+keywords: 
+contentType: docs
+productStatus: current
+apiVersion: TODO
+sdkVersion: TODO
+---
+
+# Pay with PayPal Custom Integration
+
+Follow these steps to add Card payments.
```
jaxdesmarais:
```suggestion
Follow these steps to add PayPal payments.
```

docs/PayPal/integration.md
sarahkoop: Markdown will auto-number these if all 1 👍

PR #32, opened 2022-01-05
Add PayPalRequest Object

sarahkoop:
### Reason for changes

We have a `CardRequest` object for configuring parameters for card payments, we should follow this pattern for PayPal and all other features.

### Summary of changes

- Add `PayPalRequest` and update `PayPalClient.start` to accept a `PayPalRequest`

### Checklist

- @jaxdesmarais 
- @sarahkoop

### DIFF
--- a/Sources/PayPal/Models/PayPalRequest.swift
+++ b/Sources/PayPal/Models/PayPalRequest.swift
@@ -1,0 +1,14 @@
+import Foundation
+
+/// Used to configure options for approving a PayPal order
+public struct PayPalRequest {
+    
+    /// The order ID associated with the request.
+    public let orderID: String
+    
+    /// Creates an instance of a PayPalRequest.
+    /// - Parameter orderID: The ID of the order to be approved.
+    public init(orderID: String) {
+        self.orderID = orderID
+    }
+}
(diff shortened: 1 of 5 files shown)


sshropshire (approved):
🌊

Asavarkhul:
Hey there!
Nice Idea to follow the `CardRequest ` pattern 👌

I got a question, why using a struct and not a protocol to define what should be a `PayPalRequest `?

Thanks 🙏

Sources/PayPal/Models/PayPalRequest.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/// Used to configure options for approving a PayPal order
+public struct PayPalRequest {
```
Asavarkhul, line 4: Hey there!
Nice Idea to follow the `CardRequest` pattern 👌

I got a question, why using a struct and not a protocol to define what should be a `PayPalRequest` ?

Thanks 🙇‍♂️

jaxdesmarais: While we could certainly use a protocol for the request and have the struct conform to it, the lack of reuse (at least currently) for the protocol as well as the current simplicity of the objects have lead us to use structs for the request objects. This also is a similar pattern to what Apple uses when constructing a `PKPaymentRequest` for Apple Pay. That said, if we see a strong need to move in a protocol direction in the future we would certainly explore this option.

Asavarkhul: Thanks a lot @jaxdesmarais for your answer 🙏

jaxdesmarais (approved):
⛵  looks good!

notoriousmarcos (approved):

PR #34, opened 2022-01-13
PayPal and PayPal Credit Buttons SwiftUI Representable

jaxdesmarais:
### Reason for changes

To represent a button in SwiftUI we need the buttons to be represented as a `UIViewRepresentable`. Taking a look at some best practices across other SDKs, extending the button and adding a Representable struct is the best way to do this. We also needed a coordinator class to coordinate the button actions. While there is some duplication involved, adding this to our SDK allows for us to account for merchants using either/both UIKit and SwiftUI without needing to build out this logic themselves.

Adding this to our SDK allows a merchant to do the following to create a button in their SwiftUI view:
```
PayPalButton {
    // button action here
}
```

This very closely models the way a merchant would do this in standard SwiftUI without a representable:
```
Button("Title") {
    // button action here
}
```

### Summary of changes

- Add `Coordinator` class to pass action to UIKit button in SwiftUI
- Add `PayPalButton` and `PayPalCreditButton` as `UIViewRepresentable` conforming classes
- Add both buttons to the `SwiftUIPayPalDemo`
- Moved shared PayPal logic to `BaseViewModel` - this could probably use a refactor in the future

### Checklist

- @jaxdesmarais @sarahkoop


fastlane/.xcovignore
```diff
@@ -1,2 +1,4 @@
 # Exclude PayPal Interfaces
 - Sources/PayPal/PayPalUIInterfaces
+# Exclude SwiftUI Interface until a decision is made on UITesting Frameworks
+- Sources/PayPal/Buttons
```
jaxdesmarais, line 4: There is not an easy way to test SwiftUI views via unit tests. Adding this to pass test coverage until we decide to do either standard UI testing or snapshot testing. Once a UI testing plan is created, we should remove this.

notoriousmarcos (approved):

Demo/Demo/SwiftUIComponents/SwiftUIPayPalDemo.swift
```diff
@@ -1,4 +1,5 @@
 import SwiftUI
+import PayPal
 
 // TODO: Add PayPal button once buttons are updated as SwiftUI accessible views
```
sarahkoop: Is this done?

sarahkoop (approved):
Overall looks great - can we add an example of using a SwitUI button (and maybe one customization) to the PayPal integration docs when merged?

Demo/Demo/SwiftUIComponents/SwiftUIPayPalDemo.swift
jaxdesmarais: Ope, yeah this is done 🤦

jaxdesmarais:
> Overall looks great - can we add an example of using a `SwiftUI `button (and maybe one customization) to the PayPal integration docs when merged?

Added to the PayPal docs since hopefully these two should be merged around the same time. We don't have any of the customizations yet for the buttons on the iOS side. There is a story in the backlog for it once we get confirmation from product on which ones we need to have.

sshropshire:
If `PayPalButton` conforms to `UIViewRepresentable` directly does that clean up the API for both UIKit and Swift? e.g.

```swift
// UIKit
let button = PayPalButton()

// SwiftUI
PayPalButton {
    // button action here
}
```

jaxdesmarais:
> If `PayPalButton` conforms to `UIViewRepresentable` directly does that clean up the API for both UIKit and Swift? e.g.
> 
> ```swift
> // UIKit
> let button = PayPalButton()
> 
> // SwiftUI
> PayPalButton {
>     // button action here
> }
> ```

@sshropshire  Yep, we can consolidate this into 1 class, we would need to make the `PayPalButton` and `PayPalCreditButton` final classes due to some [idiosyncrasies of Swift](https://stackoverflow.com/questions/40270118/protocol-requirement-cannot-be-satisfied-by-a-non-final-class-because-it-uses-s). Just a different approach, but would allow us to remove the extensions as well as the `Representable` struct. 

I went with an extension to make a clearer delineation between which init/protocol conformance is needed for SwiftUI vs what is needed for UIKit. That said, I don't think there is any harm to moving it one final class. There aren't really a lot of "best practices" with UI components being used in UIKit + SwiftUI so I think we have some flexibility. Thoughts?

sshropshire:
> There aren't really a lot of "best practices" with UI components being used in UIKit + SwiftUI so I think we have some flexibility. Thoughts?

That makes sense. It might be something to keep in mind. It's just clean and keeps the word "representable" out of the name. Thinking from a docs perspective here not necessarily implementation. It'd be cool if they were named the same on the UIKit side and SwiftUI side. Everyone's slowly migrating in that direction we might catch some devs at the crossroads.

jaxdesmarais:
> > There aren't really a lot of "best practices" with UI components being used in UIKit + SwiftUI so I think we have some flexibility. Thoughts?
> 
> That makes sense. It might be something to keep in mind. It's just clean and keeps the word "representable" out of the name. Thinking from a docs perspective here not necessarily implementation. It'd be cool if they were named the same on the UIKit side and SwiftUI side. Everyone's slowly migrating in that direction we might catch some devs at the crossroads.

That's a good point about the docs. Updated! I think we are in an exciting place in that we get to define best practices here. 😄

Sources/PayPal/Buttons/PayPalButton.swift
```diff
@@ -1,13 +1,52 @@
...
     }
+
+    // MARK: - UIViewRepresentable methods
+
+    public func makeCoordinator() -> Coordinator {
+        Coordinator(action: action)
+    }
+
+    public func makeUIView(context: Context) -> UIView {
+        let view = UIView()
```
sshropshire, line 32: qq: does this still work if we return a `PayPalButton` without constraints?

UnitTests/PayPalTests/PayPalButton_Tests.swift
```diff
@@ -12,4 +12,14 @@ class PayPalButton_Tests: XCTestCase {
         let payPalButton = PayPalButton()
         XCTAssertEqual(payPalButton.backgroundColor, PaymentButtonColor.gold.rawValue)
     }
+
+    func testInit_whenSwiftUIPayPalButtonCreated_canInit() {
+        let action = { }
+        let payPalButton = PayPalButton { }
+        let coordinator = Coordinator(action: action)
```
sshropshire, line 19: nit: Could we have a separate `Coordinator` test?

fastlane/.xcovignore
sshropshire: This feels right. UI testing is still coverage and since SwiftUI is mostly visual I think it makes sense. We may be able to take automated snapshots and do manual reviews before deployment or something if we really want to make sure everything checks out.

sshropshire (approved):
🌊

Sources/PayPal/Buttons/PayPalButton.swift
jaxdesmarais: Yep, still works - doesn't error/crash or anything, just looks goofy (understandably) since a frame for the button wouldn't be provided

PR #37, opened 2022-02-03
CardClient async/await and PayPalDelegate

sshropshire:
### Reason for changes

This PR migrates away from completion handlers.

### Summary of changes

- We use `async/await` for `CardClient` since it finishes without app switching.
- We use delegation for `PayPalClient` since it finishes after an app switch occurs.

### Checklist

### Authors

- @jcnoriega
- @sshropshire


Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -27,39 +27,32 @@ class BaseViewModel: ObservableObject {
...
-                print("✅ fetched orderID: \(order.id) with status: \(order.status)")
-            case .failure(let error):
-                self.updateTitle("Your order has failed, please try again")
-                print("❌ failed to fetch orderID: \(error)")
-            }
-            DispatchQueue.main.async {
-                completion(self.orderID)
-            }
+        do {
+            let order = try await DemoMerchantAPI.sharedService.createOrder(orderParams: orderRequestParams)
```
sarahkoop, line 45: Curious about the change in error handling behavior here - Is do/catch the convention with async/await? Could we instead return a tuple with order, or error in the await block so a merchant doesn't have to catch errors?

sarahkoop: Also realized this is demo code - but meant to call this out about the CardClient.approveOrder error handling block below (line 60-75)

jcnoriega: @sarahkoop that seems to be the convention which originated in JS i think (https://javascript.info/async-await). It is the same case for `android` and `coroutines`

jaxdesmarais (commented):
This approach looks good to me and is similar to the plan we had talked about previously

Demo/Demo/ViewModels/BaseViewModel.swift
sshropshire: We could do a tuple but most of the examples I've seen have the catch block. I think it's similar with coroutines on the Kotling side as well.

One thing it does allow for is bubbling up exceptions. Like for BT when we have to "propagate" exceptions via callback, with `throw` they'll naturally propagate to the initial call site.

sshropshire: For the record, [URLSession data(for:delegate:)](https://developer.apple.com/documentation/foundation/urlsession/3767352-data) does return a tuple, but the original `completionHandler` variant has three params `(Data?, URLResponse?, Error?)`. The new `async/await` variant will throw when it fails.

jaxdesmarais: Do catch is a pretty common pattern, so it makes sense for us to use that here

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import PaymentsCore
```
jaxdesmarais, line 3: Updating this to the following should make cocoapods happy:
```
#if canImport(PaymentsCore)
import PaymentsCore
#endif
```

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -81,4 +79,24 @@ extension URLSession: URLSessionProtocol {
         let task = dataTask(with: urlRequest, completionHandler: completionHandler)
         task.resume()
     }
+
+    func performRequest(with urlRequest: URLRequest) async throws -> (Data, URLResponse) {
+        return try await data(for: urlRequest)
```
jaxdesmarais: Remove `return` - we can rely on implicit returns in Swift here since this is a 1 line function

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -248,10 +181,13 @@ class APIClient_Tests: XCTestCase {
...
         )
 
-        apiClient.fetch(endpoint: fakeRequest) { _, correlationID in
+        let accessTokenRequest = AccessTokenRequest(clientID: "")
+        do {
+            let (_, correlationID) = try await apiClient.fetch(endpoint: accessTokenRequest)
             XCTAssertEqual(correlationID, "fake-id")
-            expect.fulfill()
+        } catch {
+            print(error)
```
jaxdesmarais: Remove print

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -25,61 +24,42 @@ final class DemoMerchantAPI {
...
             do {
-                let order = try JSONDecoder().decode(Order.self, from: data)
-                completion(.success(order))
+                return try JSONDecoder().decode(Order.self, from: data)
             } catch {
-                completion(.failure(.dataParsingError))
+                throw URLResponseError.dataParsingError
             }
+        } catch {
+            throw URLResponseError.networkConnectionError
```
jaxdesmarais: This nested do/catch is a bit confusing - I wonder if there is a way to simplify this or clean it up

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -7,14 +7,13 @@ final class DemoMerchantAPI {
...
     private init() {}
 
+
     /// This function replicates a way a merchant may go about creating an order on their server and is not part of the SDK flow.
-    /// - Parameters:
-    ///   - orderParams: the parameters to create the order with
-    ///   - completion: a result object vending either the order or an error
-    func createOrder(orderParams: CreateOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
+    /// - Parameter orderParams: the parameters to create the order with
+    /// - Returns: a result object vending either the order or an error
```
jaxdesmarais: Technically, we aren't returning a result object anymore, so can probably update this docs string

.github/workflows/build.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v2
       - name: Use Xcode 13
-        run: sudo xcode-select -switch /Applications/Xcode_13.0.app
+        run: sudo xcode-select -switch /Applications/Xcode_13.2.1.app
```
sarahkoop, line 11: Does something in this addition break on Xcode 13.0 or are we just updating to use the latest

sshropshire: Yeah async/await requires Xcode 13.2.+

sarahkoop: Ah ok let's make sure to update that min version in the README and the PayPal/Card getting started docs

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -53,13 +59,12 @@ class PayPalDemoViewController: FeatureBaseViewController {
...
-                self.payPalButton.isEnabled = true
-            } else {
-                print("There was an error")
-            }
+    func createOrder() async {
+        let orderID = await baseViewModel.createOrder(amount: amountTextField.text)
+        if orderID != nil {
+            self.payPalButton.isEnabled = true
+        } else {
+            print("There was an error")
```
sarahkoop, line 61: Can we update this to be how a merchant would access the actual error and display the specific error message on the screen?

sarahkoop: Scratch that - I see we do this in the View Model.

Sources/Card/CardClient.swift
```diff
@@ -23,37 +23,26 @@ public class CardClient {
...
+                orderID: result.id,
+                lastFourDigits: result.paymentSource.card.lastDigits,
+                brand: result.paymentSource.card.brand,
+                type: result.paymentSource.card.type
+            )
+            return cardResult
         } catch {
-            // We don't expect this block to ever execute.
-            completion(.failure(CardClientError.encodingError))
+            throw error
```
sarahkoop: Do we need to `do {} catch {}` here since we just throw the error caught? Would not wrapping the block just throw the error from the method as intended?

Sources/PayPal/PayPalClient.swift
```diff
@@ -37,31 +39,37 @@ public class PayPalClient {
...
     ///   - completion: Completion block to handle buyer's approval, cancellation, and error.
-    public func start(
-        request: PayPalRequest,
-        presentingViewController: UIViewController? = nil,
-        completion: @escaping (PayPalCheckoutResult) -> Void
-    ) {
+    public func start(request: PayPalRequest, presentingViewController: UIViewController? = nil) {
         CheckoutFlow.set(config: config, returnURL: returnURL)
 
         CheckoutFlow.start(
```
sarahkoop, line 45: I'm realizing this PR is based off the NXO not the web flow branch. There were quite a few changes to PayPalClient on that branch - do we want to update this for the web flow, or not yet?

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+public protocol PayPalDelegate: AnyObject {
```
sarahkoop, line 7: I think this class needs doc strings

Sources/PaymentsCore/Networking/APIClientDecoder.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  APIClientDecoder.swift
+//  PaymentsCore
+//
+//  Created by Jose Noriega on 08/02/2022.
+//
+
```
sarahkoop: Nitpick - we should remove these auto generated header comments

Sources/PayPal/PayPalClient.swift
jcnoriega: @sarahkoop I would do it the other way around, because the Web Flow PR is blocked from merging until further notice. So i would merge this, and then "update" the web flow.

Sources/PaymentsCore/Networking/URLSession+URLSessionProtocol.swift
```diff
@@ -0,0 +1,30 @@
+//
+//  URLSession+URLSessionProtocol.swift
+//  PaymentsCore
+//
+//  Created by Steven Shropshire on 2/8/22.
+//
```
sarahkoop:
```suggestion
```

Sources/PayPal/PayPalClient.swift
sarahkoop: Ok that make sense

Sources/PaymentsCore/Networking/URLSessionProtocol.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  URLSessionProtocol.swift
+//  PaymentsCore
+//
+//  Created by Steven Shropshire on 2/8/22.
+//
```
sarahkoop:
```suggestion
```

sarahkoop (approved):
A few nitpicks but this looks great! How do we want to get these changes in the PayPal Web PR? Should we have another JIRA ticket for that?

jcnoriega:
@sarahkoop i dont think its necessary. I think we should try to update it an if we notice that it requieres to much work, then we create a ticket for that.

Demo/Demo/Networking/DemoMerchantAPI.swift
sshropshire: Good call! Cleaned up.

Sources/Card/CardClient.swift
sshropshire: We removed this.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -7,14 +7,13 @@ final class DemoMerchantAPI {
 
     private init() {}
 
+
```
jaxdesmarais: remove extra space

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -25,61 +24,43 @@ final class DemoMerchantAPI {
...
+        let data = try await data(for: urlRequest)
+        return try parseOrder(from: data)
     }
 
     /// This function replicates a way a merchant may go about authorizing/capturing an order on their server and is not part of the SDK flow.
     /// - Parameters:
     ///   - processOrderParams: the parameters to process the order with
-    ///   - completion: a result object vending either the order or an error
-    func processOrder(processOrderParams: ProcessOrderParams, completion: @escaping ((Result<Order, URLResponseError>) -> Void)) {
+    /// - Returns: a result object vending either the order or an error
```
jaxdesmarais: We don't return a result object anymore, so these docs should be updated

README.md
```diff
@@ -11,7 +11,7 @@ As the SDK is moved to general availability, we will be adding a contribution gu
 
 ## Support
 
-The PayPal iOS SDK supports a minimum deployment target of iOS 13+ and requires Xcode 13+ and macOS Big Sur 11.3+. See our [Client Deprecation policy](https://developer.paypal.com/braintree/docs/guides/client-sdk/deprecation-policy/ios/v5) to plan for updates.
+The PayPal iOS SDK supports a minimum deployment target of iOS 13+ and requires Xcode 13.2+ and macOS Big Sur 11.3+. See our [Client Deprecation policy](https://developer.paypal.com/braintree/docs/guides/client-sdk/deprecation-policy/ios/v5) to plan for updates.
```
jaxdesmarais, line 14: Can we point to the same version in our GH actions as here? Looks like for GH Actions we point to 13.2.1 and here 13.2. Does GH Actions require us to point to the patch version?

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,25 @@
...
+    /// Notify that the PayPal flow finished with a successfull result
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    ///   - didFinishWithResult: the succesfull result from the flow
+    func paypal(client paypalClient: PayPalClient, didFinishWithResult result: PayPalResult)
+
+    /// Notify that an error ocurred in the PayPal flow
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    ///   - didFinishWithError: the error returned by the paypal flow
```
jaxdesmarais:
```suggestion
    ///   - didFinishWithError: the error returned by the PayPal flow
```

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,25 @@
...
+/// PayPal delegate to handle events from PayPalClient
+public protocol PayPalDelegate: AnyObject {
+
+    /// Notify that the PayPal flow finished with a successfull result
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    ///   - didFinishWithResult: the succesfull result from the flow
+    func paypal(client paypalClient: PayPalClient, didFinishWithResult result: PayPalResult)
+
+    /// Notify that an error ocurred in the PayPal flow
```
jaxdesmarais:
```suggestion
    /// Notify that an error occurred in the PayPal flow
```

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,25 @@
...
+import PaymentsCore
+#endif
+
+/// PayPal delegate to handle events from PayPalClient
+public protocol PayPalDelegate: AnyObject {
+
+    /// Notify that the PayPal flow finished with a successfull result
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    ///   - didFinishWithResult: the succesfull result from the flow
```
jaxdesmarais:
```suggestion
    ///   - didFinishWithResult: the successful result from the flow
```

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+/// PayPal delegate to handle events from PayPalClient
+public protocol PayPalDelegate: AnyObject {
+
+    /// Notify that the PayPal flow finished with a successfull result
```
jaxdesmarais:
```suggestion
    /// Notify that the PayPal flow finished with a successful result
```

Sources/PayPal/PayPalClient.swift
```diff
@@ -37,31 +39,37 @@ public class PayPalClient {
...
             },
-            onCancel: {
-                completion(.cancellation)
+            onCancel: { [weak self] in
+                self?.notifyCancellation()
             },
-            onError: { errorInfo in
-                completion(.failure(error: PayPalError.nativeCheckoutSDKError(errorInfo)))
+            onError: { [weak self] errorInfo in
+                self?.notifyFailure(with: errorInfo)
```
jaxdesmarais: It looks like we are deallocating too quickly by using [weak self] in these closures and never hitting the delegate functions running this currently. Removing [weak self] allows us to hit the delegate functions as expected.

Sources/PayPal/PayPalDelegate.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+/// PayPal delegate to handle events from PayPalClient
+public protocol PayPalDelegate: AnyObject {
```
jaxdesmarais, line 7: It looks like we are never using the client piece of the function signatures in the delegate. I think it's probably safe to remove unless there is a specific reason that was included that's not illustrated in this PR. It'll clean up our implementation as well as the merchant implementation of our delegate if it's not needed.

README.md
sshropshire: The 13.2+ minimum callout is mostly for async await and we made the CI use the latest version of Xcode to line up with our local build environments.

Sources/PayPal/PayPalClient.swift
sshropshire: It's less of a dealloc and more of a `no retain`. This way if the VC has been deallocated the optional `?` guards against a `nil` reference when the callback executes and it becomes a no-op.

Sources/PayPal/PayPalDelegate.swift
sshropshire: Yeah we're mostly following an [iOS pattern](https://www.swiftbysundell.com/articles/delegation-in-swift/#protocols) here similar to UICollectionView, UITableView etc. This is a good way to callout to merchants that it's a delegate method.

jaxdesmarais: Gotcha, I think that's fine then, but we should make the first parameter name anonymous to match how other delegates are formatted in the [swift docs](https://docs.swift.org/swift-book/LanguageGuide/Protocols.html) and [UITableView protocol functions](https://developer.apple.com/documentation/uikit/uitableviewdelegate/1614883-tableview) vs it being a named param.

Sources/PayPal/PayPalClient.swift
jaxdesmarais: Right now, if you add a breakpoint to any of the `notify_` functions, they are not reliably being hit and therefore the delegate functions are never being called as the reference to self has already been deallocated in some cases.

sshropshire: AH ok yeah that makes sense. Does checkout `nil` out those closure references internally at some point?

jaxdesmarais: Yep, part of the shutdown steps as the paysheet closes is to call [`Checkout.resetForSession()`](https://github.paypal.com/nativexo/ios-native-checkout/blob/fdc6f913fabe600d8956dbeb6a0f0510c05e1260/Sources/PayPalCheckoutUI/Sources/Lifecycle%20and%20State/Shutdown%20Steps/ResetEnvironmentStep.swift#L27) which calls [`resetSessionCallbacks()`](https://github.paypal.com/nativexo/ios-native-checkout/blob/d2695f5f2ec886e559e4ac5cf2364de3253fd107/Sources/PayPalCheckoutUI/Sources/Lifecycle%20and%20State/Checkout.swift#L514) and sets the callbacks to `nil`

sshropshire: Nice ok yeah that's good. 3 closures = 3 hard retains lol wanted to make sure. If that cleanup didn't happen a leak would be highly probable, only one of the callbacks will be invoked.

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -4,6 +4,12 @@ import PaymentsCore
 
 class PayPalDemoViewController: FeatureBaseViewController {
 
+    lazy var payPalClient: PayPalClient = {
+        let environment = DemoSettings.environment.paypalSDKEnvironment
+        let config = CoreConfig(clientID: DemoSettings.clientID, environment: environment)
+        return PayPalClient(config: config, returnURL: DemoSettings.paypalReturnUrl)
+    }()
+
```
jaxdesmarais: We can probably remove this since we aren't using it in this class

Sources/Card/CardClient.swift
```diff
@@ -23,37 +23,23 @@ public class CardClient {
...
-                    completion(.failure(error))
-                }
-            }
-        } catch {
-            // We don't expect this block to ever execute.
-            completion(.failure(CardClientError.encodingError))
-        }
+    /// - Parameter request: The request containing the card and order id for approval
+    /// - Returns: Card result
+    /// - Throws: PayPalSDK error if approve order could not complete successfully
```
jaxdesmarais, line 28: It looks like in some of the docs strings we put that it throws and error in the returns parameter and others we seperate out returns and throws. I don't feel strongly one way or the other, but we should make it consistent in the PR.

jaxdesmarais (approved):
Couple of very small cleanup comments, otherwise this looks great! Excited to get to use async/await.

Sources/Card/CardClient.swift
sshropshire: That's a good call. I went with `Throws` to make it stand out more on the docs site.

PR #39, opened 2022-02-18
GitHub Actions Release Script

sshropshire:
### Reason for changes

This PR adds a workflow action to automate the release process.

### Summary of changes

- Add release.yml to `.github` folder
- Add initial CHANGELOG entries

### Checklist

- [x] Added a changelog entry

### Authors
- @jcnoriega
- @sshropshire

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -25,8 +25,13 @@ jobs:
       - name: Update version
         run: |
           today=$(date +'%Y-%m-%d')
+          
           sed -i '' 's/## unreleased.*/## '"${{ github.event.inputs.version }}"' ('"$today"')/' CHANGELOG.md
           sed -i '' 's/\(s\.version *= *\).*/\1"'"${{ github.event.inputs.version }}"'\"/' PayPal.podspec
+
+          plutil -replace CFBundleVersion -string ${{ github.event.inputs.version }} -- 'Demo/Demo/Info.plist'
+          plutil -replace CFBundleShortVersionString -string ${{ github.event.inputs.version }} -- 'Demo/Demo/Info.plist'
+
           # TODO: update plist files for Demo App and Core plist
           # Ref: https://github.com/braintree/braintree_ios/blob/192c05c9621a9d7ea0893a7acc9db73cca90d75a/.github/workflows/release.yml#L35
           git add .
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
...
+  * Add `APIClient`
+  * Add `APIRequest`
+  * Add `AccessTokenRequest`
+  * Add `AccessTokenResponse`
+  * Add `CoreConfig`
+  * Add `Environment`
+  * Add `HTTPHeader`
+  * Add `HTTPMethod`
+  * Add `PayPalSDKError`
+
```
sshropshire, line 12: I left out the PayPal module for now since some of the interface is in flux with the web fallback. I also imagine we'll clean up the CHANGELOG before the first GA version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          tag_name: ${{ github.event.inputs.version }}
+          release_name: ${{ github.event.inputs.version }}
+          body_path: changelog_entries.md
+          draft: false
+          prerelease: false
+      - name: Publish to CocoaPods
+        env:
+          COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
+        run: pod trunk push PayPal.podspec
```
jaxdesmarais, line 55: Curious to hear is @demerino has thoughts on a name for the SDK. Currently, it's PayPal and once we push the pod that will be its name unless we delete the pod, update the name, and push with a new name. That's certainly not an impossible change to make, but if we _have_ a name for the SDK will be better to update that now vs later since it would be a rather large breaking change for anyone that picks this up in the meantime.

CHANGELOG.md
sarahkoop: Are all these Core classes meant to be public facing?

sarahkoop (approved):

CHANGELOG.md
jaxdesmarais: They are public because we need them to be used across module, but I don't think we need to call them out as public in our changelog.

sshropshire: Got it yeah I can remove the Core piece from here for now. We can re-evaluate which Core classes should be advertised before GA for now this helps the script generate release notes.

andrespelaezp (approved):

jaxdesmarais (approved):
⛵

PR #40, opened 2022-02-22
Spike Integration Tests

sshropshire:
### Reason for changes

This PR spikes out iOS integration testing.

### Summary of changes

- Add `CardUITests` to project

### Checklist

### Authors

- @jcnoriega
- @sshropshire


Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -100,9 +100,8 @@ class BaseViewModel: ObservableObject, PayPalDelegate {
         let cardRequest = CardRequest(orderID: orderID, card: card)
 
         do {
-            let result = try await cardClient.approveOrder(request: cardRequest)
+            _ = try await cardClient.approveOrder(request: cardRequest)
             updateTitle("\(DemoSettings.intent.rawValue.capitalized) status: APPROVED")
-            await processOrder(orderID: result.orderID)
```
sshropshire, line 105: This fails every time with an HTTP 500. I think there's a problem with the sample merchant server. Ended up removing since we don't make this call on Android, and the orders v2 `approve-order` request is working correctly.

Demo/DemoUITests/CardUITests.swift
```diff
@@ -0,0 +1,31 @@
...
+
+        app.textField(named: "Expiration").tap()
+        app.typeText("0223")
+
+        app.textField(named: "CVV").tap()
+        app.typeText("123\n")
+
+        app.button(named: "Capture Order").tap()
+
+        let elementExists = app.staticText(containing: "status: APPROVED").waitForExistence(timeout: 20.0)
```
sarahkoop: We should maybe consider standardizing the timeouts when we fully implement the tests

sarahkoop (approved):

andrespelaezp (approved):
They are looking good!

Demo/DemoUITests/CardUITests.swift
sshropshire: Good call yeah. I made a helper that applies a default `20.0` second value.

PR #43, opened 2022-03-03
Magnes integration

sshropshire:
### Reason for changes

This PR integrates PayPal data collection with the SDK.

### Summary of changes

- WIP

### Checklist

- [ ] Added a changelog entry

### Authors

- @jaxdesmarais
- @sshropshire


Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    private let magnesSDK: MagnesSDKProtocol
+    private let deviceInspector: DeviceInspectorProtocol
+
+    init(magnesSDK: MagnesSDKProtocol = MagnesSDK.shared(), deviceInspector: DeviceInspectorProtocol = DeviceInspector()) {
+        self.magnesSDK = magnesSDK
+        self.deviceInspector = deviceInspector
+    }
+
+    public func collectDeviceData(additionalData: [String: String] = [:]) -> String {
```
jaxdesmarais, line 14: I think we will want to pass the env in here and get the value from the config. I know we have plans to do this on the BT SDK but since it's a breaking change it's slated for the next major version.

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,33 @@
...
+        let deviceIdentifier = deviceInspector.paypalDeviceIdentifier()
+        let params = MagnesSetupParams(
+            env: .LIVE,
+            appGuid: deviceIdentifier,
+            apnToken: "",
+            isRemoteConfigDisabled: false,
+            isBeaconDisabled: false,
+            source: .BRAINTREE
```
jaxdesmarais, line 22: Should we use a different source for this SDK?

Sources/PayPalDataCollector/DeviceInspector.swift
```diff
@@ -0,0 +1,29 @@
+import Foundation
+
+class DeviceInspector: DeviceInspectorProtocol {
+
+    func paypalDeviceIdentifier() -> String {
```
jaxdesmarais, line 5:
```suggestion
    func payPalDeviceIdentifier() -> String {
```

Sources/PayPalDataCollector/DeviceInspectorProtocol.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+protocol DeviceInspectorProtocol {
+    func paypalDeviceIdentifier() -> String
```
jaxdesmarais, line 4:
```suggestion
    func payPalDeviceIdentifier() -> String
```

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
```diff
@@ -0,0 +1,25 @@
...
+        withAdditionalData additionalData: [String: String]
+    ) throws -> MagnesResult
+}
+
+extension MagnesSDK: MagnesSDKProtocol {
+
+    func setUpWithParams(_ params: MagnesSetupParams) throws {
+        // workaround to avoid lint errors
+        try setUp(
+            setEnviroment: params.env,
```
jaxdesmarais, line 17: I know this is a magnes func, but we should probably tell them `environment` is spelled wrong. 😆

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
```diff
@@ -0,0 +1,25 @@
...
+    func collectAndSubmit(
+        withPayPalClientMetadataId cmid: String,
+        withAdditionalData additionalData: [String: String]
+    ) throws -> MagnesResult
+}
+
+extension MagnesSDK: MagnesSDKProtocol {
+
+    func setUpWithParams(_ params: MagnesSetupParams) throws {
+        // workaround to avoid lint errors
```
jaxdesmarais, line 15: What was the linter warning that this threw without the workaround? Just curious.

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: I think we could also do something like this in `MagnesSetupParams`:
```
var magnesEnvironment: MagnesSDK.Environment {
    switch CoreConfig.environment {
    case .sandbox:
        return .SANDBOX
    case .production:
        return .LIVE
    }
}
```
We could then either have the merchant pass the env in and call that var below or we could just do that for the merchant and pass `CoreConfig.environment.magnesEnvironment` into the params env variable. I think the second would be a little cleaner and the merchant wouldn't have to do anything really, we just grab that from the config for them. (Would also be worth testing this since this is just pseudo code, but I think something like that should work)

jaxdesmarais: Ah, actually I think we would need to pass the environment into the init for this class like:
```
    self.config = config
    self.magnesSDK = MagnesSDK.shared()
    self.deviceInspector = DeviceInspector()
}
```
Then either do the computed var in the `collectDeviceData` or create something like a `getMagnesEnvironment` that could pass in the env from the config and return the magnes env.

sshropshire: Yeah that's a good callout. Would we expect the merchant to instantiate `PayPalDataCollector`? Or is it something that could be done behind the scenes. Either way changing the constructor makes sense I can update it.

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: Yeah def went with `.PAYPAL`.

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
sshropshire: It was the line width greater than 200 or something. And honestly it's because of the Magnes API but it made testing easier as a result.

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
sshropshire: Yeah def good thing it's internal haha.

Sources/PayPalDataCollector/DeviceInspectorProtocol.swift
sshropshire: I can roll with that. I've seen a few variations when the first p is lowercase. The camel casing makes it more predictable though I can change this.

PR #44, opened 2022-03-04
Magnes Integration

sshropshire:
### Reason for changes

This PR integrates PayPal data collection with the SDK.

### Summary of changes

- Add PayPalDataCollector module to project.

### Checklist

### Authors

- @jaxdesmarais
- @sshropshire


Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,54 @@
...
+    // MARK: - Initializers
+    public convenience init(config: CoreConfig) {
+        self.init(config: config, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
+    }
+
+    init(config: CoreConfig, magnesSDK: MagnesSDKProtocol, deviceInspector: DeviceInspectorProtocol) {
+        self.config = config
+        self.magnesSDK = magnesSDK
+        self.deviceInspector = deviceInspector
+    }
```
jaxdesmarais: Is there a specific reason we need 2 inits here vs doing:
```
    self.config = config
    self.magnesSDK = MagnesSDK.shared()
    self.deviceInspector = DeviceInspector()
}
```

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,54 @@
...
+
+    // MARK: - Computed Properties
+    private var magnesEnvironment: MagnesSDK.Environment {
+        switch config.environment {
+        case .sandbox:
+            return .SANDBOX
+        case .production:
+            return .LIVE
+        }
+    }
```
jaxdesmarais, line 40: Nice! 😄

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,54 @@
...
+            source: .PAYPAL
+        )
+        try? magnesSDK.setUpWithParams(params)
+
+        let result = try? magnesSDK.collectDeviceData(withPayPalClientMetadataId: "", withAdditionalData: additionalData)
+        let clientMetadataId = result?.getPayPalClientMetaDataId() ?? ""
+
+        return """
+            {"correlation_id":"\(clientMetadataId)"}
+        """.trimmingCharacters(in: .whitespacesAndNewlines)
```
jaxdesmarais: Could we simplify this by doing `return "{"correlation_id":"\(clientMetadataId)"}".trimmingCharacters(in: .whitespacesAndNewlines)`?

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,54 @@
+import Foundation
+import PPRiskMagnes
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+public class PayPalDataCollector {
```
jaxdesmarais, line 8: We should probably add docs to this class and public methods in the class

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
```diff
@@ -0,0 +1,31 @@
...
+    func collectDeviceData(
+        withPayPalClientMetadataId cmid: String,
+        withAdditionalData additionalData: [String: String]
+    ) throws -> MagnesSDKResult
+}
+
+extension MagnesSDK: MagnesSDKProtocol {
+
+    func setUpWithParams(_ params: MagnesSetupParams) throws {
+        // workaround to avoid max line column width linter error
```
jaxdesmarais: I think we can remove this comment since the linter would throw a warning if we tried to single line it in the future

Sources/PayPalDataCollector/MagnesSDKProtocol.swift
```diff
@@ -0,0 +1,31 @@
...
+            disableBeacon: params.isBeaconDisabled,
+            magnesSource: params.source
+        )
+    }
+
+    func collectDeviceData(
+        withPayPalClientMetadataId cmid: String,
+        withAdditionalData additionalData: [String: String]
+    ) throws -> MagnesSDKResult {
+        return try collectAndSubmit(withPayPalClientMetadataId: cmid, withAdditionalData: additionalData)
```
jaxdesmarais: We can remove the return here and rely on Swift implicit return for single line functions

Sources/PayPalDataCollector/DeviceInspector.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+
+class DeviceInspector: DeviceInspectorProtocol {
+
+    private let accountName = "com.paypal.ios-sdk.PayPalDataCollector.DeviceGUID"
+
+    func payPalDeviceIdentifier() -> String {
+        return payPalDeviceIdentifier(newIdentifier: UUID())
```
jaxdesmarais: We can remove the return here and rely on Swift implicit return for single line functions

PayPal.podspec
```diff
@@ -22,6 +22,11 @@ Pod::Spec.new do |s|
     s.dependency "PayPalCheckout", "0.77.0"
   end
 
+  s.subspec "PayPalDataCollector" do |s|
+    s.source_files = "Sources/PayPalDataCollector/*.swift"
+    s.vendored_frameworks = "Frameworks/XCFrameworks/PPRiskMagnes.xcframework"
+  end
```
jaxdesmarais: Adding `s.dependency "PayPal/PaymentsCore"` will resolve the current cocoapods error

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: Yeah the `convenience` initializer is merchant-facing, the other one is used for the unit tests to allow mock injection.

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: Yeah I used a multi-line string here to make the JSON output as clear as possible. Making it all one line breaks the build 😞 . The alternative is to use traditional `"` string, but we would have to escape the quotes in the JSON. I like the increased readability here but escaped quotes aren't terrible.

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: 💯  docs coming up.

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: Ah, gotcha gotcha. Could we add docs around the second init that it's for testing? Just to make this clear. The public init will have docs once you add those to this class.

sshropshire: Yeah I'd be down to put a `/// for testing` comment we typically don't document non-public methods in the BT SDK that has helped keep doc maintenance costs low.

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -0,0 +1,65 @@
...
+        let params = MagnesSetupParams(
+            env: magnesEnvironment,
+            appGuid: deviceIdentifier,
+            isRemoteConfigDisabled: false,
+            isBeaconDisabled: false,
+            source: .PAYPAL
+        )
+        try? magnesSDK.setUpWithParams(params)
+
+        let result = try? magnesSDK.collectDeviceData(withPayPalClientMetadataId: "", withAdditionalData: additionalData)
```
jcnoriega, line 58: @sshropshire shouldnt we be calling `magnesSDK.collectAndSubmit()`?
reference: https://github.com/braintree/braintree_ios/blob/7600408a7c459bd104f4c9ed56b932947ffbf6bf/Sources/PayPalDataCollector/Public/PayPalDataCollector/PPDataCollector.swift#L38

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: True, yeah I guess I thought we had one in `PayPalClient` where we do the same thing (have a public init and an internal for testing - though they are set up a little different there) but maybe that was in my mind. Up to you on if you want to docs it or not.

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: Yeah, I think we just escape the quotes in the BT SDK. I think since it's only 1 param the readability is still pretty high. Also curious why one line single quotes broke the build, that's weird for sure!

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: Yeah we are under the hood but we had to add an interface to mock this out properly.

Sources/PayPalDataCollector/PayPalDataCollector.swift
sshropshire: Yeah I'd vote no. The less docs the better.

Sources/PayPalDataCollector/PayPalDataCollector.swift
jcnoriega: gotcha, missed the protocol you implemented

jcnoriega (approved):

jaxdesmarais (approved):
⛵

PR #36, opened 2022-01-27
PayPal Web Flow

jaxdesmarais:
### Reason for changes

Replace PayPal Native Checkout with the PayPal web flow using ASWebAuthenticationSession.

### Summary of changes

- Move `paypalReturnUrl` to an Environment extension in the PayPal module
- Update demo app to use new flow
- Renamed `PayPalSDKError` to `CoreSDKError` for clarity
- Added `PayPalClientError`
- Remove PayPal Checkout
- Updated PayPal to return a result object to match `CardClient`
- Updated `PayPalClient` to use `WebAuthenticationSession`
- Added `WebAuthenticationSession` to use `ASWebAuthenticationSession`
- Removed unused files
- Added unit tests and mocks for tests

### Checklist

- ~[ ] Added a changelog entry~

- @sarahkoop @jaxdesmarais @jcnoriega @andrespelaezp

PR #46, opened 2022-03-22
PayPal Web Integration

sshropshire:
### Summary of changes

 - This pr adds the paypal web fallback in a separate module

### Authors

- @jaxdesmarais
- @sshropshire


Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
```diff
@@ -1,5 +1,5 @@
 import UIKit
-import PayPal
+import PayPalWebCheckout
 import PaymentsCore
 
 class PayPalDemoViewController: FeatureBaseViewController {
```
jaxdesmarais: Should we rename this `PayPalWebCheckoutViewController` and have a seperate `PayPalNativeCheckoutViewController` to demo the PayPal checkout SDK?

jaxdesmarais: I guess if the NativeCheckout isn't being released, probably don't need to add it to the demo, but we should rename this ViewController.

jaxdesmarais (approved):
This looks good. That's for seperating out the modules/renaming/pulling in the code from #36!

One small comment but otherwise 🚢

Demo/Demo/FeatureViewControllers/PayPalDemoViewController.swift
jaxdesmarais: We should also change the demo type from PayPal > PayPal Web Checkout

Demo/Demo/DemoSettings/DemoType.swift
```diff
@@ -3,24 +3,24 @@ import SwiftUI
 
 enum DemoType: String {
     case card
-    case paypal
+    case payPalWebCheckout
```
sshropshire, line 6: There's some inconsistency with this naming convention. I'm wondering is `paypal` or `payPal` a better prefix when at the beginning of a camel case variable?

jaxdesmarais: I think we want to do `payPal` - I feel like we talked about this with Sammy :pour-one-out: at some point and all agreed on that.

sshropshire: Nice aight. I just fixed camel casing throughout to keep it consistent.

jcnoriega (approved):

PR #48, opened 2022-03-24
PayPalUI Module

sshropshire:
### Reason for changes

This PR creates the PayPalUI module for iOS to separate PayPal SDK UI elements from other modules.

### Summary of changes

- Create PayPalUI module

### Checklist


jonathajones:
What is the overall scope of `PayPalUI` going to be? I'm generally wondering how `PayPalUI` would interact with the future inclusion of NXO, given that NXO is a fully UI experience.

sshropshire:
> What is the overall scope of `PayPalUI` going to be? I'm generally wondering how `PayPalUI` would interact with the future inclusion of NXO, given that NXO is a fully UI experience.

I'm actually not sure what the boundary will be by the time we reach GA. The ticket was to de-couple the buttons to give merchants an option to either use their own buttons or use a pre-built SDK button.

Package.swift
```diff
@@ -17,6 +17,10 @@ let package = Package(
         //    name: "PayPalNativeCheckout",
         //    targets: ["PayPalNativeCheckout"]
         //),
+        .library(
+            name: "PayPalUI",
+            targets: ["PayPalUI"]
```
jaxdesmarais, line 22: Just a general question, but will we include other UI Components in this module in the future? For example Venmo buttons, text fields, validation logic potentially, or the Drop-in? If so, calling it `PayPalUI` will almost certainly be confusing if we want this to be the entry point for merchants who don't want to handle visual elements themselves.

PayPal.xcodeproj/xcshareddata/xcschemes/PayPalUITests.xcscheme
```diff
@@ -0,0 +1,52 @@
...
+      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
+      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      shouldUseLaunchSchemeArgsEnv = "YES">
+      <Testables>
+         <TestableReference
+            skipped = "NO">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "BCFAC70A27ED043100C3AF00"
+               BuildableName = "PayPalTests.xctest"
```
jaxdesmarais: `PayPalTests` > `PayPalUITests`?

Package.swift
sshropshire: Yeah I'm not sure I'd say the name is subject to change. We called it PayPalUI on Android so I just went for it.

PayPal.xcodeproj/xcshareddata/xcschemes/PayPalUITests.xcscheme
sshropshire: Yeah this is odd. Can I change this value from an Xcode GUI?

Package.swift
sarahkoop: I _think_ `PayPalUI` was meant to be UI components related to the PayPal flow, and each payment method with UI components would have it's own UI module (i.e. `VenmoUI`), since it wouldn't make sense for merchants not offering Venmo to have to include a bunch of Venmo related UI elements in order to get PayPal buttons. But this is another reason we need to differentiate the SDK name as a whole from the PayPal payment flow module names.

sarahkoop:
We should add this module to the `podspec` and `Package.swift` too

sshropshire:
> We should add this module to the `podspec` and `Package.swift` too

Yup it should be in there.

sarahkoop (approved):

Package.swift
jaxdesmarais: Ah, gotcha gotcha. Yeah some of those may end up pretty small. Like a Venmo module would probably just be 1 button? Unless there are other UI elements for Venmo that I don't know about. 🤷‍♂️ Wonder if it's also worth exploring NXO decoupling the smart payment buttons into it's own module vs us duplicating what already exists in the package size of NXO but isn't being used.

But yes, totally agree on differentiating the name of the SDK!

PayPal.xcodeproj/xcshareddata/xcschemes/PayPalUITests.xcscheme
jaxdesmarais: Yeah, you should be able to change it in the Build Settings of the module. I think this is pulled from the "target name". If you do a search in the build settings for `PayPalTests` there is a user defined target name at the bottom of the list. Updating that should update it everywhere it needs to be updated.

jaxdesmarais: Hmmm. Looks like this one didn't get updated. May need to just update it manually outside of the GUI?

sshropshire: It should be good now with the caveat of Xcode adding more xml to the file. I guess it's necessary, might have to do with the new version of Xcode.

jaxdesmarais (approved):
Thanks for addressing comments!

PR #49, opened 2022-03-30
Remove PaymentsCore dependency on PayPalDataCollector

jcnoriega:
### Reason for changes
This PR removes the PaymentsCore dependency on PayPalDataCollector

### Summary of changes

- We expose PayPalDataCollector environment to the merchant, which it transforms to Magnes environment, so that we can not depend on PaymentsCore environment

### Checklist

- [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Sources/PayPalDataCollector/PayPalDataCollector.swift
+++ b/Sources/PayPalDataCollector/PayPalDataCollector.swift
@@ -1,42 +1,35 @@
 import Foundation
 import PPRiskMagnes
-#if canImport(PaymentsCore)
-import PaymentsCore
-#endif
 
 /// Enables you to collect data about a customer's device and correlate it with a session identifier on your server.
 public class PayPalDataCollector {
 
     // MARK: - Properties
 
-    private let config: CoreConfig
     private let magnesSDK: MagnesSDKProtocol
     private let deviceInspector: DeviceInspectorProtocol
+    private let magnesEnvironment: MagnesSDK.Environment
 
     // MARK: - Initializers
 
     /// Construct an instance to collect device data to send to your server.
     /// - Parameter config: configuration to use when collecting device data
-    public convenience init(config: CoreConfig) {
-        self.init(config: config, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
+    public convenience init(environment: PayPalDataCollectorEnvironment) {
+        self.init(environment: environment, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
     }
 
     /// internal constructor for testing
-    init(config: CoreConfig, magnesSDK: MagnesSDKProtocol, deviceInspector: DeviceInspectorProtocol) {
-        self.config = config
+    init(environment: PayPalDataCollectorEnvironment, magnesSDK: MagnesSDKProtocol, deviceInspector: DeviceInspectorProtocol) {
+        magnesEnvironment = {
+            switch environment {
+            case .sandbox:
+                return .SANDBOX
+            case .production:
+                return .LIVE
+            }
+        }()
         self.magnesSDK = magnesSDK
         self.deviceInspector = deviceInspector
-    }
-
-    // MARK: - Computed Properties
-
-    private var magnesEnvironment: MagnesSDK.Environment {
-        switch config.environment {
-        case .sandbox:
-            return .SANDBOX
-        case .production:
-            return .LIVE
-        }
     }
 
     // MARK: - PayPalDataCollector
--- a/Sources/PayPalDataCollector/PayPalDataCollector.swift
+++ b/Sources/PayPalDataCollector/PayPalDataCollector.swift
@@ -20,14 +20,7 @@ public class PayPalDataCollector {
 
     /// internal constructor for testing
     init(environment: PayPalDataCollectorEnvironment, magnesSDK: MagnesSDKProtocol, deviceInspector: DeviceInspectorProtocol) {
-        magnesEnvironment = {
-            switch environment {
-            case .sandbox:
-                return .SANDBOX
-            case .production:
-                return .LIVE
-            }
-        }()
+        self.magnesEnvironment = environment.magnesEnvironment
         self.magnesSDK = magnesSDK
         self.deviceInspector = deviceInspector
     }
--- a/Sources/PayPalDataCollector/PayPalDataCollector.swift
+++ b/Sources/PayPalDataCollector/PayPalDataCollector.swift
@@ -13,7 +13,7 @@ public class PayPalDataCollector {
     // MARK: - Initializers
 
     /// Construct an instance to collect device data to send to your server.
-    /// - Parameter config: configuration to use when collecting device data
+    /// - Parameter environment: enviroment for the data collector
     public convenience init(environment: PayPalDataCollectorEnvironment) {
         self.init(environment: environment, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
     }
--- a/Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
+++ b/Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
@@ -1,0 +1,6 @@
+import Foundation
+
+public enum PayPalDataCollectorEnvironment {
+    case sandbox
+    case production
+}
--- a/Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
+++ b/Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
@@ -1,6 +1,16 @@
-import Foundation
+import PPRiskMagnes
 
+/// Enum of environments to use with PayPalDataCollector
 public enum PayPalDataCollectorEnvironment {
     case sandbox
     case production
+    
+    internal var magnesEnvironment: MagnesSDK.Environment {
+        switch self {
+        case .sandbox:
+            return .SANDBOX
+        case .production:
+            return .LIVE
+        }
+    }
 }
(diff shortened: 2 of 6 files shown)


andrespelaezp (approved):
LGTM!

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -1,44 +1,37 @@
...
 
     // MARK: - Initializers
 
     /// Construct an instance to collect device data to send to your server.
-    /// - Parameter config: configuration to use when collecting device data
-    public convenience init(config: CoreConfig) {
-        self.init(config: config, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
+    /// - Parameter environment: enviroment for the data collector
+    public convenience init(environment: PayPalDataCollectorEnvironment) {
+        self.init(environment: environment, magnesSDK: MagnesSDK.shared(), deviceInspector: DeviceInspector())
```
jaxdesmarais, line 18: 🤔 Do we want to expose this for merchants? My thought was that for the external merchant facing piece, we could just have them pass nothing, just simply do `let dataCollector = DataCollector()`. 

A merchant shouldn't really need to know that the Magnes SDK needs an environment. Then our internal implementation when the SDKs (NXO and Unified) instantiate the data collector in PayPal (or wherever else it's used), we would pass in the environment on behalf of the merchant via the CoreConfig for Unified and the CheckoutConfig for NXO then pass the respective environment to Magnes when we internally initialize the data collector.

Essentially thinking, I don't know that merchants need to know about this and we can do the obfuscating in our respective SDKs. Thoughts?

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
+
+public enum PayPalDataCollectorEnvironment {
```
sarahkoop, line 4: Doc strings for this enum

Sources/PayPalDataCollector/PayPalDataCollector.swift
sarahkoop: I think this is needed because we are removing Core as a dependency so Data Collector wont have access to the environment from CoreConfig

jaxdesmarais: Yeah, I think I was thinking that something like the PayPal module could instantiate the DataCollector with the environment from the config passed into the PayPalClient for example vs having a merchant pass the environment both on the CoreConfig and when setting up the data collector. 

If that doesn't make sense or will be more work then it's worth, that's fine, just trying to eliminate a potential mistake a merchant could make that I assume would make it so that device data wouldn't be passed correctly.

jaxdesmarais (commented):
Also need to update the Package.swift and PayPal.podspec to remove core as a dependency

Sources/PayPalDataCollector/PayPalDataCollector.swift
```diff
@@ -1,44 +1,37 @@
...
-        self.config = config
+    init(environment: PayPalDataCollectorEnvironment, magnesSDK: MagnesSDKProtocol, deviceInspector: DeviceInspectorProtocol) {
+        magnesEnvironment = {
+            switch environment {
+            case .sandbox:
+                return .SANDBOX
+            case .production:
+                return .LIVE
+            }
+        }()
```
jaxdesmarais: I think we can clean this up a bit by moving this logic to the PayPalDataCollectorEnvironment enum. Then in there you could do:
```
var magnesEnvironment: MagnesSDK.Environment {
    switch self {
    case .sandbox:
        return .SANDBOX
    case .production:
        return .LIVE
    }
}
```
And in the init here just do: `magnesEnvironment = environment.magnesEnvironment`

Sources/PayPalDataCollector/PayPalDataCollector.swift
jcnoriega: Why would the PayPal module instantiate DataCollector? Isnt the responsability of the merchant to insantiate and use DataCollector?
It is my understanding that the merchant could, for example, use DataCollector in conjunction with the Card module, so they would need to pass a enviornment to it. Or am I missing something?

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
```diff
@@ -0,0 +1,6 @@
+import Foundation
```
jaxdesmarais: This shouldn't need to import foundation I don't think

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: We might not need to unless we were doing it to do something like passing the env so a merchant doesn't have to. I know on Braintree support, merchants mess up the data collector _a lot_ so just trying to think through ways that we can reduce that friction for merchants, if possible. Like I said, this may not be possible just trying to think through possible options we have to eliminate merchant errors.

jcnoriega: @jaxdesmarais to be honest, im not familiarized very much with how magnes is used, but could we just default it to production? 
In this implementation merchants can only instantiate it and use the method `collectDevice` and thats it. But if they screw it up so bad, perhaps just defaulting it to production is the way to go.

sshropshire: Ideally the merchant won't have to explicitly collect data, and we do it on their behalf internally. That would simplify the onboarding process significantly.

jonathajones: I think that the environment parameter is necessary, and we shouldn't use a default value here. In order to get the interface you're looking for @jaxdesmarais, we should create an extension in `Core` that handles the injection using the `CoreConfig` object. for example, in core you could: 

```swift
// SomewhereInCore.swift
// 

extension PayPalDataCollector {
  func map(environment: Environment) -> PayPalDataCollectorEnvironment {
    switch environment {
       ...
    }
  }

  public convenience init(config: CoreConfig) {
    self.init(environment: mapEnvironments(environment: config.environment)
  }
}
```
This way, the merchant could take advantage of the same interface you have now

jonathajones (approved):

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: Agree, Steven. That would be potentially best case scenario if merchants don't need to think about it at all. 

And nice, Jon! Yeah, that's what I was thinking we could do something like just couldn't quite explain it, but that would be a good way to go about it! :teamwork:

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
```diff
@@ -0,0 +1,16 @@
+import PPRiskMagnes
+
+/// Enum of environments to use with PayPalDataCollector
+public enum PayPalDataCollectorEnvironment {
+    case sandbox
+    case production
+
+    internal var magnesEnvironment: MagnesSDK.Environment {
```
jaxdesmarais: Swift defaults to internal if nothing is provided, so we can remove internal here. 😄

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
```diff
@@ -0,0 +1,16 @@
+import PPRiskMagnes
+
+/// Enum of environments to use with PayPalDataCollector
+public enum PayPalDataCollectorEnvironment {
```
vladfilyakov, line 4: Curious why we didn't make it `PayPalDataCollector.Environment` (place it inside `PayPalDataCollector` type).

Sources/PayPalDataCollector/PayPalDataCollector.swift
vladfilyakov: How do we expect merchants to use `PayPalDataCollector`?

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
vladfilyakov: +1

Sources/PayPalDataCollector/PayPalDataCollector.swift
jcnoriega: @jonathajones @jaxdesmarais making the extension function would make Core dependant on PayPalDataCollector right? We would just be reversing dependencies.

@vladfilyakov im not really sure. The idea is that the merchant calls `dataCollector.collectDeviceData()` and that will return some `metadata`. Here is were I im a bit confused; they use that metadata to send it to their own server to then send it to paypal, OR they attach that metadata to whatever request they will do to `paypal`. So the idea i think is that we offer `dataCollector` as an independet feature and its the responsability of the merchant to use it correctly. 
Is this correct @demerino?

Sources/PayPalDataCollector/PayPalDataCollectorEnvironment.swift
jcnoriega: @vladfilyakov just following the convention in this SDK, where every enum is in a diferent file.

Sources/PayPalDataCollector/PayPalDataCollector.swift
jaxdesmarais: Your understanding is correct, @jcnoriega, doing an extension within core would make it so that Core has a dependency of PayPalDataCollector. The more that I thought about it last night, the less I liked it as that would be pulling in a dependency that merchants may not need, which is the point of the modularity. 

I think your solution may ultimately be the best approach to get everything we want and we will just need to be explicit in our docs to make sure folks are passing in the correct environment in both the config and to the data collector.

jaxdesmarais (approved):
Thanks for making this update!

PR #53, opened 2022-05-26
Payment Button Configuration Demo in UKit

jcnoriega:
### Reason for changes
We need a way to test the button configurations in our demo app

### Summary of changes

- Adds a demo screen to test all the payment button configurations on UKit. It is a copy of the NXO demo app.

- @jcnoriega

### DIFF
--- a/Demo/Demo/DemoSettings/DemoType.swift
+++ b/Demo/Demo/DemoSettings/DemoType.swift
@@ -23,8 +23,10 @@ enum DemoType: String {
         switch self {
         case .card:
             return AnyView(SwiftUICardDemo())
-        case .payPalWebCheckout, .paymentButtonCustomization: // we do this until swiftUI is implemented
+        case .payPalWebCheckout:
             return AnyView(SwiftUIPayPalDemo())
+        case .paymentButtonCustomization:
+            return AnyView(SwiftUIPaymentButtonDemo())
         }
     }
 }
--- a/Demo/Demo/DemoSettings/DemoType.swift
+++ b/Demo/Demo/DemoSettings/DemoType.swift
@@ -4,6 +4,7 @@ import SwiftUI
 enum DemoType: String {
     case card
     case payPalWebCheckout
+    case paymentButtonCustomization
 
     var viewController: UIViewController {
         let baseViewModel = BaseViewModel()
@@ -13,6 +14,8 @@ enum DemoType: String {
             return CardDemoViewController(baseViewModel: baseViewModel)
         case .payPalWebCheckout:
             return PayPalWebCheckoutViewController(baseViewModel: baseViewModel)
+        case .paymentButtonCustomization:
+            return PaymentButtonCustomizationViewController()
         }
     }
 
@@ -20,7 +23,7 @@ enum DemoType: String {
         switch self {
         case .card:
             return AnyView(SwiftUICardDemo())
-        case .payPalWebCheckout:
+        case .payPalWebCheckout, .paymentButtonCustomization: // we do this until swiftUI is implemented
             return AnyView(SwiftUIPayPalDemo())
         }
     }
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -1,0 +1,96 @@
+import PayPalUI
+
+extension PayPalPayLaterButton.Color {
+
+    static func allCases() -> [PayPalPayLaterButton.Color] {
+        return [
+            PayPalPayLaterButton.Color.gold,
+            PayPalPayLaterButton.Color.white,
+            PayPalPayLaterButton.Color.black,
+            PayPalPayLaterButton.Color.silver,
+            PayPalPayLaterButton.Color.blue
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return allCases().map { $0.rawValue }
+    }
+}
+
+extension PayPalButton.Color {
+
+    static func allCases() -> [PayPalButton.Color] {
+        return [
+            PayPalButton.Color.gold,
+            PayPalButton.Color.white,
+            PayPalButton.Color.black,
+            PayPalButton.Color.silver,
+            PayPalButton.Color.blue
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return allCases().map { $0.rawValue }
+    }
+}
+
+extension PayPalCreditButton.Color {
+
+    static func allCases() -> [PayPalCreditButton.Color] {
+        return [
+            PayPalCreditButton.Color.white,
+            PayPalCreditButton.Color.black,
+            PayPalCreditButton.Color.darkBlue
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return allCases().map { $0.rawValue }
+    }
+}
+
+extension PaymentButtonEdges {
+
+    static func allCases() -> [PaymentButtonEdges] {
+        return [
+            PaymentButtonEdges.hardEdges,
+            PaymentButtonEdges.softEdges,
+            PaymentButtonEdges.rounded
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return allCases().map { $0.description }
+    }
+}
+
+extension PaymentButtonSize {
+
+    static func allCases() -> [PaymentButtonSize] {
+        return  [
+            PaymentButtonSize.mini,
+            PaymentButtonSize.collapsed,
+            PaymentButtonSize.expanded,
+            PaymentButtonSize.full
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return allCases().map { $0.description }
+    }
+}
+
+extension PaymentButtonFundingSource {
+
+    static func allCases() -> [PaymentButtonFundingSource] {
+        return  [
+            PaymentButtonFundingSource.payPal,
+            PaymentButtonFundingSource.payLater,
+            PaymentButtonFundingSource.credit
+        ]
+    }
+
+    static func allCasesAsString() -> [String] {
+        return PaymentButtonFundingSource.allCases().map { $0.rawValue }
+    }
+}
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -3,7 +3,7 @@ import PayPalUI
 extension PayPalPayLaterButton.Color {
 
     static func allCases() -> [PayPalPayLaterButton.Color] {
-        return [
+        [
             PayPalPayLaterButton.Color.gold,
             PayPalPayLaterButton.Color.white,
             PayPalPayLaterButton.Color.black,
@@ -13,14 +13,14 @@ extension PayPalPayLaterButton.Color {
     }
 
     static func allCasesAsString() -> [String] {
-        return allCases().map { $0.rawValue }
+        allCases().map { $0.rawValue }
     }
 }
 
 extension PayPalButton.Color {
 
     static func allCases() -> [PayPalButton.Color] {
-        return [
+        [
             PayPalButton.Color.gold,
             PayPalButton.Color.white,
             PayPalButton.Color.black,
@@ -30,14 +30,14 @@ extension PayPalButton.Color {
     }
 
     static func allCasesAsString() -> [String] {
-        return allCases().map { $0.rawValue }
+        allCases().map { $0.rawValue }
     }
 }
 
 extension PayPalCreditButton.Color {
 
     static func allCases() -> [PayPalCreditButton.Color] {
-        return [
+        [
             PayPalCreditButton.Color.white,
             PayPalCreditButton.Color.black,
             PayPalCreditButton.Color.darkBlue
@@ -45,14 +45,14 @@ extension PayPalCreditButton.Color {
     }
 
     static func allCasesAsString() -> [String] {
-        return allCases().map { $0.rawValue }
+        allCases().map { $0.rawValue }
     }
 }
 
 extension PaymentButtonEdges {
 
     static func allCases() -> [PaymentButtonEdges] {
-        return [
+        [
             PaymentButtonEdges.hardEdges,
             PaymentButtonEdges.softEdges,
             PaymentButtonEdges.rounded
@@ -60,14 +60,14 @@ extension PaymentButtonEdges {
     }
 
     static func allCasesAsString() -> [String] {
-        return allCases().map { $0.description }
+        allCases().map { $0.description }
     }
 }
 
 extension PaymentButtonSize {
 
     static func allCases() -> [PaymentButtonSize] {
-        return  [
+        [
             PaymentButtonSize.mini,
             PaymentButtonSize.collapsed,
             PaymentButtonSize.expanded,
@@ -76,14 +76,14 @@ extension PaymentButtonSize {
     }
 
     static func allCasesAsString() -> [String] {
-        return allCases().map { $0.description }
+        allCases().map { $0.description }
     }
 }
 
 extension PaymentButtonFundingSource {
 
     static func allCases() -> [PaymentButtonFundingSource] {
-        return  [
+        [
             PaymentButtonFundingSource.payPal,
             PaymentButtonFundingSource.payLater,
             PaymentButtonFundingSource.credit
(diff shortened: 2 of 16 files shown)


Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,233 @@
+//
+//  PaymentButtonCustomizationViewController.swift
+//  Demo
+//
+//  Created by Jose Noriega on 25/05/2022.
+//
+
```
sarahkoop:
```suggestion
```

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+        )
+        return stackView
+    }()
+
+    // MARK: - Init
+
+    //    override init(baseViewModel: BaseViewModel) {
+    //        super.init(baseViewModel: baseViewModel)
+    //        configure()
+    //    }
```
sarahkoop: Should this be removed?

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+            return PayPalCreditButton.Color.allCases.map { $0.rawValue }
+        }
+    }
+
+    private func setupPaymentButton() -> PaymentButton {
+        let fundingSource = PaymentButtonFundingSource.allCases[spbFundingPicker.selectedSegmentIndex]
+        let edges = PaymentButtonEdges.allCases[spbEdgesPicker.selectedSegmentIndex]
+        let size = PaymentButtonSize.allCases[spbSizePicker.selectedSegmentIndex]
+
+        let spb: PaymentButton
```
sarahkoop: Can we make `spb` a more explicit variable name?

Sources/PayPalUI/PayPalPayLaterButton.swift
```diff
@@ -5,7 +5,7 @@ import SwiftUI
 public final class PayPalPayLaterButton: PaymentButton, UIViewRepresentable {
 
     /// Available colors for PayPalPayLaterButton.
-    public enum Color: String {
+    public enum Color: String, CaseIterable {
```
sarahkoop: Does `CaseIterable` have impact to merchant integrations? I'm not sure we want to change the public facing API in order to have our demo app be cleaner

jcnoriega: This is also on NXO, I took it away at first because i didnt know what it was. Then doing the demo i realized, and IMO it it does create a better feature, by allowing them to iterate over the values, and keeps us better aligned with NXO.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
sshropshire: +1 Here we should always remove the Xcode generated header.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
sshropshire: +1

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
sshropshire: +1

Sources/PayPalUI/PayPalPayLaterButton.swift
sshropshire: +1. I don't see a use case where iterating over color enums would be necessary for merchants. We should take this out to minimize the surface of the SDK API.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
andrespelaezp: Should we have a standard IDE template for new files? I sometimes forget to delete this, or other generated code, and also people tend to have their own formatting style that may look different to what the SDK implements

sshropshire: That would be ideal. I'm unaware of a way to have a versioned template that Xcode can reference when generating a new file header, but there must be a way.

Removing the headers has been the easiest way so far, but I agree that it's a good idea to keep improving our workflows whenever possible.

Demo/Demo/Extensions/UIViewController+Extension.swift
```diff
@@ -9,4 +9,11 @@ extension UIViewController {
             viewController.removeFromParent()
         }
     }
+
+    func label(_ text: String) -> UILabel {
```
sshropshire:
```suggestion
    func makeLabel(_ text: String) -> UILabel {
```
To align with [Swift Naming Conventions](https://www.swift.org/documentation/api-design-guidelines/#begin-factory-name-with-make)

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,221 @@
...
+
+
+    // MARK: - Views
+
+    /// Scroll view which will auto size
+    let scrollView = UIScrollView(frame: .zero)
+
+    lazy var paymentButton: PaymentButton = setupPaymentButton()
+
+    lazy var spbFundingPicker: UISegmentedControl = {
```
sshropshire:
```suggestion
    lazy var fundingPicker: UISegmentedControl = {
```
Same here, the `spb` prefix isn't very expressive and we can align with [naming conventions](https://www.swift.org/documentation/api-design-guidelines/#avoid-abbreviations) as well.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,221 @@
...
+
+    init() {
+        super.init(nibName: nil, bundle: nil)
+        configure()
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
```
sshropshire, line 74: If we ~override~ call `configure()` in `viewDidLoad` these shouldn't be needed.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,221 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        reloadColorPicker()
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+    }
```
sshropshire:
```suggestion
```

Sources/PayPalUI/PayPalPayLaterButton.swift
sshropshire: We should follow the [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle here. We've had experience in the past where the Demo was too tightly coupled to the BT SDK. It's a good rule of thumb to avoid tight coupling like this because the merchant wins in the end.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
kgangineni: btw, what's SPB?

kgangineni (approved):
LGTM
running the branch functionality looks good
I didn't liked going to settings and enabling feature to take a look on buttons this is different discussion
Please wait for other's review

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
jcnoriega: @kgangineni achronym for Smart Payment Button

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
jcnoriega: i tried it but it crashes :(

Sources/PayPalUI/PayPalPayLaterButton.swift
jcnoriega: Sure thing! i did some extension functions on the demo app to provide what we need

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
jcnoriega: i think there is a conflict in rendering, its like its trying to do some stuff when the views havent been stacked or anything yet.

Demo/Demo/DemoSettings/DemoType.swift
```diff
@@ -13,14 +14,16 @@ enum DemoType: String {
...
+            return PaymentButtonCustomizationViewController()
         }
     }
 
     var swiftUIView: some View {
         switch self {
         case .card:
             return AnyView(SwiftUICardDemo())
-        case .payPalWebCheckout:
+        case .payPalWebCheckout, .paymentButtonCustomization: // we do this until swiftUI is implemented
```
jaxdesmarais: I think instead we can have this as an empty case and just return either something like `EmptyViews` and turn this note into a TODO comment or create a SwiftUI view for this with a TODO at the top and something like this for the body:
```
    var body: some View {
        ZStack {
            Text("WIP - SwiftUI Button Customization View")
        }
    }
```

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -0,0 +1,96 @@
+import PayPalUI
+
+extension PayPalPayLaterButton.Color {
```
jaxdesmarais, line 3: Since all the functions in here are single line functions we can remove the return keyword and rely on Swift's implicit return.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,210 @@
+import UIKit
+import PayPalUI
+
+class PaymentButtonCustomizationViewController: UIViewController {
+
+
```
jaxdesmarais: Remove extra space here

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
jaxdesmarais: Yeah, we don't need this in the other views because they conform to `FeatureBaseViewController` which has this in it. Since this view doesn't I believe it's needed. 😢

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,210 @@
...
+            )
+        }
+        colorPicker.selectedSegmentIndex = 0
+    }
+
+    private func colorPickerSegments(with funding: PaymentButtonFundingSource) -> [String] {
+        switch funding {
+        case .payPal:
+            return PayPalButton.Color.allCasesAsString()
+        case .payLater:
```
jaxdesmarais, line 167: We should add a space before `case .payLater`

jcnoriega: @jaxdesmarais if i add a space i get a swiftlint warning. it says that switch statements should vertically align with their enclosing statement

jcnoriega: @jaxdesmarais never mind, i just undrestood what you said lol

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+            ),
+            stackView.trailingAnchor.constraint(
+                equalTo: scrollView.contentLayoutGuide.trailingAnchor
+            ),
+            stackView.topAnchor.constraint(
+                equalTo: scrollView.contentLayoutGuide.topAnchor
+            ),
+            stackView.bottomAnchor.constraint(
+                equalTo: scrollView.contentLayoutGuide.bottomAnchor
+            )
```
jaxdesmarais: Nit: I think we can single line these and not hit our line length linter warnings

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    // MARK: - Target / Actions
+
+    @objc private func reloadSmartPaymentButton(_ sender: UISegmentedControl) {
+        if sender == fundingPicker {
+            reloadColorPicker()
+        }
+        paymentButton.removeFromSuperview()
+
```
jaxdesmarais: remove space

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        let size = PaymentButtonSize.allCases()[sizePicker.selectedSegmentIndex]
+
+        let paymentButton: PaymentButton
+
+        switch fundingSource {
+
+        case .payPal:
+            let color = PayPalButton.Color.allCases()[colorPicker.selectedSegmentIndex]
+            paymentButton = PayPalButton(color: color, edges: edges, size: size)
+
```
jaxdesmarais, line 186: remove extra space

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
```diff
@@ -0,0 +1,12 @@
+import SwiftUI
+
+
```
jaxdesmarais: remove extra space

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
```diff
@@ -0,0 +1,12 @@
...
+@available(iOS 13.0.0, *)
```
jaxdesmarais: Do we get a warning without this? I thought we only supported iOS 13+ but maybe I am misremembering.

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    private func setupPaymentButton() -> PaymentButton {
+        let fundingSource = PaymentButtonFundingSource.allCases()[fundingPicker.selectedSegmentIndex]
+        let edges = PaymentButtonEdges.allCases()[edgesPicker.selectedSegmentIndex]
+        let size = PaymentButtonSize.allCases()[sizePicker.selectedSegmentIndex]
+
+        let paymentButton: PaymentButton
+
+        switch fundingSource {
+
```
jaxdesmarais: Nit: typically we would have no space between switch and the first case and then spaces between each case like so:
```
switch thing {
case 1:
    return 1

case 2:
    return 2

case 3:
    return 3
}

File: Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift [hunk 19]
```diff
@@ -0,0 +1,213 @@
...
+                at: colorPicker.numberOfSegments,
+                animated: false
+            )
+        }
+        colorPicker.selectedSegmentIndex = 0
+    }
+
+    private func colorPickerSegments(with funding: PaymentButtonFundingSource) -> [String] {
+        switch funding {
+
```
jaxdesmarais: remove space between switch and first case

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
jcnoriega: i removed it and no warning!

jaxdesmarais (approved):
Thanks for addressing comments! This looks good!

PR #56, opened 2022-06-07
Eligibility query

kgangineni:
### Reason for changes

1. Initial GraphQL setup
2. Fetch Eligibility Query

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

- @kgangineni @andrespelaezp


Sources/PaymentsCore/Networking/GraphQL/Common/GraphQLClient.swift
```diff
@@ -0,0 +1,64 @@
...
+//  GraphQLClient.swift
+//  Card
+//
+//  Created by Andres Pelaez on 19/05/22.
+//
+
+import Foundation
+import PaymentsCore
+
+internal class GraphQLClient {
```
jcnoriega: i think we dont need the `internal` access modifier, as it is the default in swift

Sources/PaymentsCore/Networking/GraphQL/Common/GraphQLClient.swift
```diff
@@ -0,0 +1,64 @@
...
+        urlRequest.httpMethod = HTTPMethod.post.rawValue
+        urlRequest.httpBody = requestBody
+        return urlRequest
+    }
+    
+    func headers() -> [String: String]{
+        return [
+            "Content-type": "application/json",
+            "Accept": "application/json",
+            "x-app-name": "nativecheckout",
```
jcnoriega: `nativecheckout` in app-name required? what happens if we put `northstar` or something else? this could be a good spike ticket to actually get our app name in here, instead of nativecheckout

Sources/PaymentsCore/Networking/GraphQL/Common/GraphQLClient.swift
```diff
@@ -0,0 +1,64 @@
...
+            "Origin": "https://www.sandbox.paypal.com"
```
jcnoriega: shouldnt this be dynamic depending on the environment?

jcnoriega, line 50: shouldnt this be dynamic depending on the environment?

Sources/PaymentsCore/Networking/GraphQL/FundingEligibility/FundingEligibilityQuery.swift
```diff
@@ -0,0 +1,83 @@
...
+            }
+            paylater{
+                  eligible
+                  reasons
+            }
+            credit{
+                eligible
+                reasons
+            }
+            }
```
jcnoriega: formatting

jcnoriega, line 70: formatting

Sources/PaymentsCore/api/EligibilityAPI.swift
```diff
@@ -0,0 +1,35 @@
...
+
+public class EligibilityAPI{
+    private var graphQLClient: GraphQLClient
+    private var coreConfig: CoreConfig
+    public init(coreConfig: CoreConfig){
+        self.coreConfig = coreConfig
+        graphQLClient = GraphQLClient(environment: coreConfig.environment)
+    }
+    
+    public func checkEligibility() async throws -> APIResult<Eligibility>{
```
jcnoriega: i think here we should do what we do in Android with `suspend` functions, which is the equivalent.
We return the value or throw an exception. @jaxdesmarais @sshropshire thoughts? Im not that well familiarized with iOS best practices.

jcnoriega, line 12: i think here we should do what we do in Android with `suspend` functions, which is the equivalent.
We return the value or throw an exception. @jaxdesmarais @sshropshire thoughts? Im not that well familiarized with iOS best practices.

jcnoriega:
remove all `internal` access modifiers. Also, i think github bugged and posted all my comments twice

jcnoriega:
remove all `internal` access modifiers. Also, i think github bugged and posted all my comments twice

Sources/PaymentsCore/api/EligibilityAPI.swift
jaxdesmarais: It looks like we are creating our own [`Result` type](https://www.swiftbysundell.com/articles/the-power-of-result-types-in-swift/) here vs using Swift's built in `Result` type. Is there a reason we are implementing our own here? I think it's important that we also demonstrate how a merchant could handle success/failure in our demo.

I believe async/away as used here would be similar to Kotlin's suspend but maybe I am not fully understanding coroutines/the suspend functionality.

sshropshire: Yeah coroutines and async/await are essentially the same in this case.

I agree if we don't need the `APIResult` wrapper type we should return `Eligibility` directly to simplify the interface. Then we can `throw` in the case of an error.

Sources/PaymentsCore/Networking/GraphQL/Common/Models/GraphQLQueryResponse.swift
```diff
@@ -0,0 +1,5 @@
+
+
+internal struct GraphQLQueryResponse<T: Codable>: Codable {
+    let data: T?
+}
```
mattwylder, line 4: Will more properties be added to this struct? Also we don't need `internal`, since it's the default.

Sources/PaymentsCore/Networking/GraphQL/Common/GraphQLClient.swift
```diff
@@ -0,0 +1,64 @@
+//
+//  GraphQLClient.swift
+//  Card
+//
+//  Created by Andres Pelaez on 19/05/22.
+//
+
```
sshropshire:
```suggestion
```

Sources/PaymentsCore/Networking/GraphQL/Common/Models/GraphQLQueryResponse.swift
andrespelaezp: I don't think we'll be adding any more values, this is the current structure of the response for the GraphQL query response

Sources/PaymentsCore/Networking/GraphQL/Common/GraphQLClient.swift
andrespelaezp: good one! I like this, we'll give it a try

Sources/PaymentsCore/Networking/GraphQL/Common/Models/GraphQLQueryResponse.swift
mattwylder: Do we need the struct? Could we expect to consume and use something that conforms to `Codable` instead of unwrapping it from this strict? The generic conforming to `Codable` feels redundant to me.

sshropshire: Yeah all GraphQL responses have a top-level `data` property unfortunately.

jcnoriega (approved):

Sources/PaymentsCore/Networking/Enums/Environment.swift
```diff
@@ -13,5 +13,13 @@ public enum Environment {
             return URL(string: "https://api.paypal.com")!
         }
     }
-    // swiftlint:enable force_unwrapping
+
+    public var graphqlURL: URL {
```
sshropshire:
```suggestion
    public var graphQLURL: URL {
```
It's doesn't look the best, but it is consistent with the casing we have for `graphQL`.

Sources/PaymentsCore/api/EligibilityAPI.swift
```diff
@@ -0,0 +1,34 @@
...
+            return Result.failure(GraphQLError(message: "error fetching eligibility", extensions: nil))
+        } else {
+            return Result.success(
+                Eligibility(
+                    isVenmoEligible: response.data?.fundingEligibility.venmo.eligible ?? false,
+                    isPaypalEligible: response.data?.fundingEligibility.paypal.eligible ?? false,
+                    isPaypalCreditEligible: response.data?.fundingEligibility.credit.eligible ?? false,
+                    isPayLaterEligible: response.data?.fundingEligibility.paylater.eligible ?? false,
+                    isCreditCardEligible: response.data?.fundingEligibility.card.eligible ?? false)
+            )
```
sshropshire:
```suggestion
            let fundingEligibility = response.data?.fundingEligibility
            return Result.success(
                Eligibility(
                    isVenmoEligible: fundingEligibility?.venmo.eligible ?? false,
                    isPaypalEligible: fundingEligibility?.paypal.eligible ?? false,
                    isPaypalCreditEligible: fundingEligibility?.credit.eligible ?? false,
                    isPayLaterEligible: fundingEligibility?.paylater.eligible ?? false,
                    isCreditCardEligible: fundingEligibility?.card.eligible ?? false)
            )
```
This is a nit. We could extract a variable here to make the code more DRY.

UnitTests/PaymentsCoreTests/EligibilityParsing_Tests.swift
```diff
@@ -0,0 +1,86 @@
...
+@testable import PaymentsCore
+@testable import Card
+
+class EligibilityParsing_Tests: XCTestCase {
+
+    func testFundingEligibility_matchesResponse() throws {
+        if let jsonData = validFundingEligibilityResponse.data(using: .utf8) {
+            let fundingEligibilityObject = try JSONDecoder().decode(FundingEligibilityResponse.self, from: jsonData)
+            XCTAssertTrue(fundingEligibilityObject.fundingEligibility.venmo.eligible == true)
+            XCTAssertTrue(fundingEligibilityObject.fundingEligibility.card.eligible == false)
```
sshropshire: Can we add assertions for the other payment methods as well? `paypal`, `credit`, `paylater`

sshropshire (approved):
🌊

Sources/PaymentsCore/api/EligibilityAPI.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+public class EligibilityAPI {
```
scannillo, line 3: 👋 Hey @kgangineni - I am catching up on some of the PRs since I've been away.

Question for you - can you provide some background on what the purpose of this the EligilibityAPI is?

I see that it is public, which means it is meant for merchant use - when would a merchant use this?

kgangineni: hey..
this is for merchants to check if they have access to payment source type like paypal, venmo etc as check point before showing it to user

scannillo: Gotcha, thanks Karthik! Do we have a ticket for adding docstrings to this public class?

Also curious - was this feature a request from product/PMs?

kgangineni: good callout, need to decide on it, weahter it will be public API or we will be used by us while showing payment methods, either way will need a ticket for this

scannillo: 👋 I made a JIRA for this work `DTNOR-578`

PR #61, opened 2022-07-11
Remove Client secret

kgangineni:
### Reason for changes
End goal is to remove client secrets from merchant apps and just use Low Scope Access Token (LSAT) to communicate with the server
### Summary of changes
1. Remove client secrets from core config
2. fetch access token from sample server in demo app
- 

### Checklist

- [ ] Added a changelog entry

@kgangineni 
-


Sources/Card/GetOrderInfoRequest.swift
```diff
@@ -8,15 +8,14 @@ struct GetOrderInfoRequest: APIRequest {
...
-    let secret: String
     let pathFormat: String = "v2/checkout/orders/%@"
 
     var path: String
     var method: HTTPMethod = .get
     var body: Data?
 
     var headers: [HTTPHeader: String] {
-        let encodedClientID = "\(clientID):\(secret)".data(using: .utf8)?.base64EncodedString() ?? ""
+        let encodedClientID = "\(clientID)".data(using: .utf8)?.base64EncodedString() ?? ""
```
sshropshire: We may need to use an access token here. I'm not sure if Basic auth with `clientID` only will work for this endpoint.

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -5,25 +5,26 @@ public final class APIClient {
...
+        self.coreConfig = coreConfig
         self.urlSession = urlSession
     }
 
     public func fetch<T: APIRequest>(endpoint: T) async throws -> (T.ResponseType, CorrelationID?) {
-        guard let request = endpoint.toURLRequest(environment: environment) else {
+        guard var request = endpoint.toURLRequest(environment: coreConfig.environment) else {
             throw APIClientError.invalidURLRequestError
         }
+        request.setValue(coreConfig.accessToken, forHTTPHeaderField: AUTHORIZATION)
```
sshropshire:
```suggestion
        request.setValue("Bearer \(coreConfig.accessToken)", forHTTPHeaderField: AUTHORIZATION)
```

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
```diff
@@ -2,7 +2,7 @@
 
 public struct AccessTokenResponse: Codable {
     let scope: String
-    let accessToken: String
+    public let accessToken: String
```
sshropshire: q: can we remove public on this property?

Sources/Card/GetOrderInfoRequest.swift
sshropshire: We should be able to use `Bearer <access_token>` here now too instead of clientID.

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
kgangineni: it's not working when accessing, something missing here

Demo/Demo/DemoSettings/DemoSettings.swift
```diff
@@ -48,14 +47,5 @@ enum DemoSettings {
...
-    /// Added secret to get order from api. This shouldnt be needed
-    static var secret: String {
-        var nsDictionary: NSDictionary?
-        if let path = Bundle.main.path(forResource: "Credentials", ofType: "plist") {
-            nsDictionary = NSDictionary(contentsOfFile: path)
-            return String(describing: nsDictionary?.value(forKey: "Secret") ?? "")
-        }
-        return ""
-    }
+
```
jaxdesmarais:
```suggestion
```

jaxdesmarais: This is resulting in a linter warning. You can follow the steps in [DEVELOPMENT.md](https://github.com/paypal/iOS-SDK/blob/main/DEVELOPMENT.md) to install Swiftlint. This will allow the linter warnings to be shown in Xcode as a warning throughout the development process.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
     private func buildBaseURL(with endpoint: String) -> URL? {
         URL(string: DemoSettings.environment.baseURL + endpoint)
     }
+    
+    public func getAccessToken(clientId: String, environment: PaymentsCore.Environment) async -> String?{
```
```suggestion
    public func getAccessToken(clientId: String, environment: PaymentsCore.Environment) async -> String? {
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
...
+        guard let token = self.accessToken else{
```
```suggestion
        guard let token = self.accessToken else {
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
...
+            self.accessToken = await fetchAccessToken(clientId: clientId, environment: environment)
+            return self.accessToken
+        }
+        return token
+    }
+    
+    private func fetchAccessToken(clientId: String, environment: PaymentsCore.Environment) async -> String? {
```
jaxdesmarais: Is there a reason we can't have this logic in the `getAccessToken` function? It seems like all `getAccessToken` does is call into this function and sets `self.accessToken` to the access token. I'm not fully understanding why we need a variable for this and cannot just condense this down to 1 function returning either the access token or throwing an error.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
...
+                throw URLResponseError.networkConnectionError
+            }
+            switch response.statusCode {
+                case 200..<300:
+                    let accessTokenResponse: AccessTokenResponse = try parse(from: data)
+                    return accessTokenResponse.accessToken
+            default: throw URLResponseError.serverError
+            }
+        }
+        catch{
```
```suggestion
        catch {
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
...
+            print("Error in fetching token")
+            return nil
```
jaxdesmarais, line 88: I think we should throw an error here vs printing a message and returning `nil`

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +70,36 @@ final class DemoMerchantAPI {
...
+        }
+    }
+
```
```suggestion
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -23,23 +23,23 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+    var payPalWebCheckoutClient: PayPalWebCheckoutClient? = nil
 
     /// order ID shared across views
     @Published var orderID: String?
 
-    lazy var payPalClient: PayPalWebCheckoutClient = {
-        let clientID = DemoSettings.clientID
-        let environment = DemoSettings.environment.paypalSDKEnvironment
-        let config = CoreConfig(clientID: clientID, environment: environment, secret: DemoSettings.secret)
+    func getPayPalClient() async throws -> PayPalWebCheckoutClient {
```
jaxdesmarais: We should move this function down with the other functions in this file

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -120,12 +120,11 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
     }
 
-    func checkoutWithCard(_ card: Card, orderID: String, context: ASWebAuthenticationPresentationContextProviding) async {
-        let config = CoreConfig(
-            clientID: DemoSettings.clientID,
-            environment: DemoSettings.environment.paypalSDKEnvironment,
-            secret: DemoSettings.secret
-        )
+    func checkoutWithCard(_ card: Card, orderID: String,
+            context: ASWebAuthenticationPresentationContextProviding) async {
```
jaxdesmarais: I think we can keep this as 1 line unless we are getting a linter warning for line length. If we are this is the format we would follow:

```suggestion
    func checkoutWithCard(
        _ card: Card, 
        orderID: String,
        context: ASWebAuthenticationPresentationContextProviding
    ) async {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         paymentButtonTapped(context: context, funding: .paypal)
     }
 
-    private func paymentButtonTapped(context: ASWebAuthenticationPresentationContextProviding, funding: PayPalWebCheckoutFundingSource) {
+    private func paymentButtonTapped(
+        context: ASWebAuthenticationPresentationContextProviding,
+        funding: PayPalWebCheckoutFundingSource) {
```
```suggestion
        funding: PayPalWebCheckoutFundingSource
    ) {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
     }
 
     func checkoutWithPayPal(
         orderID: String,
         context: ASWebAuthenticationPresentationContextProviding,
         funding: PayPalWebCheckoutFundingSource
-    ) {
-        let payPalRequest = PayPalWebCheckoutRequest(orderID: orderID, fundingSource: funding)
-        payPalClient.start(request: payPalRequest, context: context)
+    )  {
```
```suggestion
    ) {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        Task.init{
```
```suggestion
        Task {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+            do{
```
```suggestion
            do {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+                payPalWebCheckoutClient = try await getPayPalClient()
+                guard let client = payPalWebCheckoutClient else {
+                    return
+                }
+                let payPalRequest = PayPalWebCheckoutRequest(orderID: orderID, fundingSource: funding)
+                client.start(request: payPalRequest, context: context)
+            }
+        catch{
```
```suggestion
        catch {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+            print("Error in starting paypal webcheckout client")
```
jaxdesmarais: We should throw an error in this catch vs printing a message

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +163,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        }
+    }
```
jaxdesmarais, line 190: Looks like the spacing may be off in this file a bit. Highlighting the code then `cmd` + `i` should indent things correctly in Xcode.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -235,4 +248,17 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         updateTitle("3DS challenge has finished")
         print("3DS challenge has finished")
     }
+    
+    func getAccessToken() async -> String? {
+        return await DemoMerchantAPI.sharedService.getAccessToken(
```
```suggestion
        await DemoMerchantAPI.sharedService.getAccessToken(
```

PayPal.xcodeproj/project.pbxproj
```diff
@@ -765,6 +765,29 @@
...
+			isa = PBXGroup;
+			children = (
+				CB4BE285284802C200EA2DD1 /* PaymentSource.swift */,
+				CB16E6E7285BBE6B00FD6F52 /* PurchaseUnit.swift */,
+				CB16E6E9285BBE7700FD6F52 /* Amount.swift */,
+				CB16E6EB285BBE8C00FD6F52 /* Link.swift */,
+			);
+			path = OrdersV2;
+			sourceTree = "<group>";
+		};
```
jaxdesmarais: I think we can revert this entire diff. Doesn't look like it's needed for this change and was likely caused by some merges being out of sync slightly.

Sources/Card/CardClient.swift
```diff
@@ -15,7 +15,7 @@ public class CardClient {
     /// - Parameter config: The CoreConfig object
     public init(config: CoreConfig) {
         self.config = config
-        self.apiClient = APIClient(environment: config.environment)
+        self.apiClient = APIClient(coreConfig: self.config)
```
jaxdesmarais: Can we use `config` directly here vs `self.config`?

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -40,4 +41,6 @@ public final class APIClient {
             throw APIClientError.serverResponseError(errorData.readableDescription)
         }
     }
+    
+    let AUTHORIZATION = "Authorization"
```
jaxdesmarais: Move to the top with rest of variables and constants in this file, though if this is only being used in 1 place we can likely just us the string value directly. We can always extract into a constant later if needed.

Sources/PaymentsCore/Networking/Models/Responses/AccessTokenResponse.swift
sshropshire: At a second glance we could move this entire model into the Demo application. That may solve access issues.

Demo/Demo/Networking/DemoMerchantAPI.swift
kgangineni: here `getAccessToken` will check variable if it exists, if not makes a call to fetch token, idea is not to repeat the call, get it save once and use it

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: haaa.. thanks was spending lot of time on fixing lint

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -134,14 +134,14 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         }
 
         Task {
-            await baseViewModel.checkoutWithCard(card, orderID: orderID, context: self)
+            await baseViewModel.checkoutWithCard(card: card, orderID: orderID, context: self)
```
jaxdesmarais: I think it's a bit more Swift-y to have this as:
```
baseViewModel.checkoutWithCard(card, orderID: orderID, context: self)
```
or:
```
baseViewModel.checkoutWith(card: card, orderID: orderID, context: self)
```

Demo/Demo/Models/AccessTokenRequest.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+import PaymentsCore
+
+/// Describes request to get a low scoped access token (LSAT) given some Client ID
+public struct AccessTokenRequest: APIRequest {
+
+    public init() {
+    }
```
jaxdesmarais:
```suggestion
    public init() { }
```

jaxdesmarais: This init should come after the properties if it's needed. If it's not needed since we default the variables lets remove.

Demo/Demo/Models/AccessTokenRequest.swift
```diff
@@ -0,0 +1,22 @@
...
+
+    public typealias ResponseType = AccessTokenResponse
+
+    public var path: String = "/access_tokens"
+    public var method: HTTPMethod = .post
+    public var body: Data?
+
+    public var headers: [HTTPHeader: String] {
+        return [
```
jaxdesmarais: return is not needed here, we can rely on Swift's [implicit returns](https://www.hackingwithswift.com/swift/5.1/implicit-returns)

Demo/Demo/Models/AccessTokenResponse.swift
```diff
@@ -0,0 +1,7 @@
+public struct AccessTokenResponse: Codable {
+
+    enum CodingKeys: String, CodingKey {
+        case accessToken = "access_token"
+    }
+    public let accessToken: String
```
jaxdesmarais:
```suggestion
    }
    
    public let accessToken: String
```

Demo/Demo/Networking/DemoMerchantAPI.swift
jaxdesmarais: Bumping this comment

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -67,4 +60,50 @@ final class DemoMerchantAPI {
...
+    private func createUrlRequest(accessTokenRequest: AccessTokenRequest, environment: Demo.Environment) -> URLRequest? {
+        var completeUrl = environment.baseURL
+        completeUrl.append(contentsOf: accessTokenRequest.path)
+        guard let url = URL(string: completeUrl) else {
+            return nil
```
jaxdesmarais: Should we return an error here if we cannot construct the URL?

Demo/Demo/ViewModels/BaseViewModel.swift
jaxdesmarais: Bumping this comment

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +166,33 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        funding: PayPalWebCheckoutFundingSource
+    ) {
         guard let orderID = orderID else {
             self.updateTitle("Failed: missing orderID.")
             return
         }
-
         checkoutWithPayPal(orderID: orderID, context: context, funding: funding)
     }
-
```
jaxdesmarais, line 175: Revert removing this space

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +166,33 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
     func checkoutWithPayPal(
         orderID: String,
         context: ASWebAuthenticationPresentationContextProviding,
         funding: PayPalWebCheckoutFundingSource
     ) {
-        let payPalRequest = PayPalWebCheckoutRequest(orderID: orderID, fundingSource: funding)
-        payPalClient.start(request: payPalRequest, context: context)
+        Task.init {
```
jaxdesmarais: I think we can just use `Task {`

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -235,4 +250,16 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         updateTitle("3DS challenge has finished")
         print("3DS challenge has finished")
     }
+
+    func getAccessToken() async -> String? {
+        await DemoMerchantAPI.sharedService.getAccessToken(
+            environment: DemoSettings.environment)
```
```suggestion
        await DemoMerchantAPI.sharedService.getAccessToken(environment: DemoSettings.environment)
```

PayPal.xcodeproj/project.pbxproj
jaxdesmarais: Bumping this comment

Sources/Card/CardClient.swift
jaxdesmarais: Bumping this comment

Sources/Card/CardClient.swift
```diff
@@ -47,12 +47,9 @@ public class CardClient {
...
         Task {
             do {
-                let confirmPaymentRequest = try ConfirmPaymentSourceRequest(
-                    cardRequest: request,
-                    clientID: config.clientID
-                )
+                let confirmPaymentRequest = try ConfirmPaymentSourceRequest(cardRequest: request)
                 let (result, _) = try await apiClient.fetch(endpoint: confirmPaymentRequest)
-                if let url = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                if let url: String = result.links?.first(where: { $0.rel == "payer-action" })?.href {
```
jaxdesmarais, line 52: I _think_ this can be simplified to:
```
result.links?.first { $0.rel == "payer-action" }?.href
```

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -19,21 +18,17 @@ struct ConfirmPaymentSourceRequest: APIRequest {
     private let jsonEncoder = JSONEncoder()
 
     var headers: [HTTPHeader: String] {
-        let encodedClientID = "\(clientID):".data(using: .utf8)?.base64EncodedString() ?? ""
 
         return [
-            .contentType: "application/json",
-            .acceptLanguage: "en_US",
-            .authorization: "Basic \(encodedClientID)"
+            .contentType: "application/json", .acceptLanguage: "en_US"
```
jaxdesmarais, line 23:
```suggestion
            .contentType: "application/json", 
            .acceptLanguage: "en_US"
```

Sources/PaymentsCore/Networking/Models/Requests/AccessTokenRequest.swift
```diff
@@ -3,24 +3,19 @@ import Foundation
 /// Describes request to get a low scoped access token (LSAT) given some Client ID
 public struct AccessTokenRequest: APIRequest {
 
-    public typealias ResponseType = AccessTokenResponse
+    public init() {
+    }
```
jaxdesmarais: Not sure if we need this init but if we do it can move under the properties.

UnitTests/CardTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -43,16 +42,13 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
         )
 
         let expectedPath = "/v2/checkout/orders/\(mockOrderId)/confirm-payment-source"
         let expectedMethod = HTTPMethod.post
-        let encodedCredentials = "\(mockClientId):".data(using: .utf8)?.base64EncodedString() ?? ""
         let expectedHeaders: [HTTPHeader: String] = [
-            .contentType: "application/json",
-            .acceptLanguage: "en_US",
-            .authorization: "Basic \(encodedCredentials)"
+            .contentType: "application/json", .acceptLanguage: "en_US"
```
jaxdesmarais, line 51:
```suggestion
            .contentType: "application/json", 
            .acceptLanguage: "en_US"
```

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
kgangineni: can we have mixed way
`baseViewModel.checkoutWithCard(card, orderID: orderID, context: self)` with some parameters having label and others not, I'm getting compilation error

Demo/Demo/Networking/DemoMerchantAPI.swift
kgangineni: which error to throw here? any idea

jaxdesmarais: Eh, thinking about it now this is the demo app so maybe not as important? Up to you. 😄

Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
jaxdesmarais: That should be fine. Was is also updated in the `baseViewModel.checkoutWithCard` function signature? Guessing the compiler error is from updating here but not the method in the `baseViewModel`

PayPal.xcodeproj/project.pbxproj
kgangineni: i think I fixed this one, let me know if more changes are needed

sshropshire (approved):
🌊

Demo/Demo/Networking/DemoMerchantAPI.swift
kaylagalway: Seems like we handle the error portion for this failing on line 77 - since these are private helper functions, I feel like it's probably fine either way? This function can only fail for one reason, which is the `URL` initialization failure, so I think returning `nil` in this case is kind of okay, as we aren't going to be confused between `nil` representing multiple error cases. 

We could change the error on line 77 to be a little clearer though? Right now it is: `URLResponseError.dataParsingError` - could change it to something like `URLResponseError.urlInitializationError` or something that indicates its related to the URL string having a problem?

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -164,22 +158,34 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
         context: ASWebAuthenticationPresentationContextProviding,
         funding: PayPalWebCheckoutFundingSource
     ) {
-        let payPalRequest = PayPalWebCheckoutRequest(orderID: orderID, fundingSource: funding)
-        payPalClient.start(request: payPalRequest, context: context)
+        Task {
+            do {
+                payPalWebCheckoutClient = try await getPayPalClient()
+                guard let client = payPalWebCheckoutClient else {
+                    return
```
kaylagalway, line 182: Should we be doing any error handling here? Could we add a print like we do in the `catch` below? 

`print("Error initializing paypal webcheckout client")`

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -200,9 +206,11 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
     }
 
     // TODO: this is just for test, need to be removed after final integration
-    func testEligibility() async {
-        let config = CoreConfig(clientID: DemoSettings.clientID, environment: DemoSettings.environment.paypalSDKEnvironment)
-            let eligibilityAPI = EligibilityAPI(coreConfig: config)
+    func testEligibility() async throws {
+        guard let config = await getCoreConfig() else {
+            return
```
kaylagalway, line 213: Same `print` question here - should we add `print("error making config for test eligibility")` ?

PayPal.xcodeproj/project.pbxproj
kaylagalway: If we still see this diff here - it's probably not fixed, unless you see the correct lines locally? Seems like this code should be on line 819. Below it looks like it was removed from those lines

Demo/Demo/Models/AccessTokenRequest.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
```
kaylagalway, line 1: Should these files only be in the Demo app now? Are they not required in the main framework? Sorry if this is a silly question - couldn't figure it out from the PR description

Demo/Demo/Models/AccessTokenResponse.swift
```diff
@@ -0,0 +1,8 @@
+public struct AccessTokenResponse: Codable {
```
kaylagalway, line 1: Same question for this file - just want to make sure it's only supposed to be in the demo app

kaylagalway (commented):
Made some comments on the demo app error printing code that are more "nits" and not required to change.

My only question is about the files that were moved from the main project to the demo app - trying to get a better understanding on why those were moved. Would it be possible to write a longer PR description for future folks who might need context into why we removed the client id and which files were moved around?

Demo/Demo/Models/AccessTokenRequest.swift
kgangineni: Yes, these are communicating with sample merchant server not paypal api
so merchants will have their own thing to communicate with their servers

Demo/Demo/Networking/DemoMerchantAPI.swift
kgangineni: changed error handling part, check again

Demo/Demo/Models/AccessTokenRequest.swift
kaylagalway: Would it be possible for you to explain that a bit more in the PR description? Why these changes are made/what merchants are fetching instead of us?

kgangineni: added description, let me know if it's not clear

kaylagalway (approved):
Thanks for the updates! lgtm!

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -1,33 +1,29 @@
...
+import PaymentsCore
 
 /// API Client used to create and process orders on sample merchant server
 final class DemoMerchantAPI {
 
     static let sharedService = DemoMerchantAPI()
-
+    var accessToken: String?
     private init() {}
-
```
jaxdesmarais, line 9: I think we still need this space?

jaxdesmarais: Seems like a lot of spacing in this file was deleted that we should revert so there is some seperation between items 😄

jaxdesmarais (approved):
Just a couple of very small cleanup comments but otherwise looks great! Thanks for addressing feedback.

Demo/Demo/Networking/DemoMerchantAPI.swift
kgangineni: spacings in in xcode is so confusing.. i think it's result of ctrl + i

Demo/Demo/Models/AccessTokenRequest.swift
kaylagalway: Works for me, thanks!

PR #65, opened 2022-07-21
Update gems so that jmespath version is greater than 1.6.1

kaylagalway:
### Reason for changes

Update gems so that dependabot security alert is resolved:
https://github.com/paypal/iOS-SDK/security/dependabot/1

jmespath.rb (aka JMESPath for Ruby) before 1.6.1 uses JSON.load in a situation where JSON.parse is preferable.

### Summary of changes

- Upgraded gems so that jmespath is increased to version 1.6.1 or later

### Checklist

- [x] Added a changelog entry

-


jaxdesmarais (approved):
😎

sshropshire (approved):

PR #67, opened 2022-07-28
Minimal check for native checkout

kgangineni:
### Reason for changes
Give a try for native checkout
### Summary of changes
1. Consume native checkout in demo
2. change client id which can use native checkout
3. update native checkout to latest version
- 

### Checklist

- [ ] Added a changelog entry

@kgangineni @jcnoriega 
-


Demo/Demo/FeatureViewControllers/CardDemoViewController.swift
```diff
@@ -61,6 +62,12 @@ class CardDemoViewController: FeatureBaseViewController, UITextFieldDelegate {
         return eligibilityButton
     }()
 
+    lazy var nativeCheckoutButton: CustomButton = {
+        let nativeCheckoutButton = CustomButton(title: "Native Checkout")
+        nativeCheckoutButton.addTarget(self, action: #selector(didTapNativeCheckoutButton), for: .touchUpInside)
+        return nativeCheckoutButton
+    }()
+
```
jaxdesmarais: Native Checkout should be in its own demo ViewController not in the card ViewController - this will allow merchants to see each module and the modules code independent of one another.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
```diff
@@ -12,7 +12,7 @@ protocol CheckoutProtocol {
     typealias CancelCallback = () -> Void
     typealias ErrorCallback = (PayPalCheckoutErrorInfo) -> Void
 
-    static func set(config: CoreConfig, returnURL: String)
+    static func set(config: CoreConfig)
```
jcnoriega, line 15: I dont see that the `returnUrl` is passed to `Checkout`, is it not needed on iOS?

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -126,6 +127,44 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         cardClient.approveOrder(request: cardRequest, context: context)
     }
 
+    func checkoutWithNativeClient(orderId: String?) async throws -> NativeCheckoutResult {
+        guard let orderId = orderId else {
+
+            return NativeCheckoutResult.error(CoreSDKError(code: 0, domain: "Order Id is null", errorDescription: "Order Id is null"))
```
jaxdesmarais, line 149: I don't think we'd want merchants to build their own errors using `CoreSDKError`. If we look at some of the other examples in this file we'd likely want to bubble up the errors from the service vs a merchants creating their own.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -126,6 +127,44 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        }
+        let nativeCheckoutClient = try await getNativeCheckoutClient()
+        let paypalRequest = PayPalRequest(orderID: orderId)
+
+        class PaypalCheckoutDelegate: PayPalDelegate {
```
jaxdesmarais: Class within a class feels a little weird. I think we can just make the main class conform to PayPalDelegate or make an extension of the main class that conforms to our delegate at the end of the file and add the protocol stubs there.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -126,6 +127,44 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+
+            var nativeCheckoutResult: NativeCheckoutResult?
+            func paypal(_ payPalClient: PayPalClient, didFinishWithResult result: PayPalResult) {
+                nativeCheckoutResult = NativeCheckoutResult.approved(ApprovalResult(
+                    orderId: result.orderID, payerId: result.payerID
+                ))
```
```suggestion
               let approvalResult = ApprovalResult(orderId: result.orderID, payerId: result.payerID)
                nativeCheckoutResult = NativeCheckoutResult.approved(approvalResult)
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -126,6 +127,44 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        }
+
+        let callbackDelegate = PaypalCheckoutDelegate()
+        nativeCheckoutClient.delegate = callbackDelegate
+        DispatchQueue.main.async {
+            nativeCheckoutClient.start(request: paypalRequest)
+        }
+        guard let nativeCheckoutResult = callbackDelegate.nativeCheckoutResult
+        else {
+            throw CoreSDKError(code: 0, domain: "checkout", errorDescription: "didn't get result from checkout")
```
jaxdesmarais, line 159: Same comment here on errors, we would want to bubble up the error from the SDK vs a merchant creating their own using our CoreSDKError.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -257,6 +296,13 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         return CoreConfig(clientID: DemoSettings.clientID, accessToken: token, environment: DemoSettings.environment.paypalSDKEnvironment)
     }
 
+    func getNativeCheckoutClient() async throws -> PayPalClient {
+        guard let config = await getCoreConfig() else {
+            throw CoreSDKError(code: 0, domain: "Error initializing paypal webcheckout client", errorDescription: nil)
```
jaxdesmarais, line 297: Same error comment here, we can instead bubble up an error when we attempt to fetch the config.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -257,6 +296,13 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
+        }
+        return PayPalClient(config: config, returnURL: "com.paypal.android://paypalpay")
```
jaxdesmarais: I don't think we need the return URL anymore on iOS.

PayPal.xcodeproj/project.pbxproj
```diff
@@ -740,6 +740,29 @@
...
+			isa = PBXGroup;
+			children = (
+				CB4BE285284802C200EA2DD1 /* PaymentSource.swift */,
+				CB16E6E7285BBE6B00FD6F52 /* PurchaseUnit.swift */,
+				CB16E6E9285BBE7700FD6F52 /* Amount.swift */,
+				CB16E6EB285BBE8C00FD6F52 /* Link.swift */,
+			);
+			path = OrdersV2;
+			sourceTree = "<group>";
+		};
```
jaxdesmarais: Looks like this diff got moved again. We should revert this.

PayPal.xcodeproj/project.pbxproj
```diff
@@ -3199,17 +3197,44 @@
...
+				CLANG_WARN_INT_CONVERSION = YES;
+				CLANG_WARN_NON_LITERAL_NULL_CONVERSION = YES;
+				CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF = YES;
+				CLANG_WARN_OBJC_LITERAL_CONVERSION = YES;
+				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES;
+				CLANG_WARN_RANGE_LOOP_ANALYSIS = YES;
+				CLANG_WARN_STRICT_PROTOTYPES = YES;
+				CLANG_WARN_SUSPICIOUS_MOVE = YES;
+				CLANG_WARN_UNREACHABLE_CODE = YES;
+				CLANG_WARN__DUPLICATE_METHOD_MATCH = YES;
```
jaxdesmarais: Looks like a lot of changes got added here, if these were all needed along with the other changes, would you mind adding some comments to why they were needed?

jaxdesmarais: Same in several other places above and below this.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
```diff
@@ -25,11 +25,9 @@ protocol CheckoutProtocol {
 
 extension Checkout: CheckoutProtocol {
 
-    static func set(config: CoreConfig, returnURL: String) {
+    static func set(config: CoreConfig) {
         let nxoConfig = CheckoutConfig(
-            clientID: config.clientID,
-            returnUrl: returnURL,
-            environment: config.environment.toNativeCheckoutSDKEnvironment()
```
jaxdesmarais, line 32: Does NXO not need the environment from us anymore?

UnitTests/PayPalTests/MockCheckout.swift
```diff
@@ -3,14 +3,15 @@ import UIKit
 
 @testable import PayPalNativeCheckout
 @testable import PaymentsCore
+import PayPalCheckout
```
jaxdesmarais: Do we need this import?

UnitTests/PayPalTests/MockCheckout.swift
```diff
@@ -48,4 +49,4 @@ struct MockPayPalError: PayPalCheckoutErrorInfo {
     var reason: String
     var error: Error
 }
-// swiftlint:enable space_after_main_type
+// swiftlint:disable space_after_main_type
```
jaxdesmarais: Do we need this change?

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -34,7 +34,9 @@ class PayPalClient_Tests: XCTestCase {
 
     func testStart_whenNativeSDKOnApproveCalled_returnsPayPalResult() async {
         let request = PayPalRequest(orderID: "1234")
-        let approval = MockApproval(intent: "intent", payerID: "payerID", ecToken: request.orderID)
+        // swiftlint: disable force_unwrapping
+        let approval = MockApproval(intent: ApprovalOrderIntent(rawValue: 0)!, payerID: "payerID", ecToken: request.orderID)
+        // swiftlint: enable force_unwrapping
```
jaxdesmarais: Do we need this change or can we use a string like this test used previously?

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
jaxdesmarais: Return URL is no longer needed on iOS - I don't think it's needed on Android either (at least was not for the Braintree integration)

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: this would be refactored and moved to lib, this is for trying placed all code in demo

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: removed

PayPal.xcodeproj/project.pbxproj
kgangineni: not sure if these changes are needed, I just updated lib and these changes got adde

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
kgangineni: it's working without return url

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
kgangineni: fixed, it does need environment

UnitTests/PayPalTests/MockCheckout.swift
kgangineni: yeah `ApprovalOrderIntent` is from this

UnitTests/PayPalTests/MockCheckout.swift
kgangineni: yeah both lines are disable, enabling at end

UnitTests/PayPalTests/PayPalClient_Tests.swift
kgangineni: it's compile error, if it's string

Demo/Demo/ViewModels/BaseViewModel.swift
jaxdesmarais: Is there a follow up ticket for that work? If not, we should just print an error here as using `CoreSDKError` directly is not a behavior we want merchants to follow.

PayPal.xcodeproj/project.pbxproj
jaxdesmarais: 🤔 updating the nxo module should not cause these settings to be updated, so we should try to revert these.

UnitTests/PayPalTests/MockCheckout.swift
jaxdesmarais: For NXO classes the approach we have been taking is to wrap the types vs pulling them in directly. Is there a specific reason we needed to move to use `ApprovalOrderIntent` vs using the String as we were previously?

UnitTests/PayPalTests/MockCheckout.swift
jaxdesmarais: Now we are disabling for any future structs/classes in this file instead of re-enabling the rule after these 2 struct's. I think we want to revert this to re-enable the rule again after these 2 struct's as expected.

Demo/Demo/DemoSettings/DemoType.swift
```diff
@@ -27,6 +30,8 @@ enum DemoType: String {
             return AnyView(SwiftUIPayPalDemo())
         case .paymentButtonCustomization:
             return AnyView(SwiftUIPaymentButtonDemo())
+        case .payPalNativeCheckout:
+            return AnyView(SiftUINativeCheckoutDemo())
```
mattwylder: small typo here.

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: yep will be creating a follow-up ticket for this

jcnoriega (approved):

andrespelaezp (approved):

Demo/Demo/FeatureViewControllers/NativeCheckoutDemoViewController.swift
```diff
@@ -0,0 +1,50 @@
...
+                stackView.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: Constants.layoutSpacing),
+                stackView.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: Constants.layoutSpacing)
+            ]
+        )
+    }
+
+    @objc func didTapNativeCheckoutButton() {
+        Task {
+            nativeCheckoutButton.startAnimating()
+            guard let orderId = baseViewModel.orderID else {
```
jaxdesmarais, line 43: nit: `orderId` > `orderID`

Demo/Demo/Models/ApprovalResult.swift
```diff
@@ -0,0 +1,10 @@
+struct ApprovalResult {
+
+    let orderId: String
```
jaxdesmarais, line 3:
```suggestion
    let orderID: String
```

jaxdesmarais (approved):
Couple of very small comments mainly moving from `Id` to `ID` as thats our naming convention on iOS. Otherwise, this looks good - thanks for addressing comments!

Demo/Demo/ViewModels/BaseViewModel.swift
scannillo: Hey @kgangineni - can you post the ticket number here so we make sure this doesn't get forgotten? I agree with @jaxdesmarais that we want our Demo to mirror how a merchant will integrate with (and interact with) our SDK. 

We don't want them constructing NativeCheckoutResult error types. Actually come to think of it, they shouldn't even be able to. The constructors for internal error and result types should be inaccessible to merchants. <-- This is also worth following up on. This was doable in ObjC, we should look into if it is for Swift.

kgangineni: @scannillo changes ar in this [PR](https://github.com/paypal/iOS-SDK/pull/69)

PR #69, opened 2022-08-09
NXO Integration

kgangineni:
### Reason for changes
1. Provide interface for merchants to access nxo features

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

@kgangineni @jcnoriega 
-


Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
```diff
@@ -0,0 +1,26 @@
...
+
+
+    func start(presentingViewController: UIViewController?,
+                    createOrder: CheckoutConfig.CreateOrderCallback?,
+                    onApprove: CheckoutConfig.ApprovalCallback?,
+                    onShippingChange: CheckoutConfig.ShippingChangeCallback?,
+                    onCancel: CheckoutConfig.CancelCallback?,
+                    onError: CheckoutConfig.ErrorCallback?) {
+        Checkout.set(config: nxoConfig)
+        Checkout.start(presentingViewController: presentingViewController, createOrder: createOrder, onApprove: onApprove, onShippingChange: onShippingChange, onCancel: onCancel, onError: onError)
```
sshropshire: Fundamentally this is a good start. There is a small difference [on Android](https://github.com/paypal/Android-SDK/blob/c622da9382bac770b40ed396b48afc06dd30d710/PayPalNativeCheckout/src/main/java/com/paypal/android/checkout/PayPalClient.kt#L67) where we're using a listener to encapsulate the native checkout callback setup. The iOS equivalent here would be a delegate, similar to the one used in the [PayPalWebCheckout](https://github.com/paypal/iOS-SDK/blob/main/Sources/PayPalWebCheckout/PayPalWebCheckoutDelegate.swift) module.

jcnoriega: this is an internal interface for testing, what you are refering to is on `PayPalClient`

Sources/PayPalNativeCheckout/PayPalDelegate.swift
```diff
@@ -1,25 +1,24 @@
...
-    func paypal(_ payPalClient: PayPalClient, didFinishWithError error: CoreSDKError)
+    func paypal(didFinishWithError error: CoreSDKError)
 
     /// Notify that the PayPal flow has been cancelled
     /// - Parameters:
     ///   - client: the PayPalClient associated with delegate
-    func paypalDidCancel(_ payPalClient: PayPalClient)
+    func paypalDidCancel()
+
+    func paypalDidShippingAddressChange(shippingChange: ShippingChange, shippingChangeAction: ShippingChangeAction)
```
jcnoriega: in all of this methods i would add back the `payPalClient`. In the case of `shippingChange` the merchant will need to perform an action on the `payPalClient`, so we can return that as a parameter in every method on `PayPalDelegate`, so that the merchant doesnt have to mantian a reference to the `PayPalClient`
It is also a good practice for iOS that the first argument of a delegate method should be the "object delegated" (in this case, `PayPalClient`) 
Reference: https://www.swiftbysundell.com/articles/delegation-in-swift/

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
kaylagalway: Do we need to use a delegate @sshropshire  ? Is there anything wrong with passing things through closures on both platforms? I'm a bit in the dark about kotlin/java norms so just let me know if that's a silly question. Conforming to a delegate feels like an unnecessary extra step for the merchant to me, but I've also missed previous context convos on this structure.

Sources/PayPalNativeCheckout/PayPalDelegate.swift
kgangineni: @jcnoriega for second part, you mean send `ShippingAction` on every callback?

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
sshropshire: @jcnoriega ah I see what you mean now I had to double back.

@kaylagalway I will say the path of least resistance for this PR is to keep delegates and create a separate Ticket to refine the developer experience. It'll be good for us to come up with a strategy we can apply to all current (and future) modules that makes sense for the SDK as a whole.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -143,22 +155,13 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate,
         cardClient.approveOrder(request: cardRequest, context: context)
     }
 
-    func checkoutWithNativeClient(orderId: String?) async throws -> NativeCheckoutResult {
+    func checkoutWithNativeClient(orderId: String?) async throws {
         guard let orderId = orderId else {
-
-            return NativeCheckoutResult.error(CoreSDKError(code: 0, domain: "Order Id is null", errorDescription: "Order Id is null"))
+            return
```
kaylagalway, line 160: Should this throw an error?

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
     /// Initialize a PayPalClient to process PayPal transaction
     /// - Parameters:
     ///   - config: The CoreConfig object
     ///   - returnURL: The return URL provided to the PayPal Native UI experience. Used as part of the authentication process to identify your application. This value should match the one set in the `Return URLs` section of your application's dashboard on your [PayPal developer account](https://developer.paypal.com)
     public init(config: CoreConfig) {
         self.config = config
-        self.CheckoutFlow = Checkout.self
+        let nxoConfig = CheckoutConfig(
+            clientID: config.clientID, createOrder: nil, onApprove: nil,
+                    onShippingChange: nil, onCancel: nil, onError: nil, environment: config.environment.toNativeCheckoutSDKEnvironment())
```
kaylagalway: Newlines seem off here, I think every parameter should be on a newline since the first one is. Like this:

```
  clientID: config.clientID, 
  createOrder: nil, 
  onApprove: nil,
  onShippingChange: nil, 
  onCancel: nil, 
  onError: nil, 
  environment: config.environment.toNativeCheckoutSDKEnvironment()
)
```

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
-                self.notifySuccess(for: approval)
-            },
-            onCancel: {
-                self.notifyCancellation()
-            },
-            onError: { errorInfo in
-                self.notifyFailure(with: errorInfo)
-            }
-        )
+    public func start(presentingViewController: UIViewController?, orderID: String, deleagate: PayPalDelegate?){
```
kaylagalway: typo in the `delegate` parameter here. Currently says `deleagate`. Need one additional space after the end between the `)` and '{' as well.

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
 
     let config: CoreConfig
 
-    // swiftlint:disable identifier_name
-    let CheckoutFlow: CheckoutProtocol.Type
-    // swiftlint:enable identifier_name
+    let nativeCheckout: CheckoutProtocol
```
kaylagalway: Could we add documentation for this variable?

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
kaylagalway: @sshropshire totally agree, that comment was pre-standup convo :)

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
+                }
+            )
+        }
     }
 
-    private func notifySuccess(for approval: PayPalCheckoutApprovalData) {
-        let payPalResult = PayPalResult(orderID: approval.ecToken, payerID: approval.payerID)
-        delegate?.paypal(self, didFinishWithResult: payPalResult)
+    private func notifySuccess(for approval: PayPalCheckout.Approval) {
+        delegate?.paypal(didFinishWithResult: approval)
```
kaylagalway: Do we need these helper functions for one single line of code? Might be easier to just call `delegate?.paypal(didFinishWithResult: approval)` directly instead of using this helper function? Unless we think we will be utilizing these a lot.

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
     }
 
     private func notifyFailure(with errorInfo: PayPalCheckoutErrorInfo) {
         let error = PayPalError.nativeCheckoutSDKError(errorInfo)
-        delegate?.paypal(self, didFinishWithError: error)
+        delegate?.paypal(didFinishWithError: error)
     }
```
kaylagalway: Same [question](https://github.com/paypal/iOS-SDK/pull/69/files#r941815212) for this helper function

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
 
     private func notifyCancellation() {
-        delegate?.paypalDidCancel(self)
+        delegate?.paypalDidCancel()
+    }
```
kaylagalway, line 85: Same [question](https://github.com/paypal/iOS-SDK/pull/69/files#r941815212) for this helper function

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -13,60 +9,67 @@ public class PayPalClient {
...
+    private func notifyShippingChange(shippingChange: ShippingChange, shippingChangeAction: ShippingChangeAction) {
+        delegate?.paypalDidShippingAddressChange(shippingChange: shippingChange, shippingChangeAction: shippingChangeAction)
```
kaylagalway: Same [question](https://github.com/paypal/iOS-SDK/pull/69/files#r941815212) for this helper function

kaylagalway (commented):
Left some small syntax change requests and just have one other question about the private helper functions in PayPalClient!

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,26 +4,38 @@ import PayPalWebCheckout
 import PaymentsCore
 import AuthenticationServices
 import PayPalNativeCheckout
+import PayPalCheckout
 
 /// This class is used to share the orderID across shared views, update the text of `bottomStatusLabel` in our `FeatureBaseViewController`
 /// as well as share the logic of `processOrder` across our duplicate (SwiftUI and UIKit) card views.
 class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate, PayPalDelegate {
+    func paypalDidShippingAddressChange(shippingChange: ShippingChange, shippingChangeAction: ShippingChangeAction) {
```
jaxdesmarais: we should move all of these after our functions for this file and add a mark like:
```
// MARK: - PayPalDelegate conformance
```

This file is a bit all over so having more organization to it will be helpful for merchants. Alternatively, we can move each of the delegates to a new file and do an extension on this class with the delegate conformance to clean this class up a bit more.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,26 +4,38 @@ import PayPalWebCheckout
...
+        updateTitle("an error occurred: \(error.errorDescription)")
     }
 
-    func paypalDidCancel(_ payPalClient: PayPalClient) {
-        nativeCheckoutResult = NativeCheckoutResult.cancel
+    func paypalDidCancel() {
+        updateTitle("order is cancelled")
     }
 
+//    func paypal(didFinishWithResult approvalResult: Approval) {
```
jaxdesmarais: Can remove this commented out code

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,26 +4,38 @@ import PayPalWebCheckout
...
+        updateTitle("shipping address changed")
+    }
 
-    var nativeCheckoutResult: NativeCheckoutResult?
-    func paypal(_ payPalClient: PayPalClient, didFinishWithResult result: PayPalResult) {
-        nativeCheckoutResult = NativeCheckoutResult.approved(ApprovalResult(
-            orderId: result.orderID, payerId: result.payerID
-        ))
+    func paypal(didFinishWithResult approvalResult: Approval) {
+        updateTitle("order placed with orderId: \(approvalResult.data.intent.stringValue)")
```
jaxdesmarais: Should we be passing the order ID in here since that's what the comment says? Right now we are passing the intent which doesn't line up with the first half of the comment.

Demo/Demo/ViewModels/BaseViewModel.swift
jaxdesmarais: +1 illustrating how a merchant would use CoreSDKError to construct an error (which we don't actually want merchants to do) is a bad practice for our demo app, we should either print an error message to the console or use the updateTitle method to illustrate what went wrong.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -143,22 +155,13 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate,
...
         nativeCheckoutClient.delegate = self
-        DispatchQueue.main.async {
-            nativeCheckoutClient.start(request: paypalRequest)
-        }
-        guard let nativeCheckoutResult = nativeCheckoutResult
-        else {
-            throw CoreSDKError(code: 0, domain: "checkout", errorDescription: "didn't get result from checkout")
-        }
-        return nativeCheckoutResult
+        nativeCheckoutClient.start(presentingViewController: nil, orderID: orderId, deleagate: self)
```
jaxdesmarais: We can default `presentingViewController` to `nil` in the start function so merchants can omit `presentingViewController` entirely if they don't need it. The this can be updated to:
```
nativeCheckoutClient.start(orderID: orderID, deleagate: self)
```

Sources/PayPalNativeCheckout/PayPalClient.swift
jaxdesmarais: +1

Sources/PayPalNativeCheckout/PayPalUIInterfaces/CheckoutProtocol.swift
```diff
@@ -1,52 +1,16 @@
...
-            onCancel: onCancel,
-            onError: onError
-        )
-    }
+    func start(presentingViewController: UIViewController?,
+                createOrder: CheckoutConfig.CreateOrderCallback?,
+                onApprove: CheckoutConfig.ApprovalCallback?,
+                onShippingChange: CheckoutConfig.ShippingChangeCallback?,
+                onCancel: CheckoutConfig.CancelCallback?,
+                onError: CheckoutConfig.ErrorCallback?)
```
jaxdesmarais:
```suggestion
    func start(
        presentingViewController: UIViewController?,
        createOrder: CheckoutConfig.CreateOrderCallback?,
        onApprove: CheckoutConfig.ApprovalCallback?,
        onShippingChange: CheckoutConfig.ShippingChangeCallback?,
        onCancel: CheckoutConfig.CancelCallback?,
        onError: CheckoutConfig.ErrorCallback?
)
```

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
```diff
@@ -0,0 +1,26 @@
...
+        self.nxoConfig = nxoConfig
+    }
+
+
+    func start(presentingViewController: UIViewController?,
+                    createOrder: CheckoutConfig.CreateOrderCallback?,
+                    onApprove: CheckoutConfig.ApprovalCallback?,
+                    onShippingChange: CheckoutConfig.ShippingChangeCallback?,
+                    onCancel: CheckoutConfig.CancelCallback?,
+                    onError: CheckoutConfig.ErrorCallback?) {
```
jaxdesmarais: Same comment as above on spacing here

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
```diff
@@ -0,0 +1,26 @@
+
+import Foundation
+import UIKit
+@_implementationOnly import PayPalCheckout
+import PaymentsCore
+
+class NativeCheckout: CheckoutProtocol {
```
jaxdesmarais: Wondering if there is a better name for this class since it's just used for testing. Also if that's the case do we need it in our main interface or can we make this a mock class in our tests if that's it's main purpose? We should also add some docs on what this class is/why we need it above to make this clear.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
```diff
@@ -0,0 +1,26 @@
...
+                    createOrder: CheckoutConfig.CreateOrderCallback?,
+                    onApprove: CheckoutConfig.ApprovalCallback?,
+                    onShippingChange: CheckoutConfig.ShippingChangeCallback?,
+                    onCancel: CheckoutConfig.CancelCallback?,
+                    onError: CheckoutConfig.ErrorCallback?) {
+        Checkout.set(config: nxoConfig)
+        Checkout.start(presentingViewController: presentingViewController, createOrder: createOrder, onApprove: onApprove, onShippingChange: onShippingChange, onCancel: onCancel, onError: onError)
+    }
+
+
```
jaxdesmarais: We can remove these extra newlines

Demo/Demo/ViewModels/BaseViewModel.swift
kaylagalway: That makes sense! Handling the error in some way, and the way a merchant would, seems like the best route

Demo/Demo/ViewModels/BaseViewModel.swift
kaylagalway: +1

Demo/Demo/ViewModels/BaseViewModel.swift
kaylagalway: I agree that separating the delegate conformance to an extension would make things a bit clearer to merchants. Not sure if separating into new files would make it as clear to them though. I feel like extensions in the same file would be enough since it's only a 300 line file? That way merchants don't have to go hunting around.

jaxdesmarais: True true - this class has kinda been a catch all, so I am fine with whatever allows us to clean this file up a bit for merchants to see how to do things per feature vs having to sift through this file and pick and choose what they need.

Sources/PayPalNativeCheckout/PayPalClient.swift
kgangineni: yeah just in case we move to different types helps to organize

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
kgangineni: this is uded in `PaypalClient` it's not just for testing

Sources/PayPalNativeCheckout/PayPalClient.swift
sshropshire: This is a good call out in general. I introduced these helper methods at some point. I was anxious about the strong implicit `self` retains in general. Like if NXO executes a callback but the View Controller is no longer on the screen? I was moving in the direction of `[weak self] in ... self?.notifySuccess()`.

Demo/Demo/ViewModels/BaseViewModel.swift
kaylagalway: @jaxdesmarais reflecting more on this, I personally don't think we should move them if we think this file will stay under 500 lines, as that's the general ideal max length of a file on most xcode linters. Seems like a win to make it so that merchants can easily see working examples in one place as long as they are organized. But, if we are going to be continuously conforming to more delegates with this file, then I think you're right and it might be best to just start breaking them out now.

jaxdesmarais: That makes sense to me! For now, I think organizing this file with `MARK`s for each feature would go a long way.

fastlane/Fastfile
```diff
@@ -37,6 +37,7 @@ platform :ios do
       configuration: "Debug",
       json_report: true,
       ignore_file_path: "fastlane/.xcovignore",
+      exclude_targets: "PayPalNativeCheckout"
```
kgangineni: any idea how to ignore `PaypalClient.swift` from code coverage, decided to do it in different story because there's lot of mocking needed to cover this code
@kaylagalway @jaxdesmarais ^^

jaxdesmarais: We can remove this line and add the file path(s) here: https://github.com/paypal/iOS-SDK/blob/main/fastlane/.xcovignore

Sources/PayPalNativeCheckout/PayPalDelegate.swift
```diff
@@ -2,24 +2,29 @@ import Foundation
...
-    ///   - client: the PayPalClient associated with delegate
     ///   - didFinishWithError: the error returned by the PayPal flow
     func paypal(_ payPalClient: PayPalClient, didFinishWithError error: CoreSDKError)
 
     /// Notify that the PayPal flow has been cancelled
     /// - Parameters:
     ///   - client: the PayPalClient associated with delegate
     func paypalDidCancel(_ payPalClient: PayPalClient)
+
+    func paypalDidShippingAddressChange(
```
scannillo, line 25: This is a public func and needs a merchant-friendly docstring

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
scannillo: Using a protocol here (`CheckoutProtocol`) is a great idea for us to be able to test `NativeCheckout.start()`.

Though, I agree with @jaxdesmarais that the naming is confusing, especially since `NativeCheckout` is a pretty loaded term/ has a lot of use throughout the SDK already. I think we can come up with better names for this class, along with the `CheckoutProtocol`.

---
Here are the [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/) on how to name protocols:

```
- Protocols that describe what something is should read as nouns (e.g. Collection).
- Protocols that describe a capability should be named using the suffixes `able`, `ible`, or `ing` (e.g. Equatable, ProgressReporting).
```

UnitTests/PayPalTests/MockCheckout.swift
```diff
@@ -3,49 +3,34 @@ import UIKit
...
+    required init(nxoConfig: CheckoutConfig) {
+    }
 
-    static func set(config: CoreConfig, clientID: String) { }
+    var onCancel: CheckoutConfig.CancelCallback?
+    var onError: CheckoutConfig.ErrorCallback?
 
-    static func start(
+    // todo: implemenet cases for other callbacks
+    // swiftlint:disable function_parameter_count
```
scannillo: Instead of silencing the linter for these violations, we should update the linter to suit our needs. It [warns at 5 params by default](https://realm.github.io/SwiftLint/function_parameter_count.html) which is low, especially for an SDK whose main job is to wrap lengthy/complex API calls.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
scannillo: Ideas? There already is a `PayPalClient.start()` so I wonder if we can think of something to remind us this class just basically wraps the 3rd party NXO framework 🤔 ...

`NativeCheckoutProvider: NativeCheckoutStartable`
`NativeCheckoutProvider: NativeCheckoutable`

Thoughts?

fastlane/.xcovignore
```diff
@@ -11,3 +11,6 @@
 - Sources/PaymentsCore/Networking/URLSession+URLSessionProtocol.swift
 # Exclude WebAuthSession for now
 - Sources/PaymentsCore/WebAuthenticationSession.swift
+# Excluding PayPal native checkout until figured out mocking interfaces with oxo
```
scannillo: What is this file for?

scannillo: Ah, I see [this comment](https://github.com/paypal/iOS-SDK/pull/69/files#r945901294)

fastlane/Fastfile
scannillo: Karthik can you explain more what you mean by removing this file from code coverage? Do you mean from linting?

Demo/Demo/ViewModels/BaseViewModel.swift
scannillo: I agree that this file has a lot of responsibility, but I think it's fine for now, as it isn't too long and our demo app is pretty simple. 

Organizing it by pragma mark (`// MARK: - <PROTOCOL_NAME>`) or class extensions within the file both sound like good options to me!

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
sshropshire: I'm 1️⃣  for `NativeCheckoutStartable` it does fit the naming guidelines well.

kgangineni: `NativeCheckoutProvider: NativeCheckoutStartable` this sounds good or `NativeCheckoutProvider: NativeCheckoutProtocol`

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: this class does have lot of responsibility will go with `//Mark:` separation now, but we can limited in base and create viewModels for different views later

fastlane/Fastfile
kgangineni: @scannillo  oh this shouldn't be there, I will take it off, this was to exclude few files since we couldn't figure out how to mock them but that's done in below `xcovignore`

UnitTests/PayPalTests/MockCheckout.swift
kgangineni: I agree 5 is too low, may be 7? ignored this because it's in tests not part of sdk

scannillo: We could exclude this rule entirely too, adding it [here](https://github.com/paypal/iOS-SDK/blob/295dbaaa0f0e98bb0bd60d2e721bf15262d8afbf/.swiftlint.yml#L11-L18).

Due to the nature of us being an SDK that does a lot of wrapping API requests and results, I don't think the limit on function param count really helps us that much. What do you think?

fastlane/Fastfile
scannillo: Ok - I think I don't understand why the file was adding to `xcovignore`?

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,27 +4,38 @@ import PayPalWebCheckout
...
 /// This class is used to share the orderID across shared views, update the text of `bottomStatusLabel` in our `FeatureBaseViewController`
 /// as well as share the logic of `processOrder` across our duplicate (SwiftUI and UIKit) card views.
 class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate, PayPalDelegate {
 
-    var nativeCheckoutResult: NativeCheckoutResult?
-    func paypal(_ payPalClient: PayPalClient, didFinishWithResult result: PayPalResult) {
-        nativeCheckoutResult = NativeCheckoutResult.approved(ApprovalResult(
-            orderId: result.orderID, payerId: result.payerID
-        ))
+    // MARK: - PayPalDelegate conformance
```
jaxdesmarais, line 13: Can we move these to the end of the file? Typically a class would go properties > initializers > public methods > internal methods > protocol conformance.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,27 +4,38 @@ import PayPalWebCheckout
...
+        guard let orderId = orderID else {
+            updateTitle("native checkout result: \(approvalResult.data.intent.stringValue)")
+            return
+        }
+        updateTitle("order \(orderId): \(approvalResult.data.intent.stringValue)")
     }
 
     func paypal(_ payPalClient: PayPalClient, didFinishWithError error: CoreSDKError) {
-        nativeCheckoutResult = NativeCheckoutResult.error(error)
+        updateTitle("an error occurred: \(String(describing: error.errorDescription))")
```
```suggestion
        updateTitle("an error occurred: \(error.localizedDescription)")
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,27 +4,38 @@ import PayPalWebCheckout
...
     }
 
     func paypalDidCancel(_ payPalClient: PayPalClient) {
-        nativeCheckoutResult = NativeCheckoutResult.cancel
+        updateTitle("order is cancelled")
```
```suggestion
        updateTitle("order is canceled")
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -143,21 +154,14 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate,
         cardClient.approveOrder(request: cardRequest, context: context)
     }
 
-    func checkoutWithNativeClient(orderId: String?) async throws -> NativeCheckoutResult {
-        guard let orderId = orderId else {
-            return NativeCheckoutResult.error(CoreSDKError(code: 0, domain: "Order Id is null", errorDescription: "Order Id is null"))
+    func checkoutWithNativeClient() async throws {
+        guard let orderId = self.orderID else {
```
```suggestion
        guard let orderID = self.orderID else {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -143,21 +154,14 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate,
...
         nativeCheckoutClient.delegate = self
-        Task {
-            await nativeCheckoutClient.start(request: paypalRequest)
-        }
-        guard let nativeCheckoutResult = nativeCheckoutResult
-        else {
-            throw CoreSDKError(code: 0, domain: "checkout", errorDescription: "didn't get result from checkout")
-        }
-        return nativeCheckoutResult
+        await nativeCheckoutClient.start(orderID: orderId, delegate: self)
```
```suggestion
        await nativeCheckoutClient.start(orderID: orderID, delegate: self)
```

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -10,77 +8,71 @@ import PaymentsCore
...
+                onApprove: { approval in
+                    self.notifySuccess(for: approval)
+                },
+                onShippingChange: { shippingChange, shippingChangeAction in
+                    self.notifyShippingChange(shippingChange: shippingChange, shippingChangeAction: shippingChangeAction)
+                },
+                onCancel: {
+                    self.notifyCancellation()
+                },
+                onError: {error in
```
jaxdesmarais:
```suggestion
                onError: { error in
```

UnitTests/PayPalTests/MockCheckout.swift
jaxdesmarais: Agree with Sammy, we should probably just disable this rule since we are likely to run into this often with building SDKs

fastlane/.xcovignore
jaxdesmarais: Do we have a follow up ticket to add this back in and test it? Feel a bit odd to leave PayPalClient untested. Also is PayPalClient still the best name for this or do we need to be more specific to the feature?

Demo/Demo/Models/NativeCheckoutResult.swift
```diff
@@ -1,8 +0,0 @@
-import Foundation
-import PaymentsCore
-
-enum NativeCheckoutResult {
```
scannillo, line 4: If we're removing this result type, what type are we returning to the merchant?

fastlane/.xcovignore
kgangineni: this is also just following android, I just paused development for this PR, because won't be able to complete until I get descision on callbacks, let me know if any suggestions for the name

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: I think that's correct spelling `cancelled` googled it lol!!

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: will make this change for now, but any reason why we chose `orderID` over `orderId`? because latter is the standard generally anywhere

Demo/Demo/ViewModels/BaseViewModel.swift
sshropshire: So.. this was hotly debated on the BT SDK. It turns out `ll` is [british](https://www.merriam-webster.com/words-at-play/canceled-or-cancelled)

Demo/Demo/Models/NativeCheckoutResult.swift
kgangineni: this is done with callbacks from nxo previously we were encapsulating result

kgangineni:
@jaxdesmarais yep we still have other tickets for nxo integration and even this ticket is carried over not closed

Demo/Demo/ViewModels/BaseViewModel.swift
kgangineni: oh ok, will commit this change

Demo/Demo/ViewModels/BaseViewModel.swift
sshropshire: On iOS in Objective-C `id` is technically a [reserved word](https://stackoverflow.com/a/9505640), so in some cases it can legit break the compiler.

jaxdesmarais: ^ this, but additionally Apple and most Swift style guides have moved to using `ID` or `Identifier` over `Id`. [NSManagedObjectID](https://developer.apple.com/documentation/coredata/nsmanagedobjectid) is a good example of this. When abbreviating and not using identifier Apple uses the uppercase "ID", for example NSManagedObject has an objectID, CKNotification has a notificationID and subscriptionID, CKRecord has a createrUserRecordID etc.

The [Swift Style Guide](https://www.swift.org/documentation/api-design-guidelines/) also indicates that:

> Acronyms and initialisms that commonly appear as all upper case in American English should be uniformly up- or down-cased according to case conventions

Some similar examples are `UTF8`, `HTTP`, etc. I think when we can following Apple's style for these objects ensures our SDK also feels Swift-y.

fastlane/.xcovignore
scannillo: > won't be able to complete until I get descision on callbacks

Can you elaborate on what this means and why these files were added to `xcovignore`?

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -10,77 +8,71 @@ import PaymentsCore
 public class PayPalClient {
 
     public weak var delegate: PayPalDelegate?
-
+    /// instance to communicate with native checkout
```
scannillo: Nitpick - we can remove this comment, in favor of a docstring at the top of the `CheckoutProtocol`, or whatever name we land on.

fastlane/.xcovignore
scannillo: I chatted with @kgangineni in person -turns out we have a code coverage minimum percentage rule in effect [here](https://github.com/paypal/iOS-SDK/blob/295dbaaa0f0e98bb0bd60d2e721bf15262d8afbf/fastlane/Fastfile#L24). The reason this PR ignores these files, is because several NXO classes are final and therefore untestable. Karthik has communicated these with the NXO team and they are working on updates.

fastlane/Fastfile
scannillo: See https://github.com/paypal/iOS-SDK/pull/69/files#r953970884

scannillo: @kgangineni - we still need to remove this line

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
scannillo: @kgangineni - this still needs to be addressed in the code

scannillo: @kgangineni - Responding to your question in Slack DM here: the `NativeCheckoutProvider: NativeCheckoutStartable` is preferred since the later (`NativeCheckoutProvider: NativeCheckoutProtocol`) still uses `Protocol` in the name, which isn't in line with Swift style guide.

scannillo:
A lot of these comments are styling related. I made a JIRA `DTBTSDK-1895` for us to start a Swift style guide.

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -5,11 +5,11 @@ import PaymentsCore
 #endif
 
 /// PayPal Paysheet to handle PayPal transaction
+/// encapsulates instance to communicate with nxo
 public class PayPalClient {
```
scannillo, line 9: This is a public facing docstring, so it's message should be helpful to a merchant. We want to obfuscate "nxo" details from a merchant, and have them not even need to know what NXO is at all.

scannillo: I notice a lot of our other public classes are missing docstrings, so I created a [JIRA](https://engineering.paypalcorp.com/jira/browse/DTNOR-574) for us to tackle this holistically. 🏈

scannillo (approved):

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -1,86 +1,78 @@
...
-@_implementationOnly import PayPalCheckout
-
+import PayPalCheckout
 #if canImport(PaymentsCore)
 import PaymentsCore
 #endif
 
 /// PayPal Paysheet to handle PayPal transaction
+/// encapsulates instance to communicate with nxo
 public class PayPalClient {
```
scannillo, line 9: Also, I just saw @jaxdesmarais's comment regarding the naming of this file. I notice that our other PayPal Checkout client (for the web flow) is called `PayPalWebCheckoutClient`. I agree that this is a little confusing and their naming should be consistent.

Right now we have
`PayPalWebCheckout.PayPalWebCheckoutClient` and `PayPalNativeCheckout.PayPalClient`

scannillo: Also side note - can someone fill me in the background for splitting our PayPal flows into 2 modules on Northstar (native vs web)? Did we discuss consolidating?

sshropshire: Yeah it's `PayPalNativeCheckout`/`PayPalClient` and `PayPalWebCheckout`/`PayPalWebCheckoutClient` on Android we'll have to rename in the future too.

I think the reasoning was we were able to implement web checkout faster initially, and to offer support for some merchants who want a hyper-optimized app with only the PayPal web flow.

Might not have been the exact reasons but I remember those coming up in conversation.

kgangineni: @scannillo agree, naming is incosistent, is it good time to change, or let's keep it consistent with android?

scannillo: Ah gotcha - you mean hyper-optimized to not include the PPNXO binary.

Makes sense to me to tackle it alongside Android in a later PR 🏈

Demo/Demo/ViewModels/BaseViewModel.swift
jaxdesmarais: I think this still needs to be addressed unless we are planning to do this in a follow up ticket.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,27 +4,38 @@ import PayPalWebCheckout
...
+    func paypalDidShippingAddressChange(
+        _ payPalClient: PayPalClient,
+        shippingChange: ShippingChange,
+        shippingChangeAction: ShippingChangeAction
+    ) {
+        updateTitle("shipping address changed")
+    }
+
+    func paypal(_ payPalClient: PayPalClient, didFinishWithResult approvalResult: Approval) {
+        guard let orderId = orderID else {
```
```suggestion
        guard let orderID = orderID else {
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -4,27 +4,38 @@ import PayPalWebCheckout
...
+            updateTitle("native checkout result: \(approvalResult.data.intent.stringValue)")
+            return
+        }
+        updateTitle("order \(orderId): \(approvalResult.data.intent.stringValue)")
```
```suggestion
        updateTitle("order \(orderID): \(approvalResult.data.intent.stringValue)")
```

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -1,86 +1,78 @@
...
         self.apiClient = apiClient
     }
 
     /// Present PayPal Paysheet and start a PayPal transaction
     /// - Parameters:
-    ///   - request: the PayPalRequest for the transaction
     ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
-    ///   - completion: Completion block to handle buyer's approval, cancellation, and error.
-    public func start(request: PayPalRequest, presentingViewController: UIViewController? = nil) async {
+    ///   - delegate: Completion block to handle buyer's approval, cancellation, error, create order callback, shipping change callback
```
jaxdesmarais, line 36: I think this is missing a docstring for the `orderID` param. I think we'd also want the orderID first since it's required?

Sources/PayPalNativeCheckout/Extensions/Environment+Extension.swift
```diff
@@ -1,4 +1,4 @@
-@_implementationOnly import PayPalCheckout
+import PayPalCheckout
```
jaxdesmarais, line 1: Were these moved to `import` instead of `@_implementationOnly` because of how we are integrating the SDK now? Just curious if that's the reasoning for the change or something else I am missing.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
```diff
@@ -0,0 +1,33 @@
+import Foundation
+import UIKit
+import PayPalCheckout
+#if canImport(PaymentsCore)
+import PaymentsCore
+#endif
+
+class NativeCheckoutProvider: NativeCheckoutStartable {
+
+    // swiftlint:disable function_parameter_count
```
jaxdesmarais, line 10: Do we have a follow up ticket to remove this rule?

jaxdesmarais (approved):
Just a few general questions, not sure if they are to be resolved in follow up tickets or should be addressed in this PR.

Sources/PayPalNativeCheckout/PayPalUIInterfaces/NativeCheckout.swift
kgangineni: yep there's a ticket to create create/update lint rules

Sources/PayPalNativeCheckout/Extensions/Environment+Extension.swift
kgangineni: yep previously we were not exposing callbacks, now were doing so it can't be `@_implementationOnly`

PR #71, opened 2022-09-01
NXO Demo: architecture and basic UI

jcnoriega:
### Reason for changes
This creates the basic architecture for the NXO demo. The UI will be completed on a diferent PR. This adds the use cases for the diferent features that NXO exposes. Some of these are still in development.

### Summary of changes

-  Create Use Cases
-  Creates a Basic Swift UI layout

### Checklist

- [ ] Added a changelog entry

- @jcnoriega 
- @kgangineni


Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -33,8 +33,11 @@ public class PayPalClient {
     /// Present PayPal Paysheet and start a PayPal transaction
     /// - Parameters:
     ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
-    ///   - delegate: Completion block to handle buyer's approval, cancellation, error, create order callback, shipping change callback
-    public func start(presentingViewController: UIViewController? = nil, orderID: String, delegate: PayPalDelegate?) async {
```
kgangineni, line 37: @jcnoriega we want to take away the ability to pass delegate in method?
same for a simpler method with orderId, we just want one single method with a callback?

Demo/Demo/UseCases/GetBillingAgreementToken.swift
```diff
@@ -0,0 +1,38 @@
+import UIKit
+
+final class GetBillingAgreementToken {
+
+    func execute() async throws -> Order {
+        return try await DemoMerchantAPI.sharedService.createOrder(order: GetBillingAgreementToken.baTokenRequest)
+    }
+
+    private static let baTokenRequest = """
+        {
```
scannillo: Here's a [good site](https://www.hackingwithswift.com/read/7/3/parsing-json-using-the-codable-protocol) that talks about JSON Parsing in swift:
> Swift has built-in support for working with JSON using a protocol called Codable. When you say “my data conforms to Codable”, Swift will allow you to convert freely between that data and JSON using only a little code.

It's common practice to use `Codable` for encoding/decoding JSON, like we do [here](https://github.com/paypal/iOS-SDK/blob/main/Demo/Demo/Models/CreateOrderParams.swift) in `CreateOrdersParams`. 

It also then allows us to separate out our "business logic" of determining what data to populate our JSON request with, from the model layer ([see BaseViewModel](https://github.com/paypal/iOS-SDK/blob/4ef75dcf25609846445dc5bd2eac70b975a41798/Demo/Demo/ViewModels/BaseViewModel.swift#L84-L88)).

scannillo: This looks like the same request as in [CreateOrderParams.swift](https://github.com/paypal/iOS-SDK/blob/main/Demo/Demo/Models/CreateOrderParams.swift), but with additional values.

We can extend the existing model with optional params, something like:

```
    let intent: String
    let purchaseUnits: [PurchaseUnit]
    let applicationContext: ApplicationContext
    let paymentSource: PaymentSource?
}

struct PurchaseUnit: Codable {
    let amount: Amount
}

struct Amount: Codable {
    let currencyCode: String
    let value: String
}

struct ApplicationContext: Codable {
    let returnUrl: String
    let cancelUrl: String
    let shippingPreference: String?
}

struct PaymentSource: Codable {
    let paypal: Attributes
}

struct Attributes: Codable {
    let vault: Vault
}

struct Vault: Codable {
    let confirmPaymentToken: String
    let usageType: String
}
```

scannillo: Also, `struct`s are pass by value (instead of `class`es pass by reference) and are preferred in Swift for data/model types as they're thread safe. They're also lighter weight and faster than classes in Swift. Here's a [good read ](https://www.avanderlee.com/swift/struct-class-differences/)on it.

Demo/Demo/UseCases/GetOrderIdUseCase.swift
```diff
@@ -0,0 +1,112 @@
...
+import UIKit
+
+final class GetOrderIdUseCase {
+
+    func execute() async throws -> String {
+        let order = try await DemoMerchantAPI.sharedService.createOrder(order: GetOrderIdUseCase.orderWithFixedShipping)
+        return order.id
+    }
+
+    private static let orderWithFixedShipping = """
```
scannillo: Same comment here about using a `Codable` struct.

scannillo: Also happy to pair on this too!

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,135 @@
+import UIKit
+import PayPalNativeCheckout
+import PayPalCheckout
+import PaymentsCore
+
+class PayPalViewModel: ObservableObject, PayPalDelegate {
```
scannillo: Good idea to separate this PayPalViewModel out of the BaseViewModel!

scannillo: However, I think `BaseViewModel` was made with the intent of being subclassed by each feature's ViewModel. It was meant to house shared functionality b/w all ViewModels. So for example, we have some redundancies now b/w the BaseVM and this new VM such as creating an access token and fetching an orderID.

Demo/Demo/UseCases/GetAccessToken.swift
```diff
@@ -0,0 +1,8 @@
+import UIKit
+
+final class GetAccessToken {
+
+    func execute() async -> String? {
```
scannillo: Why are we wrapping `getAccessToken()` in another method?

Demo/Demo/Models/BAToken.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+struct BAToken: Codable {
```
jaxdesmarais:
```suggestion
struct BillingAgreementToken: Codable {
```

Demo/Demo/Models/VaultSessionId.swift
```diff
@@ -0,0 +1,7 @@
+import UIKit
+
+struct VaultSessionId: Codable {
```
jaxdesmarais:
```suggestion
struct VaultSessionID: Codable {
```

Will also need to update the file name too

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
         let data = try await data(for: urlRequest)
         return try parse(from: data)
     }
+
+    /// This function replicates a way a merchant may go about creating an order on their server and is not part of the SDK flow.
+    /// - Parameter order: order in json format
```
```suggestion
    /// - Parameter order: order JSON in String format
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
...
+    /// - Returns: an order
+    /// - Throws: an error explaining why create order failed
+    func createOrder(order: String) async throws -> Order {
+        guard let url = buildBaseURL(with: "/orders") else {
+            throw URLResponseError.invalidURL
+        }
+        var urlRequest = URLRequest(url: url)
```
```suggestion
        }

        var urlRequest = URLRequest(url: url)
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
...
+        urlRequest.httpMethod = "POST"
+        urlRequest.httpBody = Data(order.utf8)
+        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        let data = try await data(for: urlRequest)
+        return try parse(from: data)
+    }
+
+    func createApprovalSessionId(accessToken: String, approvalSessionRequest: String) async throws -> VaultSessionId {
```
```suggestion
    func createApprovalSessionID(accessToken: String, approvalSessionRequest: String) async throws -> VaultSessionID {
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
...
+        guard let url = buildPayPalURL(with: "/v2/vault/payment-tokens") else {
+            throw URLResponseError.invalidURL
+        }
+        var urlRequest = URLRequest(url: url)
```
```suggestion
        }

        var urlRequest = URLRequest(url: url)
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
...
+        urlRequest.httpMethod = "POST"
+        urlRequest.httpBody = Data(approvalSessionRequest.utf8)
+        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        urlRequest.addValue("Bearer \(accessToken)", forHTTPHeaderField: "Authorization")
+        let data = try await data(for: urlRequest)
+        return try parse(from: data)
+    }
+
+    func createBAToken(accessToken: String, baTokenRequest: String) async throws -> BAToken {
```
jaxdesmarais: Can we write out billing agreement in these vs BA?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +24,49 @@ final class DemoMerchantAPI {
...
+        guard let url = buildPayPalURL(with: "/v1/billing-agreements/agreement-tokens") else {
+            throw URLResponseError.invalidURL
+        }
+        var urlRequest = URLRequest(url: url)
```
```suggestion
        }

        var urlRequest = URLRequest(url: url)
```

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -4,16 +4,66 @@ import PayPalUI
 
 struct SwiftUINativeCheckoutDemo: View {
 
-    @StateObject var baseViewModel = BaseViewModel()
+    enum CheckoutType: String, CaseIterable, Identifiable {
+        case order = "Order"
+        case orderId = "Order ID"
```
jaxdesmarais:
```suggestion
        case orderID = "Order ID"
```

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,135 @@
...
+    func checkoutWithOrder() {
+        startNativeCheckout {
+            let orderRequest = self.getOrderRequestUseCase.execute()
+            await self.payPalClient?.start { createOrderAction in
+                createOrderAction.create(order: orderRequest)
+            }
+        }
+    }
+
+    func checkoutWithOrderId() {
```
jaxdesmarais:
```suggestion
    func checkoutWithOrderID() {
```

Demo/Demo/UseCases/GetApprovalSessionId.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  GetApprovalSessionId.swift
+//  Demo
+//
+//  Created by Jose Noriega on 31/08/2022.
+//
+
```
jaxdesmarais:
```suggestion
```

Demo/Demo/UseCases/GetApprovalSessionId.swift
```diff
@@ -0,0 +1,45 @@
...
+import UIKit
+
+final class GetApprovalSessionId {
```
jaxdesmarais:
```suggestion
final class GetApprovalSessionID {
```

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,135 @@
...
+        startNativeCheckout {
+            let orderId = try await self.getOrderIdUseCase.execute()
+            await self.payPalClient?.start { createOrderAction in
+                createOrderAction.set(orderId: orderId)
+            }
+        }
+    }
```
scannillo: I am confused the difference between these two?

Demo/Demo/UseCases/GetApprovalSessionId.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    func execute(accessToken: String) async throws -> String? {
+        let vaultSessionId = try await DemoMerchantAPI.sharedService.createApprovalSessionId(
```
jaxdesmarais:
```suggestion
        let vaultSessionID = try await DemoMerchantAPI.sharedService.createApprovalSessionID(
```

Demo/Demo/UseCases/GetBATokenWithoutPurchase.swift
```diff
@@ -0,0 +1,40 @@
+import UIKit
+
+final class GetBATokenWithoutPurchase {
+
+    func execute(accessToken: String) async throws -> String {
+        let baToken = try await DemoMerchantAPI.sharedService.createBAToken(
```
jaxdesmarais: ba > billingAgreement

Demo/Demo/UseCases/GetBillingAgreementToken.swift
jaxdesmarais: +1 to everything Sammy said

Demo/Demo/UseCases/GetOrderIdUseCase.swift
```diff
@@ -0,0 +1,112 @@
+//
+//  GetOrderIdUseCase.swift
+//  Demo
+//
+//  Created by Jose Noriega on 31/08/2022.
+//
+
```
jaxdesmarais:
```suggestion
```

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -33,8 +33,11 @@ public class PayPalClient {
...
+    ///   - createOrder: action to perform when an order has been created
+    public func start(
+        presentingViewController: UIViewController? = nil,
+        createOrder: @escaping PayPalCheckout.CheckoutConfig.CreateOrderCallback
```
scannillo, line 39: I'm a little confused about how an `orderID` is used for our merchants integrating with NXO through Northstar. For `CardClient` and `PayPalWebCheckoutClient` flows, a merchant must provide an `orderID` on their `CardRequest` or `PayPalWebCheckoutRequest`. 

For `PayPalClient` (NXO), I'm a little confused why this doesn't follow the same pattern. Also, what is this `createOrder` callback  for?

Demo/Demo/UseCases/GetOrderIdUseCase.swift
```diff
@@ -0,0 +1,112 @@
...
+import UIKit
+
+final class GetOrderIdUseCase {
```
jaxdesmarais: Id > ID

Demo/Demo/UseCases/GetOrderRequestUseCase.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    private func createOrderRequest(
+        amountValue: String = "0.01",
+        shippingPreference: OrderApplicationContext.ShippingPreference = .getFromFile,
+        userAction: OrderApplicationContext.UserAction = .continue,
+        currency: CurrencyCode = .usd,
+        orderIntent: OrderIntent = .capture,
+        processingInstruction: ProcessingInstruction? = nil
+    ) -> OrderRequest {
+        let address = OrderAddress(
```
jaxdesmarais: Can we add some spaces between the properties here?

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,135 @@
...
+
+    // MARK: - PayPalDelegate conformance
+
+    func paypalDidShippingAddressChange(
+        _ payPalClient: PayPalClient,
+        shippingChange: ShippingChange,
+        shippingChangeAction: ShippingChangeAction
+    ) {
+    }
```
jaxdesmarais: Is this a no-op or just missing implementation?

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -32,6 +32,9 @@ class PayPalClient_Tests: XCTestCase {
         func paypalDidCancel(_ payPalClient: PayPalClient) {
             paypalDidCancel = true
         }
+
+        func paypalDidStart(_ payPalClient: PayPalClient) {
+        }
```
jaxdesmarais: no-op or missing implementation?

jaxdesmarais (commented):
Thanks for digging into tacking some of this separation of concern! The direction this is going is great and I think going to clean up/clarify things quite a bit.

Sources/PayPalNativeCheckout/PayPalClient.swift
jcnoriega: In our other clients the delegate is set separately, this is consistent with what we have.
If we do a separate method for orderId, we should do it for baToken and vaultIDs, which in essence are just other types of ids. And its more consistent with the real NXO interface

Demo/Demo/UseCases/GetBillingAgreementToken.swift
jcnoriega: I did it as string because its always going to be static values, and this json is super speficic for this specific flow. Do we want to create all of these structs with static data which are only going to be used in one place?

Demo/Demo/UseCases/GetOrderIdUseCase.swift
jcnoriega: same! we can pair on this

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: The idea for this PR was to show how could the possible architectures be for any feautre we have to demo. So i think as a first step in that direction, i considered NXO only, and if we agree with the architecture (SwiftUI, viewmodel, usecases, etc), then we can work on the other features and try to create Base classes if we see more redundancies.
But also, thats what the useCases are for, viewmodel can use the same useCases.

Demo/Demo/UseCases/GetAccessToken.swift
jcnoriega: The idea is that the viewModels interact with useCases, and inside the usecases are the calls to the data layer. So even if its a simple request, it should be inside a useCase.

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: `checkoutWithOrderId` is for a server integration. The merchant might have created the order in his server and has the `orderId` to start NXO

`checkoutWithOrder` allows for the merchants to start NXO with an `OrderRequest` object, without a server integration. The creation of the order is handled inside NXO

jcnoriega: `checkoutWithOrderId` is for a server integration. The merchant might have created the order in his server and has the `orderId` to start NXO

`checkoutWithOrder` allows for the merchants to start NXO with an `OrderRequest` object, without a server integration. The creation of the order is handled inside NXO

Sources/PayPalNativeCheckout/PayPalClient.swift
jcnoriega: We need to expose the same functionalites NXO has. NXO allows to start the checkout flow with: an `orderId`, `baToken`, an order object and an `approvalSessionId` (vault).
Thats where the `createOrder` callback comes in place. Inside the callback the merchant specifies with what is going to start the Checkout flow.

Check in `PayPalViewModel` for the methods `checkoutWithOrder`, `checkoutWithOrderId`, `checkoutWithBillingAgreement`, `checkoutBAWithoutPurchase`, `checkoutWithVault`. They are very similiar but when they call `paypalClient.start`, what inside the `createOrder` callback is diferent for each one

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: Missing implementation for now. Going to add it on a separete PR

UnitTests/PayPalTests/PayPalClient_Tests.swift
jcnoriega: missing implementation, once we move forward with the tests

Demo/Demo/ViewModels/PayPalViewModel.swift
kgangineni: this needs more discussion, thought process was to provide a convenience function to pass just `orderId` as String instead of passing `createOrderAction` and then setting orderId,
now the question do we need that convinience function or not, if we decide not to have it, then we can remove one of the functions in above

Demo/Demo/UseCases/GetOrderIdUseCase.swift
kgangineni: @jaxdesmarais is it time to rethink this convention we're following? in nxo lib all lables are like `orderId` instead of `orderID` so we might end up with these two in many places

Sources/PayPalNativeCheckout/PayPalClient.swift
kgangineni: in noxo, instead of having param for `orderId` it has `PayPalCheckout.CheckoutConfig.CreateOrderCallback` callback to set `orderId` or any other param like `vaultSessionId`, `billingAgreementToken` etc

Demo/Demo/UseCases/GetOrderIdUseCase.swift
jaxdesmarais: NXO is an external library so can follow their own style guide which may be different by our own. That said, they do use both the Id (legacy NXO code) and ID (newer code). Talking to some folks on NXO, moving forward their style guide also indicates using `ID` for all new code but this would be a breaking change for their merchants so they haven't yet made a change for legacy code as that would need to be done in a major version.

Demo/Demo/ViewModels/PayPalViewModel.swift
scannillo: Let's add a TODO so this doesn't get forgotten.

UnitTests/PayPalTests/PayPalClient_Tests.swift
scannillo: Same comment re: adding a TODO

Demo/Demo/UseCases/GetAccessToken.swift
scannillo: I don't think we need an additional layer wrapping the data layer. In [MVVM](https://www.raywenderlich.com/6733535-ios-mvvm-tutorial-refactoring-from-mvc), the ViewModel is meant to be the bridge b/w the View and Data layers. I think the UseCase layer adds extra complexity to the Demo.

Demo/Demo/UseCases/GetBillingAgreementToken.swift
scannillo: I understand - this PR opted for static JSON as its values wont change anywhere within the demo. I think we still want to use `Codable` for a few reasons:
* It's best practice for encoding/decoding JSON in swift & more closely resembles a real merchant integration
* Allows us to define our models in 1 consistent way throughout the demo
* Allows us the flexibility to modify params if needed in the future + provides encoding/decoding error handling

Demo/Demo/Models/VaultSessionID.swift
```diff
@@ -0,0 +1,7 @@
+import UIKit
+
+struct VaultSessionID: Codable {
+
+    let status: String?
+    let links: [LinkResponse]
```
scannillo: Nitpick - We can add related struct definitions to the same file to reduce the number of new files (similar to `CreateOrderParams.swift`)

Sources/PayPalNativeCheckout/PayPalClient.swift
scannillo: > We need to expose the same functionalites NXO has

Do we have to support all NXO use cases/scenarios? Another question - does our demo need to showcase all of them?

Sources/PayPalNativeCheckout/PayPalDelegate.swift
```diff
@@ -22,6 +22,16 @@ public protocol PayPalDelegate: AnyObject {
     ///   - client: the PayPalClient associated with delegate
     func paypalDidCancel(_ payPalClient: PayPalClient)
 
+    /// Notify that the PayPal flow is about to start
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    func paypalDidStart(_ payPalClient: PayPalClient)
+
+    /// Notify when a shpping method and/or address has been change
```
scannillo:
```suggestion
    /// Notify when a shipping method and/or address has changed
```

Sources/PayPalNativeCheckout/PayPalClient.swift
kgangineni: @scannillo  this also needs further discussion, from what I see, nxo has lot of functionality and they have their demo app, if needed for demo, we need to create Jiras for functionality in demo

Demo/Demo/Models/ApplicationContext.swift
```diff
@@ -0,0 +1,17 @@
...
+    var paymentMethodPreference: PaymentMethodPreference?
+    var shippingPreference: String?
+    var userAction: String?
+    var landingPage: String?
+    var brandName: String?
+}
+
+struct PaymentMethodPreference: Codable {
+
+    let payePreferred: String
```
sshropshire:
```suggestion
    let payeePreferred: String
```

Demo/Demo/Models/ApprovalSessionRequest.swift
```diff
@@ -0,0 +1,17 @@
+struct ApprovalSessionRequest: Codable {
+
+    let customerId: String
+    let source: Source
+    let applicationContext: ApplicationContext
+}
+
+struct Source: Codable {
+
+    let paypal: PayPalSource
```
sshropshire:
```suggestion
    let payPal: PayPalSource
```

sshropshire: Not sure if there's a convention in place or if this is how it is in the backend JSON.

Demo/Demo/Models/VaultSessionID.swift
```diff
@@ -0,0 +1,13 @@
...
+}
+
+struct LinkResponse: Codable {
+
+    let href: String?
+    let rel: String?
+    let method: String?
+    let encType: String?
```
sshropshire: Can we expand the abbreviation for `encType` to better fit the [Swift Naming Guidelines](https://www.swift.org/documentation/api-design-guidelines/#avoid-abbreviations)?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -24,6 +25,59 @@ final class DemoMerchantAPI {
         let data = try await data(for: urlRequest)
         return try parse(from: data)
     }
+
+    /// This function replicates a way a merchant may go about creating an order on their server and is not part of the SDK flow.
+    /// - Parameter order: order JSON in String format
+    /// - Returns: an order
+    /// - Throws: an error explaining why create order failed
+    func createOrder(order: String) async throws -> Order {
```
sshropshire: We can remove this method it looks like it isn't called anymore.

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,168 @@
...
+
+    func checkoutWithOrder() {
+        startNativeCheckout {
+            await self.payPalClient?.start { createOrderAction in
+                createOrderAction.create(order: OrderRequestHelpers.createOrderRequest())
+            }
+        }
+    }
+
+    func checkoutWithOrderID() {
```
sshropshire, line 39:
```suggestion
    func checkout(withOrderID orderID: String) {
```
Maybe this would make more sense to make it more Swifty? Then we can call `getOrderIDWithFixedShipping()` outside of this method.

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,168 @@
...
+        }
+    }
+
+    func retry() {
+        payPalClient = nil
+        accessToken = ""
+        state = .initial
+    }
+
+    func checkoutWithOrder() {
```
sshropshire:
```suggestion
    func checkout(with order: Order) {
```
🧶: Same here for Swiftiness, though i'm not sure if `Order` is the correct type.

Demo/Demo/Models/VaultSessionID.swift
jcnoriega: `encType` comes directly from the Orders API, i dont even know what `encType` is :P

Demo/Demo/Models/ApprovalSessionRequest.swift
jcnoriega: All of this structs are a map 1:1 with the orders api.

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: the problem is that this method is called from the UI layer when the user taps the button. The orderID hasnt been generated yet. Its generated inside

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: same as above!

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,168 @@
...
+    func checkoutWithOrderID() {
+        startNativeCheckout {
+            let orderID = try await self.getOrderIDWithFixedShipping()
+            await self.payPalClient?.start { createOrderAction in
+                createOrderAction.set(orderId: orderID)
+            }
+        }
+    }
+
+    func checkoutWithBillingAgreement() {
```
scannillo: We determined in stand-up that we can remove billing agreement & vault flow from the demo. We also can remove the `Order` flow just in favor of keeping the standard OrderID flow. So we'll only remain to test 2 flows in our Demo: orderID & shipping callbacks.

Sources/PayPalNativeCheckout/PayPalClient.swift
scannillo: @kgangineni - this conversation has been discussed in person in stand-up. We will only keep the orderID & shipping callback features, the rest can be removed.

Sources/PayPalNativeCheckout/PayPalDelegate.swift
```diff
@@ -22,6 +22,16 @@ public protocol PayPalDelegate: AnyObject {
     ///   - client: the PayPalClient associated with delegate
     func paypalDidCancel(_ payPalClient: PayPalClient)
 
+    /// Notify that the PayPal flow is about to start
+    /// - Parameters:
+    ///   - client: the PayPalClient associated with delegate
+    func paypalDidStart(_ payPalClient: PayPalClient)
```
scannillo: The wording here is confusing. `didStart` means the flow has already started, whereas the docstring says "is about to start". Which one is correct?

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -24,6 +24,7 @@ class PayPalClient_Tests: XCTestCase {
         var capturedResult: Approval?
         var capturedError: CoreSDKError?
         var paypalDidCancel = false
+        var paypalDidStarted = false
```
scannillo:
```suggestion
        var paypalDidStart = false
```

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -56,32 +61,40 @@ class PayPalClient_Tests: XCTestCase {
         let delegate = MockPayPalDelegate()
         payPalClient.delegate = delegate
 
-        let orderID = "orderID"
-
         let mockPaypalDelegate = MockPayPalDelegate()
-        await payPalClient.start(orderID: orderID, delegate: mockPaypalDelegate)
+        payPalClient.delegate = mockPaypalDelegate
```
scannillo:
```suggestion
        payPalClient.delegate = mockPayPalDelegate
```

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -32,6 +33,10 @@ class PayPalClient_Tests: XCTestCase {
         func paypalDidCancel(_ payPalClient: PayPalClient) {
             paypalDidCancel = true
         }
+
+        func paypalDidStart(_ payPalClient: PayPalClient) {
+            paypalDidStarted = true
+        }
     }
```
scannillo: The `MockPayPalDelegate` implementation should live in it's own file.

UnitTests/PayPalTests/PayPalClient_Tests.swift
```diff
@@ -24,6 +24,7 @@ class PayPalClient_Tests: XCTestCase {
         var capturedResult: Approval?
```
scannillo: General Swift organization puts properties above methods. We've started to document that in our Style guide [here](https://hackmd.io/kIyOnIFIRrKY4b5wQ0dUpw?both#Code-Organization).

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -24,12 +74,40 @@ struct SwiftUINativeCheckoutDemo: View {
...
+        case .order:
+            viewModel.checkoutWithOrder()
+        case .orderID:
+            viewModel.checkoutWithOrderID()
+        case .billingAgreement:
+            viewModel.checkoutWithBillingAgreement()
+        case .baWithoutPurchase:
+            viewModel.checkoutBAWithoutPurchase()
+        case .vault:
+            viewModel.checkoutWithVault()
```
scannillo: Same comment as below, re: removing unneeded Demo types as discussed in standup.

Demo/Demo/Models/VaultSessionID.swift
scannillo: Does our demo need this field at all? I'm pretty sure if we just remove the fields we don't need from the JSON response, `Codable` is fine to ignore them if omitted from this struct, while still parsing the rest.

Sources/PayPalNativeCheckout/PayPalDelegate.swift
jcnoriega: "is about to start" is correct. What wording would you suggest? something like `paypalAboutToStart`?

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -1,22 +1,86 @@
-// swiftlint:disable space_after_main_type
-import Foundation
-
+import Card
```
jaxdesmarais: missing a space after this import

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -0,0 +1,102 @@
+import Foundation
+import Card
+import PayPalCheckout
+
+enum OrderRequestHelpers {
+
+    static let requestWithFixedShipping = CreateOrderParams(
```
jaxdesmarais: I don't think all of these fields are required - can we simplify this a bit to just demo that it works vs demo-ing everything possible in the schema? Also still curious why we need our own models vs using the PayPalCheckout models already in place for these objects since we are already importing PayPalCheckout directly?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -33,13 +34,29 @@ final class DemoMerchantAPI {
...
+        var urlRequest = URLRequest(url: url)
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        urlRequest.httpMethod = method
+        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
+        if let token = accesToken {
+            urlRequest.addValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
+        }
+        urlRequest.httpBody = try? encoder.encode(body)
+        return urlRequest
```
```suggestion
        let encoder = JSONEncoder()
        encoder.keyEncodingStrategy = .convertToSnakeCase
        
        var urlRequest = URLRequest(url: url)
        urlRequest.httpMethod = method
        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
        
        if let token = accesToken {
            urlRequest.addValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
        }
        
        urlRequest.httpBody = try? encoder.encode(body)
        return urlRequest
```
nit: mainly just to cleanup the order of these to make it easier to read/understand

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    private func getOrderIDWithFixedShipping() async throws -> String {
+        let order = try await DemoMerchantAPI.sharedService.createOrder(
+            orderParams: OrderRequestHelpers.requestWithFixedShipping
+        )
+        return order.id
+    }
+
+    private func startNativeCheckout(withAction action: @escaping () async throws -> Void) {
```
jaxdesmarais: Would `_ completion` be a better name for this param? `withAction action` took me a few to understand and look where it was used to know what was going on.

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -46,26 +49,21 @@ public class PayPalClient {
...
+            onShippingChange: { shippingChange, shippingChangeAction in
+                self.notifyShippingChange(shippingChange: shippingChange, shippingChangeAction: shippingChangeAction)
+            },
+            onCancel: {
+                self.notifyCancellation()
+            },
+            onError: { error in
+                self.notifyFailure(with: error)
+            },
+            nxoConfig: nxoConfig)
```
jaxdesmarais:
```suggestion
            nxoConfig: nxoConfig
        )
```

Demo/Demo/Helpers/OrderRequestHelpers.swift
jcnoriega: @jaxdesmarais yeah took your advice. When i started implementing shippingChange, i kinda needed to use the objects of PayPalCheckout, so i deleted all the unecessary structs and just stick with the PayPalCheckout objects

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -24,12 +56,29 @@ struct SwiftUINativeCheckoutDemo: View {
...
+        }
+    }
+
+    func startNativeCheckout() {
+        viewModel.checkoutWithOrderID()
+    }
 }
 
 @available(iOS 13.0.0, *)
 struct SiftUINativeCheckoutDemo_Preview: PreviewProvider {
```
mattwylder:
```suggestion
struct SwiftUINativeCheckoutDemo_Preview: PreviewProvider {
```

Sources/PayPalNativeCheckout/PayPalDelegate.swift
scannillo: On iOS, I think `will` is pretty common. For example in the `UIViewController` lifecycle [methods](https://developer.apple.com/documentation/uikit/uiviewcontroller), there is `viewWillAppear()` and `viewWillDisappear()`.

We should also clarify in this docstring that an `ASWebAuthenticationSession` will launch, so the merchant can perform any necessary UI customization like spinners or other cleanup.

Demo/Demo/FeatureViewControllers/NativeCheckoutDemoViewController.swift
```diff
@@ -40,8 +60,16 @@ class NativeCheckoutDemoViewController: FeatureBaseViewController {
     @objc func didTapNativeCheckoutButton() {
         Task {
             nativeCheckoutButton.startAnimating()
-            try await baseViewModel.checkoutWithNativeClient()
+            viewModel.checkoutWithOrderID()
             nativeCheckoutButton.stopAnimating()
         }
     }
+
+    @objc func didTapNativeCheckoutWithOrderButton() {
```
scannillo: This method does the same as the one above. Can we delete it? Also, we don't need two PayPal Native buttons anymore right?

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import PayPalCheckout
+
+enum OrderRequestHelpers {
+
+    static var orderAmount = 100.0
+
+
```
scannillo:
```suggestion
```

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+import PayPalCheckout
+
+enum OrderRequestHelpers {
```
scannillo, line 4: This doesn't need to be an `enum, since it has no cases

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,132 @@
...
+            orderRequest: OrderRequestHelpers.getOrderRequest()
+        )
+        return order.id
+    }
+
+    func getAccessToken() async -> String? {
+        await DemoMerchantAPI.sharedService.getAccessToken(environment: DemoSettings.environment)
+    }
+
+    // MARK: - PayPalDelegate conformance
```
scannillo: In our [Swift Style Guide draft](https://hackmd.io/kIyOnIFIRrKY4b5wQ0dUpw?both#Protocol-Conformance) we call out that protocol conformances should be moved into extensions

Demo/Demo/FeatureViewControllers/NativeCheckoutDemoViewController.swift
jcnoriega: yeah you are right. This viewcontroller doesnt work any way, because i focused on making it in Swift UI. Once all of this is merged, the idea is to delete all the view controllers

Demo/Demo/Helpers/OrderRequestHelpers.swift
jcnoriega: doing it as a normal class, it throws this warning:
`Convenience Type Violation: Types used for hosting only static members should be implemented as a caseless enum to avoid instantiation. (convenience_type)`
Seems that caseless enums are a valid thing when dealing with only static methods

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -15,15 +16,30 @@ final class DemoMerchantAPI {
...
-        let encoder = JSONEncoder()
-        encoder.keyEncodingStrategy = .convertToSnakeCase
-        urlRequest.httpBody = try? encoder.encode(orderParams)
-        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
+
+        let urlRequest = buildURLRequest(
+            method: "POST",
+            url: url,
+            body: orderParams
+        )
```
jaxdesmarais: We can probably single line this, I think it'll be under our linter for line length.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -15,15 +16,30 @@ final class DemoMerchantAPI {
...
+    func createOrder(orderRequest: PayPalCheckout.OrderRequest) async throws -> Order {
+        guard let url = buildBaseURL(with: "/orders") else {
+            throw URLResponseError.invalidURL
+        }
+
+        let urlRequest = buildURLRequest(
+            method: "POST",
+            url: url,
+            body: orderRequest
+        )
```
```suggestion
        let urlRequest = buildURLRequest(method: "POST", url: url, body: orderRequest)
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -33,13 +49,36 @@ final class DemoMerchantAPI {
...
-        var urlRequest = URLRequest(url: url)
-        urlRequest.httpMethod = "POST"
-        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
-        urlRequest.httpBody = try? JSONEncoder().encode(processOrderParams)
+
+        let urlRequest = buildURLRequest(
+            method: "POST",
+            url: url,
+            body: processOrderParams
+        )
```
```suggestion
        let urlRequest = buildURLRequest(method: "POST", url: url, body: processOrderParams)
```

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,134 @@
...
+        shippingChange: ShippingChange,
+        shippingChangeAction: ShippingChangeAction
+    ) {
+        switch shippingChange.type {
+        case .shippingAddress:
+            // If user selected new address, we generate new shipping methods
+            let availableShippingMethods = OrderRequestHelpers.getShippingMethods(baseValue: Int.random(in: 0..<6))
+
+            // If shipping methods are available, then patch order with the new shipping methods and new amount
+            patchAmountAndShippingOptions(
```
jaxdesmarais: May be able to single line these if under our line length for the linter

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,134 @@
...
+                shippingMethods: availableShippingMethods,
+                action: shippingChangeAction
+            )
+
+        case .shippingMethod:
+            // If user selected new method, we patch the selected shipping method + amount
+            patchAmountAndShippingOptions(
```
jaxdesmarais: Same here

Sources/PayPalNativeCheckout/PayPalClient.swift
```diff
@@ -33,38 +33,39 @@ public class PayPalClient {
...
         do {
             let clientID = try await apiClient.getClientID()
             let nxoConfig = CheckoutConfig(
                 clientID: clientID,
                 createOrder: nil,
                 onApprove: nil,
-                onShippingChange: nil,
+                onShippingChange: { shippingChange, shippingChangeAction in
+                    self.notifyShippingChange(shippingChange: shippingChange, shippingChangeAction: shippingChangeAction)
+                }, // need to set here, otherwise it doesnt work. Possible NXO bug?
```
jaxdesmarais: We should double check with NXO team and delete this comment after we've either confirmed it's expected or a bug

jcnoriega: It is in fact an NXO bug. I updated the comment as a reminder for us to change it later. Or should I delete it?

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -0,0 +1,128 @@
+import UIKit
+import PayPalNativeCheckout
+import PayPalCheckout
+import PaymentsCore
+import Card
```
scannillo: Why does the PayPalViewModel need to import Card?

Demo/Demo/SwiftUIComponents/SwiftUICardDemo.swift
```diff
@@ -1,4 +1,5 @@
 import SwiftUI
+import AuthenticationServices
```
scannillo: Don't think we need this?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -33,13 +41,32 @@ final class DemoMerchantAPI {
...
-        urlRequest.httpMethod = "POST"
-        urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
-        urlRequest.httpBody = try? JSONEncoder().encode(processOrderParams)
+
+        let urlRequest = buildURLRequest(method: "POST", url: url, body: processOrderParams)
         let data = try await data(for: urlRequest)
         return try parse(from: data)
     }
+
+    private func buildURLRequest<T>(method: String, url: URL, body: T, accesToken: String? = nil) -> URLRequest where T: Encodable {
```
scannillo:
```suggestion
    private func buildURLRequest<T>(method: String, url: URL, body: T, accessToken: String? = nil) -> URLRequest where T: Encodable {
```

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -61,6 +88,10 @@ final class DemoMerchantAPI {
         URL(string: DemoSettings.environment.baseURL + endpoint)
     }
 
+    private func buildPayPalURL(with endpoint: String) -> URL? {
+        URL(string: "https://api.sandbox.paypal.com" + endpoint)
+    }
+
     public func getAccessToken(environment: Demo.Environment) async -> String? {
```
scannillo: Nitpick - Group public & private methods together using `MARK`. [See style guide draft](https://hackmd.io/kIyOnIFIRrKY4b5wQ0dUpw?both#Code-Organization).

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: Ah Ok - didn't realize we had a linter rule for that

Demo/Demo/Models/ApplicationContext.swift
```diff
@@ -0,0 +1,5 @@
+struct ApplicationContext: Codable {
```
scannillo: Curious why `ApplicationContext` has it's own file, while `PurchaseUnit` and `Amount` live within the `CreateOrderParams.swift` file? I think since these types are only used for the `CreateOrderParams` struct, putting them all in 1 file makes sense.

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: @scannillo it was from old code. Wonder if there is a lint rule for unecessary imports?

Demo/Demo/Models/ApplicationContext.swift
jcnoriega: Moved!

Sources/PayPalNativeCheckout/PayPalClient.swift
jaxdesmarais: Maybe add a TODO instead of a comment or a link to their jira ticket so we remember to revisit or document this in some way?

jcnoriega: Done!

scannillo (approved):

PR #72, opened 2022-09-02
NXO Demo: complete UI

jcnoriega:
### Summary of changes

-  A fully functional SwiftUI layout implemented

### Checklist

- [ ] Added a changelog entry

-


Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -34,29 +38,45 @@ struct SwiftUINativeCheckoutDemo: View {
...
-                        ForEach(CheckoutType.allCases) {type in
-                            Text(type.rawValue)
-                        }
+                    VStack {
+                        Form {
+                            Picker("Checkout ", selection: $checkoutTypeSelection) {
+                                ForEach(CheckoutType.allCases) {type in
+                                    Text(type.rawValue)
+                                }
+                            }
```
scannillo: Woah - pretty cool you can make an entire picker view with just this much code 🤯

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -34,29 +38,45 @@ struct SwiftUINativeCheckoutDemo: View {
...
+                        }.navigationBarTitle("", displayMode: .inline)
```
scannillo:
```suggestion
                        }
```

Can we just omit this if we don't want to set a title?

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -18,6 +18,10 @@ struct SwiftUINativeCheckoutDemo: View {
 
     @State var checkoutTypeSelection = CheckoutType.order
 
+    init() {
+        UITableView.appearance().tableHeaderView = UIView(frame: CGRect(x: 0, y: 0, width: 0, height: Double.leastNonzeroMagnitude))
+        UITableView.appearance().tableFooterView = UIView(frame: CGRect(x: 0, y: 0, width: 0, height: Double.leastNonzeroMagnitude))
```
scannillo: What is this for? Trying to hide the default header and footer view?

jcnoriega: its a hacky way of reducing the space the `Form` takes. Honestly this is just taken from stackoverflow, but we can pair on this to see how it works.

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jcnoriega: same here, its a way of reducing space that `Form` takes.

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
-                        ForEach(CheckoutType.allCases) {type in
-                            Text(type.rawValue)
-                        }
+        VStack(spacing: 16) {
+            HStack {
+                Text("Checkout:")
+                    .font(.system(size: 18))
+                    .frame(alignment: .leading)
+                    .padding([.leading], 16)
+                    .padding([.trailing], 32)
```
jaxdesmarais:
```suggestion
                    .padding(.leading, 16)
                    .padding(.trailing, 32)
```

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
-        NavigationView {
-            ZStack {
-                VStack(spacing: 16) {
-                    Picker("Checkout ", selection: $checkoutTypeSelection) {
-                        ForEach(CheckoutType.allCases) {type in
-                            Text(type.rawValue)
-                        }
+        VStack(spacing: 16) {
+            HStack {
+                Text("Checkout:")
```
jaxdesmarais: Maybe since this is a picker we can update this text to something like "Checkout type selection" or something to indicate that there are multiple options. May also be able to do a list picker like in the apple docs example (though not sure if you'd run into the same issues as a form): https://developer.apple.com/documentation/swiftui/picker

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
+                    .font(.system(size: 18))
+                    .frame(alignment: .leading)
+                    .padding([.leading], 16)
+                    .padding([.trailing], 32)
+                Picker("", selection: $checkoutTypeSelection) {
+                    ForEach(CheckoutType.allCases) { type in
+                        Text(type.rawValue)
```
jaxdesmarais, line 37: Should this font size be the same as "Checkout"?

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
                     }
-                    Text(title)
-                    Text(content)
-                    Button(isFlowComplete ? "Try Again" : "Start Checkout" ) {
-                        if isFlowComplete {
-                            viewModel.retry()
-                        } else {
-                            startNativeCheckout()
-                        }
+                }.frame(maxWidth: .infinity, alignment: .leading)
```
jaxdesmarais: nit: we mix both having this on a new line vs right after the curly brace. Looking through the apple SwiftUI docs it seems like newline is the preference. Should we do the same?

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
+                    .frame(maxWidth: .infinity, alignment: .leading)
+                    .padding([.leading, .trailing, .bottom], 16)
+            }
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .top)
+            VStack {
+                Button(isFlowComplete ? "Try Again" : "Start Checkout" ) {
+                    if isFlowComplete {
+                        viewModel.retry()
+                    } else {
+                        startNativeCheckout()
```
jaxdesmarais: Take it or leave it, but we should be able to use a ternary here as well:
```
isFlowComplete ? viewModel.retry() : startNativeCheckout()
```

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,50 @@ struct SwiftUINativeCheckoutDemo: View {
...
+                .padding()
+                .frame(maxWidth: .infinity)
+                .background(.blue)
+                .cornerRadius(10)
+                .padding(.horizontal, 16)
             }
+            .frame(maxWidth: .infinity, maxHeight: .infinity, alignment: .bottom)
         }
+        .frame(maxHeight: .infinity)
+        .padding([.top], 32)
```
jaxdesmarais:
```suggestion
        .padding(.top, 32)
```

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -118,7 +117,11 @@ class PayPalViewModel: ObservableObject, PayPalDelegate {
     }
 
     func paypal(_ payPalClient: PayPalClient, didFinishWithResult approvalResult: Approval) {
-        state = .mainContent(title: "Complete", content: "OrderId: \(approvalResult.data.ecToken)", flowComplete: true)
+        state = .mainContent(
+            title: "Approved",
+            content: "OrderId: \(approvalResult.data.ecToken)\nPayerId: \(approvalResult.data.payerID)",
```
jaxdesmarais:
```suggestion
            content: "OrderID: \(approvalResult.data.ecToken)\nPayerID: \(approvalResult.data.payerID)",
```

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jcnoriega: I tried with the list picker and it has the same problems as with Form, it just takes to much space and i have to do some hacky stuff to reduce it.

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jcnoriega: Reverted the label to its original font, the picker font size is fixed. Unless i put it inside a Menu

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jaxdesmarais: That makes sense! 👍

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
```diff
@@ -31,32 +30,44 @@ struct SwiftUINativeCheckoutDemo: View {
...
-                            viewModel.retry()
-                        } else {
-                            startNativeCheckout()
-                        }
+        VStack(spacing: 16) {
+            HStack {
+                Text("Checkout type selection:")
+                Spacer()
+                Picker("", selection: $checkoutTypeSelection) {
+                    ForEach(CheckoutType.allCases) { type in
```
jaxdesmarais: Updating the `ForEach` to:
```
ForEach(CheckoutType.allCases, id: \.self) { type in
```
Will get rid of the warning `Non-constant range: not an integer range` for each of these `ForEach` blocks in Xcode.

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -12,8 +12,7 @@ class PayPalViewModel: ObservableObject, PayPalDelegate {
         case mainContent(title: String, content: String, flowComplete: Bool)
     }
 
-    @Published private(set) var state = State.initial
-
+    @Published private(set) var state = State.mainContent(title: "test", content: "test", flowComplete: true)
```
jaxdesmarais: Right now this text shows when loading the view initially. Should this be updated to empty strings? Seems odd to have "test test" show in the Demo app.

jaxdesmarais (commented):
Seeing several UI warnings in `PayPalViewModel` when running the demo app currently and fetching an access token. The warning is `Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.` from the following lines:
```
state = .loading(content: "Getting access token")
state = .mainContent(title: "Error", content: "Failed to fetch access token", flowComplete: true)
state = .mainContent(title: "Access Token", content: accessToken, flowComplete: false)
```

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jcnoriega: @jaxdesmarais once we update here with checkout with only `orderID`, this `ForEach` will go away

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: this wasnt supposed to be pushed lol, i was using it to see the `Preview` of the layout in `XCode`

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -109,26 +123,31 @@ extension PayPalViewModel: PayPalDelegate {
...
-        case .shippingMethod:
-            // If user selected new method, we patch the selected shipping method + amount
-            patchAmountAndShippingOptions(
-                shippingMethods: shippingChange.shippingMethods,
-                action: shippingChangeAction
-            )
-
-        @unknown default:
-            break
+        // A bug in NXO calls callback everytime. We have to store shipping preference for cases with no shipping
```
scannillo, line 125: Interesting - so the `getFromFile` option is basically a workaround for this NXO bug?

jcnoriega: @scannillo yep. The callback should only be called when an order has `GET_FROM_FILE` as its `shippingPreference`. But now its being called every time, so i have to filter it here for now. If i try to patch an order that has `NO_SHIPPING` or `SET_PROVIDED_ADDRESS`, i get an error while patching

Demo/Demo/SwiftUIComponents/SwiftUINativeCheckoutDemo.swift
jcnoriega: @jaxdesmarais the `ForEach` remains but now its for `ShippingType`. Im not getting the warning you mentioned, where should i see it?

jaxdesmarais: Not sure which Xcode version you are using but I was seeing it in Xcode 14 when running through the demo.

jcnoriega: I'll see if i need to update my xcode, but i pushed the change anyway!

Demo/Demo/ViewModels/PayPalViewModel.swift
scannillo: Got it - this filter makes senes as a workaround. My only comment is "get from file" as a name is a bit confusing to me.

scannillo (approved):
Changes look good! Do we have a JIRA to track the NXO team fix for the "get from file" bug? Maybe we can include a link to that JIRA next to that bug comment.

jaxdesmarais (approved):
Thanks for including the gif and addressing comments, this looks good ⛵

PR #76, opened 2022-10-07
NXO: Bump version to 0.109.0

jcnoriega:
### Reason for changes
Bump to latest nxo version, where the following bugs were fixed:

Shipping callback ignored if set in `start` function: https://engineering.paypalcorp.com/jira/browse/DTNATIVEXO-1268
Shipping callback called everytime, even when no shipping allowed: https://engineering.paypalcorp.com/jira/browse/DTNATIVEXO-1281

### Summary of changes

- Change NXO version
- Remove setting onShippingCallback on CheckoutConfig
- Remove `.getFromFile` check in shipping callback

### Checklist

- [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Demo/Demo/ViewModels/PayPalViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalViewModel.swift
@@ -122,32 +122,25 @@ extension PayPalViewModel: PayPalNativeCheckoutDelegate {
         shippingChange: ShippingChange,
         shippingChangeAction: ShippingChangeAction
     ) {
-        // A bug in NXO calls callback everytime. We have to store shipping preference for cases with no shipping
-        // https://engineering.paypalcorp.com/jira/browse/DTNATIVEXO-1281
-        if shippingPreference == .getFromFile {
-            switch shippingChange.type {
-            case .shippingAddress:
-                // If user selected new address, we generate new shipping methods
-                let availableShippingMethods = OrderRequestHelpers.getShippingMethods(baseValue: Int.random(in: 0..<6))
+        switch shippingChange.type {
+        case .shippingAddress:
+            // If user selected new address, we generate new shipping methods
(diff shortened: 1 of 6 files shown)


Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -122,32 +122,25 @@ extension PayPalViewModel: PayPalNativeCheckoutDelegate {
         shippingChange: ShippingChange,
         shippingChangeAction: ShippingChangeAction
     ) {
-        // A bug in NXO calls callback everytime. We have to store shipping preference for cases with no shipping
-        // https://engineering.paypalcorp.com/jira/browse/DTNATIVEXO-1281
-        if shippingPreference == .getFromFile {
```
jaxdesmarais, line 127: 🥳

jaxdesmarais (commented):
Since the SDK is released now, we should add a changelog entry for this. We can use the [Braintree changelog](https://github.com/braintree/braintree_ios/blob/master/CHANGELOG.md#5140-2022-10-05) as a reference for how we want to format these if needed.

kgangineni (approved):

Demo/Demo/ViewModels/PayPalViewModel.swift
mattwylder: I don't fully understand the comment that's being deleted. _This_ delegate method was being called too often?

mattwylder (approved):

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: @mattwylder this delegate method is supposed to be called when an order supports changing shipping methods and/or adress. The bug on NXO was that this method got called with every order, even with ones that dont even have shipping.

PR #77, opened 2022-10-07
Bump to iOS 14 and swift tools version 5.7

jcnoriega:
### Reason for changes
Update to iOS 14 and swift tools version 5.7:
https://engineering.paypalcorp.com/jira/browse/DTNOR-593

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

- @jcnoriega


.github/workflows/build.yml
```diff
@@ -10,8 +10,8 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v2
-      - name: Use Xcode 13
-        run: sudo xcode-select -switch /Applications/Xcode_13.2.1.app
+      - name: Use Xcode 14
+        run: sudo xcode-select -switch /Applications/Xcode_14.0.1.app
```
jaxdesmarais, line 14: We are currently setting the `runs-on`  to `macOS-latest` which [according to their docs is MacOS 11](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). This version currently supports up to Xcode 13.2.1 [according to the runner image readme](https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode). We will need to bump this to run on [macOS-12 to support Xcode 14](https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode).

jcnoriega: @jaxdesmarais thank you! i was getting kinda lost as to why this wasnt working.

jaxdesmarais: 👍 The runners are a bit odd and the docs are always harder then they need to be to find. Lol.

mattwylder (approved):

jaxdesmarais (approved):
👻 This changes look good! I think we can chat with the team at stand about if we need a changelog for this.

PR #81, opened 2022-10-18
Cleanup swiftlint.yml

scannillo:
### Reason for changes
Is it just me, or is it really annoying that the default Xcode setting for auto-formatting (CTRL + i) causes trailing whitespace warnings for newlines with our linter?

We have two options here: set `ignores_empty_lines` to `true` in our swiftlint.yml, which will respect Xcode's default behavior of CTRL + i **OR** we leave the linter settings, and everyone updates their Xcode preferences by ticking the box for "_Include whitespace only lines_".

I would rather avoid manual steps for new devs Xcode setup, and just go with the Xcode default formatting implementation.

### Summary of changes

- Prefer Xcode auto formatting implementation for whitespace on newlines by setting `ignores_empty_lines` to `true`.
- Remove unneeded `disabled_rules`.

### Checklist

- ~Added a changelog entry~

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -9,11 +9,7 @@ included:
   - UnitTests
  
 disabled_rules:
-  - discarded_notification_center_observer
-  - notification_center_detachment
-  - orphaned_doc_comment
   - todo
-  - unused_capture_list
   - type_name # tests will have have the format <SUT>_Tests
   - xctfail_message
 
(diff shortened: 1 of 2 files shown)


jaxdesmarais (approved):
This makes sense to me to not have to add additional set up steps if we can help it.

jcnoriega (approved):

PR #84, opened 2022-10-20
PayPalUI: Separate buttons for SwiftUI and UIKit

jcnoriega:
### Reason for changes

The buttons all extended from `PaymentButton`, which is in essence a `UIButton`. In order to make the buttons available on `SwiftUI`, the buttons conformed to the `UIViewRepresentable` protocol as well. As of iOS 16, we started getting: `Fatal error: UIViewRepresentables must be value types: PayPalButton`. Any `SwiftUI` view must be a struct and cannot inherit from any class, only conform to protocols. So, this PR splits the buttons into two classes, one for `UIKit` and another one for `SwiftUI`. The `SwiftUI` ones are essentially a container to the`UIKit` buttons.

### Summary of changes

- Refactor to split the buttons into UIKit and SwiftUI implementations
- UIKit Buttons: `UIPaPalButton`, `UIPayPalPayLaterButton` and `UIPayPalCreditButton`
- SwiftUI buttons: `PayPalButton`, `PayPalPayLaterButton` and `PayPalCreditButton`
### Checklist

- [ ] Added a changelog entry

- @jcnoriega


jaxdesmarais:
I feel like having 2 different name for the buttons 1 for merchants using SwiftUI and 1 for merchants using UIKit is going to be confusing from a docs/integration point of view. We may have to make things more complicated on ourselves, but I wonder if there is a way to keep it consistent. This will also make it easier when merchants are moving from UIKit to SwiftUI if the button names are consistent and don't have to be tracked down. Thoughts?

jcnoriega:
@jaxdesmarais I added the `UI` prefix to the UKit buttons because it kinda seemed like the iOS convention (UIButton, UIView, etc), yet all of SwiftUI classes dont have that prefix. The buttons need to be split because of iOS 16, so they have to have diferent names, unless there is a hacky way of doing it on Swift that im not aware and can have classes with the same name (using extensions perhaps?). Or are you suggesting we directly hide the UKit buttons?

jaxdesmarais:
> @jaxdesmarais I added the `UI` prefix to the UKit buttons because it kinda seemed like the iOS convention (UIButton, UIView, etc), yet all of SwiftUI classes dont have that prefix. The buttons need to be split because of iOS 16, so they have to have diferent names, unless there is a hacky way of doing it on Swift that im not aware and can have classes with the same name (using extensions perhaps?). Or are you suggesting we directly hide the UKit buttons?

I think as far as a hacky way, it would probably require a protocol of some sort since the struct should be able to conform to a protocol. Then yeah like you said probably using extensions. I think we will want to allow merchants to use both a SwiftUI and UIKit version since SwiftUI is still early in its adoption. Just mainly trying to think through from a docs point of view if having 2 names will be confusing or not, though I do understand Apple already does this (UIButton vs Button) so maybe it's fine. 🤔

Sources/PayPalUI/UIPayPalButton.swift
```diff
@@ -0,0 +1,61 @@
+import UIKit
+import SwiftUI
+
+/// Configuration for PayPal button
+public final class UIPayPalButton: PaymentButton {
```
mattwylder, line 5: The name `UIPayPalButton` naming scheme is not the clearest in my opinion. Outside of this file it can be unclear whether a `UIPayPalButton` or `UIPayPalPayLaterButton` is a UIKit Button or a SwiftUI button. I can't think of an alternative though. Maybe the UIKit ones could be `PayPalUIButton`?

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
```diff
@@ -74,18 +74,18 @@ struct SwiftUIPaymentButtonDemo: View {
 
                 switch selectedFunding {
                 case .payPal:
-                    PayPalButton(color: PayPalButton.Color.allCases()[colorsIndex], edges: selectedEdge, size: selectedSize).id(buttonId)
+                    PayPalButton(color: UIPayPalButton.Color.allCases()[colorsIndex], edges: selectedEdge, size: selectedSize).id(buttonId)
```
mattwylder, line 77: It feels a little unintuitive that SwiftUI refers to the UIKit button to fetch colors. Would merchants do something similar?

scannillo:
👋 The two letter namespace that you see in Apple's APIs (ex: `UIButton`, `UILabel`, `PKPaymentRequest`, `NSError`) comes from the fact that ObjectiveC has no namespacing. Thus, convention is to rely on the `UI` prefix to provide namespacing to classes and types within that framework. Since we are solely supporting Swift for Northstar, we haven't been implementing this type of namespacing/prefix.

We shouldn't prefix our types with `UI` since this a [reserved Apple prefix](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/Conventions/Conventions.html).

scannillo:
I just downloaded Xcode 14.1-RC2 and this app crash no longer happens 🎊. It felt like a bug to me that `Fatal error: UIViewRepresentables must be value types` would crash on runtime instead of surface as a compile time error. I also can't find this requirement documented in Xcode's release notes.

That said - it would be useful for us to look at how other SDKs that offer buttons handle SwifUI & UIKit interopability. Some examples of place we can look:
* Apple's `PKPaymentButton` for Apple Pay
* Apple's "Sign-In  with Apple Pay" button
* Facebook's "Sign-In  with Facebook" button
* Google's "Sign-In with Google" button

jcnoriega:
@scannillo It is a Runtime error. I started having this problem when running a simulator with iOS 16. Here is someone that is having the same issue: https://developer.apple.com/forums/thread/709433
Fully agree, lets dive on how to handle SwifUI & UIKit. It seems to me, that aside the naming convention, its just doing a `struct` that conforms to the `UIViewRepresentable`. Will look into the examples you provided

scannillo:
~@jcnoriega - Did you try the latest Xcode? This seems like an Xcode bug, instead of an issue with our implementation. Xcode-14.1 RC2.~

Darn, I didn't have `SwiftUI` enabled in our Demo app. The project still crashes on Xcode 14.1-RC2.

jcnoriega:
@scannillo @mattwylder @jaxdesmarais I digged up a bit about the examples that Sammy mentioned:
**Google Sign In**
They seem to have an [Objective C button](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignIn/Sources/GIDSignInButton.m) and a [SwiftUI view](https://github.com/google/GoogleSignIn-iOS/blob/main/GoogleSignInSwift/Sources/GoogleSignInButton.swift), completely independent from each other. So their original button already had a Prefix set up.

**PKPaymentButton**
Apple provides [PayWithApplePayButton](https://developer.apple.com/documentation/passkit/paywithapplepaybutton) for SwiftUI, but people can [wrap the PKPaymentButton](https://developer.apple.com/forums/thread/649662) as well.

**Sign in with Apple**
Apple provides [SignInWithAppleButton](https://developer.apple.com/documentation/authenticationservices/signinwithapplebutton) for SwiftUI, in contrast with [ASAuthorizationAppleIDButton](https://developer.apple.com/documentation/authenticationservices/implementing_user_authentication_with_sign_in_with_apple)

**Facebook**
Facebook has a [FBLoginButton in swift](https://github.com/facebook/facebook-ios-sdk/blob/main/FBSDKLoginKit/FBSDKLoginKit/FBLoginButton.swift) but dont provide an implementation for `SwiftUI`, seems that for them the way to go is to [wrap it to conform to UIViewRepresentable](https://stackoverflow.com/questions/60720678/fbloginbutton-conform-to-view)

What I take after all of this, is that if an `obj-c` legacy button exist, they simply remove the pre fix to create a swift button (which can be for UIKit or SwiftUI). Facebook seemed to write their button in `swift` for UIKit, and its up for developers to wrapp it for Swift UI.
So the folks at native checkout have already made their buttons in swift for UIKit, with the `@objc` annotation, adding the `PPC` prefix. Example: `PaymentButtonContainer` in `obj-c` is `PPCPaymentButtonContainer`. As we are copying and pasting their buttons but removing the "smartness", we have two paths forward IMO:
1. We provide buttons in `Swift` for `UIKit` and thats it. Specifiying in the docs how to wrap them if they wish to use `SwiftUI`
2. Same as above but we provide a SwiftUI wrapper and changing the name for the SwiftUI implementations (perhaps adding a `SwiftUI` sufix?)

Im guessing that `SwiftUI` is still in early adoption, so we could just stick to the first option. Eventually, and if necessary, we could build the buttons directly in `SwiftUI` from skratch.

jaxdesmarais:
@jcnoriega and I chatted a bit earlier today about attempting to move the logic from the PaymentButton class into a protocol then doing a protocol extension with the logic currently defined in that class. From there a struct can adopt the protocol for both the SwiftUI implementation as well as UIKit implementation. He's going to explore this a bit and also planning to chat through some other options during stand.

scannillo:
@jcnoriega - this analysis is awesome and super helpful! I agree those sound like two good paths forward, and inline with the analysis of other SDKs that you did. Let me look more at the source code myself too 👀

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
```diff
@@ -26,7 +26,7 @@ struct SwiftUIPaymentButtonDemo: View {
         ZStack {
             VStack {
                 Picker("Funding Source", selection: $fundingIndex) {
-                    ForEach(fundingSources.indices) { index in
+                    ForEach(fundingSources.indices, id: \.self) { index in
```
scannillo, line 29: What is `\.self` needed for?

jcnoriega:
@scannillo thank you! Yesterday after stand, we kinda opted to go with the second option. With @jaxdesmarais, we explored if the same struct can be used for SwiftUI and UIKit, but it doesnt seem possible, as the `struct` would inevitably need to inherit from `UIButton`, and that is not possible. So we are stuck with diving in two, the only issue we have now is the naming. I was thinkning of making the SwiftUI struct be inside each button, and maybe have something like `PayPalButton` for UIKit and `PayPalButton.UI` for SwiftUI. Thoughts?

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
jcnoriega: @scannillo we get a `Non-constant range: not an integer range` warning in swift lint with the `\.self`

scannillo:
> @scannillo thank you! Yesterday after stand, we kinda opted to go with the second option. With @jaxdesmarais, we explored if the same struct can be used for SwiftUI and UIKit, but it doesnt seem possible, as the struct would inevitably need to inherit from UIButton, and that is not possible. So we are stuck with diving in two, the only issue we have now is the naming.

Got it. That makes sense to me. What about something like `PayPalUIKitButton` and `PayPalSwiftUIButton`? Though long, it is very explicit. 

>  I was thinkning of making the SwiftUI struct be inside each button, and maybe have something like PayPalButton for UIKit and PayPalButton.UI for SwiftUI. Thoughts?

This is an interesting idea. It would be cool to spike and see how it looks. It might be a little confusing though since it's essentially two nested button types? So basically a `PayPalUIKitButton.SwiftUIButton`?

jaxdesmarais:
> This is an interesting idea. It would be cool to spike and see how it looks. It might be a little confusing though since it's essentially two nested button types? So basically a `PayPalUIKitButton.SwiftUIButton`?

This is originally how the buttons were set up in the OG SwiftUI PR: https://github.com/paypal/iOS-SDK/pull/34/commits/f7bcfb211dea8b3c89826faadc9c560bc287e6a3#diff-f63d3caaeb446ce75676cc39af198241c694545df4c749791941b78cd3fff36d

This is very similar to what Stripe does: https://github.com/stripe/stripe-ios/blob/591f54d11f6f60be609e294f7dd7570c0ad32b49/IntegrationTester/IntegrationTester/Views/CardView.swift#L18

Calling them `.Representable` at least at the time was the "best practice" as the ultimate superview for SwiftUI is UIViewRepresentable. I'm not necessarily tied to any name though if we feel that something else is better.

jcnoriega:
@scannillo I think the best way to go, as @jaxdesmarais mentioned me, is to have a for example a `PayPalButton`, and then have an `extension` of the button where the `struct` would live, called `Representable`. Something like this:
```swift
class PayPalButton: PaymentButton {
    // uikit 
}

extension PayPalButton {
    struct Representable: UIViewRepresentable {
        //swiftui conformance
    }
}
```
It seems that [Stripe does something similar](https://github.com/stripe/stripe-ios/blob/591f54d11f6f60be609e294f7dd7570c0ad32b49/IntegrationTester/IntegrationTester/Views/CardView.swift#L18)

scannillo:
@jcnoriega @jaxdesmarais - that looks awesome. I love the `Representable` naming too. Thank you for including the example snippets and references 👏

mattwylder:
The `.Representable` namespacing makes sense to me! But it also assumes that SwiftUI is not widely adopted. Some day (years from now) SwiftUI will be widely adopted and a nested `struct` may not make sense. Just thinking through it out loud, but in a few years should we consider a decoupling / refactor?

jcnoriega:
@mattwylder yeah for sure. I think once `SwiftUI` becomes widely adopted, we can even consider making the buttons from scratch in pure `SwiftUI`

mattwylder:
It seems like a lot of `UIViewRepresentable` for the new buttons aren't being tested. Right now we're testing the initializer and the coordinator. SwiftUI prides itself on being very testable, but this would straddle a line between testing SwiftUI and UIKit. What does everyone think?

scannillo:
Also, the CI fixes (#87) were merged into `main`.

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
jaxdesmarais: This is a compiler warning not from the linter: https://www.hackingwithswift.com/forums/swiftui/compiler-warning-non-constant-range-argument-must-be-an-integer-literal/14878

jcnoriega:
@mattwylder im actually struggling a bit to increase the coverage in the `Representable` buttons. Seems that SwiftUI is very easy to test when its pure SwiftUI. Not so much with `UIViewRepresentable`

scannillo:
> 
> @mattwylder It seems like a lot of UIViewRepresentable for the new buttons aren't being tested. Right now we're testing the initializer and the coordinator. SwiftUI prides itself on being very testable, but this would straddle a line between testing SwiftUI and UIKit. What does everyone think?

Good callout. I agree the functionality we've added in our `UIViewRepresentable` implementations should be unit tested. But we would need to mock `UIViewRepresentableContext`, which Apple makes a pain since you can't construct an instance yourself. We will probably need an intermediate protocol for testing (wrapping `UIViewRepresentable`). 

On another note - I think the best long term solution for our UI components is snapshot view tests. Using unit tests for colors and image assets won’t catch problems like unreadable color HEX combinations, incorrect icons, missing/ mislabeled fields, etc.

That said, I think if we add TODOs regarding these concerns to this PR, we can 🏈 on doing that testing overhaul in another PR. Thoughts?

jcnoriega:
@scannillo Yes! I was having a hard time with `UIViewRepresentableContext`. Im still getting familiarised with UnitTesting on iOS, so maybe we can leave it for a different PR.

Sources/PayPalUI/PayPalButton.swift
```diff
@@ -59,29 +55,62 @@ public final class PayPalButton: PaymentButton, UIViewRepresentable {
...
+                edges: edges,
+                size: size,
+                insets: insets,
+                label: label?.label
+            )
+            self.action = action
+        }
+       
+        // MARK: - UIViewRepresentable methods
+        public func makeCoordinator() -> Coordinator {
```
jaxdesmarais:
```suggestion
        // MARK: - UIViewRepresentable methods

        public func makeCoordinator() -> Coordinator {
```

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 # PayPal iOS SDK Release Notes
 
 ## unreleased
+* PayPalUI
+    * Separate buttons into `UIKit` and `SwiftUI` implementations.
```
jaxdesmarais: Can we add a bullet here for the iOS 16 bug fix as well? Something like "Fixes fatal error crash with iOS 16 targets".

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
...
+    
+    | UIKit      | SwiftUI |
+    | ----------- | ----------- |
+    | PayPalButton      | PayPalButton.Representable       |
+    | PayPalCreditButton   | PayPalCreditButton.Representable        |
+    | PayPalPayLaterButton   | PayPalPayLaterButton.Representable        |
```
scannillo: This is awesome!

UnitTests/PayPalUI_Tests/PayPalPayLaterButton_Tests.swift
```diff
@@ -0,0 +1,32 @@
+import XCTest
+import SwiftUI
+import Foundation
+@testable import PayPalUI
+
+class PayPalPayLaterButton_Tests: XCTestCase {
+    
+    func testInit_whenPayPalPayLaterButtonnCreated_hasDefaultUIValues() {
```
scannillo:
```suggestion
    func testInit_whenPayPalPayLaterButtonCreated_hasDefaultUIValues() {
```

UnitTests/PayPalUI_Tests/PayPalCreditButton_Tests.swift
```diff
@@ -8,18 +10,27 @@ class PayPalCreditButton_Tests: XCTestCase {
...
+            expectation.fulfill()
+        }
+        let coordinator = sut.makeCoordinator()
 
-        XCTAssertNotNil(payPalCreditButton)
-        XCTAssertNotNil(payPalCreditButton.makeCoordinator())
-        XCTAssertNotNil(coordinator.onAction(action))
+        coordinator.onAction(self)
+        XCTAssertNotNil(sut)
+        waitForExpectations(timeout: 1) {error in
```
jaxdesmarais: I've not used `waitForExpectations` in this way before. Is there an advantage to capturing the error here vs `waitForExpectations(timeout: 1)`?

jaxdesmarais (approved):
Just a couple of small non-blocking comments for me. Thanks for tackling this and talking through all of the pros/cons of each solution. 🚀

UnitTests/PayPalUI_Tests/PayPalButton_Tests.swift
```diff
@@ -13,9 +13,9 @@ class PayPalButton_Tests: XCTestCase {
         XCTAssertEqual(payPalButton.containerView.backgroundColor, PaymentButtonColor.gold.color)
```
scannillo: Do we also want to update this test to match the updates you made for the other button tests? To include label, insets, collapsed, etc?

CHANGELOG.md
scannillo: Good callout!

scannillo: I think it would be helpful to call out the specific crash too `UIViewRepresentables must be value types`

UnitTests/PayPalUI_Tests/PayPalCreditButton_Tests.swift
jcnoriega: @jaxdesmarais well not in this case. It might help if we wanted to return an error with a specific message, or maybe continue with the test even if there was a failure. I removed it because it doesnt serve any purpose here

UnitTests/PayPalUI_Tests/PayPalButton_Tests.swift
jcnoriega: done!

UnitTests/PayPalUI_Tests/PayPalPayLaterButton_Tests.swift
```diff
@@ -0,0 +1,32 @@
...
+        let coordinator = sut.makeCoordinator()
+
+        coordinator.onAction(self)
+        XCTAssertNotNil(sut)
+        waitForExpectations(timeout: 1) {error in
+            if let error = error {
+                XCTFail("waitForExpectationsWithTimeout errored: \(error)")
+            }
+        }
+    }
```
scannillo: This is a really neat!

Originally I was thinking we could do something like:
```
func testSample() {
    let fakeAction: () -> Void = { print("Fake action") }
    let sut = PayPalPayLaterButton.Representable(fakeAction)
    let coordinator = sut.makeCoordinator()
        
    XCTAssertTrue(coordinator.action === fakeAction)
}
```
But this isn't possible since [Swift closures aren't equatable](https://www.jessesquires.com/blog/2021/04/05/why-swift-closures-are-not-equatable/).

scannillo: I really like your approach with this test! Some minor fixups to make the test easier to read, including renaming it  to be more clear that we're testing using `onAction`:

```suggestion
      // MARK: - PayPalPayLaterButton.Representable for SwiftUI

      func testMakeCoordinator_whenOnActionIsCalled_executesActionPassedInInitializer() {
        let expectation = expectation(description: "Action is called")
        let sut = PayPalPayLaterButton.Representable {
            expectation.fulfill()
        }
        let coordinator = sut.makeCoordinator()

        coordinator.onAction(self)
        waitForExpectations(timeout: 1) {error in
            if error != nil {
                XCTFail("Action passed in PayPalPayLaterButton.Representable was never called.")
            }
        }
    }
```

scannillo: We could also then add a MARK `// MARK: - PayPalPayLaterButton for UIKit` surrounding the UIKit button tests.

UnitTests/PayPalUI_Tests/PayPalCreditButton_Tests.swift
scannillo: I think the main benefit to the way Jose had it was it lets us show a custom failure message if `waitForExpectations` isn't called. I think this is most helpful when CI fails, we get a more detailed reason why.

Sources/PayPalUI/PayPalPayLaterButton.swift
```diff
@@ -41,29 +38,75 @@ public final class PayPalPayLaterButton: PaymentButton, UIViewRepresentable {
...
+                color: color.color,
+                edges: edges,
+                size: size,
+                insets: insets,
+                label: .payLater
+            )
+            self.action = action
+        }
+
+        // MARK: - UIViewRepresentable methods
```
scannillo, line 76: Can we leave a TODO + create a JIRA for unit testing the rest of the `UIViewRepresentable` implementation? (For a follow up PR)

UnitTests/PayPalUI_Tests/PayPalPayLaterButton_Tests.swift
```diff
@@ -21,6 +25,10 @@ class PayPalPayLaterButton_Tests: XCTestCase {
 
         coordinator.onAction(self)
         XCTAssertNotNil(sut)
```
scannillo:
```suggestion
```

Don't think this is relevant/needed for this test

UnitTests/PayPalUI_Tests/PayPalCreditButton_Tests.swift
```diff
@@ -26,6 +30,10 @@ class PayPalCreditButton_Tests: XCTestCase {
 
         coordinator.onAction(self)
         XCTAssertNotNil(sut)
```
scannillo:
```suggestion
```

UnitTests/PayPalUI_Tests/PayPalButton_Tests.swift
```diff
@@ -26,6 +30,10 @@ class PayPalButton_Tests: XCTestCase {
 
         coordinator.onAction(self)
         XCTAssertNotNil(sut)
```
scannillo:
```suggestion
```

scannillo (approved):
👏 This looks awesome. Huge shoutout for doing such thorough research and analysis for us to come up with the best solution possible, and for a tricky problem too!

UnitTests/PayPalUI_Tests/PayPalCreditButton_Tests.swift
scannillo: Nitpick

UnitTests/PayPalUI_Tests/PayPalButton_Tests.swift
scannillo: Nitpick

UnitTests/PayPalUI_Tests/PayPalPayLaterButton_Tests.swift
scannillo: Nitpick

PR #86, opened 2022-10-26
Consolidate URL header creation

scannillo:
### Reason for changes

- Standardize pattern for setting headers across all API Requests:
  - In most places, we were setting headers in the `APIRequest` protocol conformance (for example [here](https://github.com/paypal/iOS-SDK/blob/4b6c3da989e17f667f4f5b878c22d015aeb73798/Sources/PaymentsCore/GetClientIDRequest.swift#L16) and [here](https://github.com/paypal/iOS-SDK/blob/4b6c3da989e17f667f4f5b878c22d015aeb73798/Sources/Card/GetOrderInfoRequest.swift#L20)). In[ another place](https://github.com/paypal/iOS-SDK/blob/4b6c3da989e17f667f4f5b878c22d015aeb73798/Sources/PaymentsCore/Networking/APIClient.swift#L27), we were setting headers globally.

### Summary of changes

- Move the setting of all headers into `APIRequest` protocol implementations.
- Use `MARK`s to distinguish `APIRequest` protocol conformance code.
- Add comment for future `APIRequest` protocol refactor.
- Add missing test value for auth header in `ConfirmPaymentSourceRequest`.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jcnoriega (approved):

mattwylder:
This looks good to me! It also raises some other ideas for refactors. Could we ultimately move `headers` into `APIRequest`? How many of our requests are `GET` vs `POST`? Could there be protocols like `GETRequest` and `POSTRequest`? Just thinking out loud.

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -38,10 +26,29 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+    // MARK: - APIRequest
+
+    typealias ResponseType = ConfirmPaymentSourceResponse
+
+    var path: String
+    var method: HTTPMethod = .post
+    var body: Data?
+
+    var headers: [HTTPHeader: String] {
+        return [
```
jaxdesmarais: We should be able to remove the return keyword here and rely on the implicit return

jaxdesmarais (approved):
👻

mattwylder (approved):

Sources/Card/ConfirmPaymentSourceRequest.swift
scannillo: Good call! Pushed the fix - let me know how that looks

jaxdesmarais: 🚀  looks great!

scannillo:
> This looks good to me! It also raises some other ideas for refactors. Could we ultimately move headers into APIRequest? How many of our requests are GET vs POST? Could there be protocols like GETRequest and POSTRequest? Just thinking out loud.

Agreed - there is room to make the current setup easier to read / easier to test. I left a comment for a future networking layer re-write/investigation in this PR.

PR #90, opened 2022-11-03
Card: Add ApplicationContext to ConfirmPaymentSourceRequest

jcnoriega:
### Reason for changes

There was a bug in `ordersv2` where we had to add `appliction_context` to the order creation request, in order for 3DS to work. Now, we only have to add it in the `confirm_payment_source` request, removing it from order creation

Jira Ticket: https://engineering.paypalcorp.com/jira/browse/DTNOR-606

### Summary of changes

- Add Application context to ConfirmPaymentSourceRequestFactory

### Checklist

- [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Demo/Demo/ViewModels/BaseViewModel.swift
+++ b/Demo/Demo/ViewModels/BaseViewModel.swift
@@ -57,8 +57,7 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         let amountRequest = Amount(currencyCode: "USD", value: amount)
         let orderRequestParams = CreateOrderParams(
             intent: DemoSettings.intent.rawValue.uppercased(),
-            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
-            applicationContext: ApplicationContext(returnUrl: BaseViewModel.returnUrl, cancelUrl: BaseViewModel.cancelUrl)
+            purchaseUnits: [PurchaseUnit(amount: amountRequest)]
         )
 
         do {
--- a/Sources/Card/ConfirmPaymentSourceRequest.swift
+++ b/Sources/Card/ConfirmPaymentSourceRequest.swift
@@ -5,12 +5,12 @@ import PaymentsCore
 
 /// Describes request to confirm a payment source (approve an order)
 struct ConfirmPaymentSourceRequest: APIRequest {
-
+    
     private let orderID: String
     private let pathFormat: String = "/v2/checkout/orders/%@/confirm-payment-source"
     private let accessToken: String
     private let jsonEncoder = JSONEncoder()
-
+    
     /// Creates a request to attach a payment source to a specific order.
     /// In order to use this initializer, the `paymentSource` parameter has to
     /// contain the entire dictionary as it exists underneath the `payment_source` key.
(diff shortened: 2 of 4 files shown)


Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -57,8 +57,7 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         let amountRequest = Amount(currencyCode: "USD", value: amount)
         let orderRequestParams = CreateOrderParams(
             intent: DemoSettings.intent.rawValue.uppercased(),
-            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
-            applicationContext: ApplicationContext(returnUrl: BaseViewModel.returnUrl, cancelUrl: BaseViewModel.cancelUrl)
+            purchaseUnits: [PurchaseUnit(amount: amountRequest)]
```
jaxdesmarais, line 60: 🥳

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -5,50 +5,75 @@ import PaymentsCore
...
     ) throws {
+        var confirmPaymentSourceBody = ConfirmPaymentSourceBody()
         var card = cardRequest.card
         if let threeDSecureRequest = cardRequest.threeDSecureRequest {
             let verification = Verification(method: threeDSecureRequest.sca.rawValue)
             card.attributes = Attributes(verification: verification)
+            
+            let applicationContext = ApplicationContext(
+                returnUrl: threeDSecureRequest.returnUrl,
+                cancelUrl: threeDSecureRequest.cancelUrl
```
scannillo, line 29: It looks like we need to update our [GitHub docs ](https://github.com/paypal/iOS-SDK/tree/main/docs/Card#5-create-a-request-containing-the-card-payment-details)to include a code snippet on adding these URLs to the `ThreeDSecureRequest`

Sources/Card/ConfirmPaymentSourceRequest.swift
```diff
@@ -5,50 +5,75 @@ import PaymentsCore
...
-
+    
     var headers: [HTTPHeader: String] {
         [
             .contentType: "application/json", .acceptLanguage: "en_US",
             .authorization: "Bearer \(accessToken)"
         ]
     }
+    
+    private struct ConfirmPaymentSourceBody: Encodable {
```
jaxdesmarais: nit: are we able to just call this `ConfirmPaymentSource`?

jaxdesmarais (approved):
👻

Sources/Card/ConfirmPaymentSourceRequest.swift
jcnoriega: done!

docs/Card/README.md
```diff
@@ -144,9 +144,15 @@ Optionally, a merchant app can request Strong Consumer Authentication (SCA) for
...
 ```swift
-cardRequest.threeDSecureRequest = ThreeDSecureRequest(sca: .always)
+cardRequest.threeDSecureRequest = ThreeDSecureRequest(
+    sca: .always,
+    returnUrl: "\(Bundle.main.bundleIdentifier)://return_url",
+    cancelUrl: "\(Bundle.main.bundleIdentifier)://cancel_url"
+)
 ```
 
+Where `return_url` and `cancel_url` are the urls configured in your [paypal developer dashboard](https://developer.paypal.com/)
```
scannillo, line 154: I remember Jon mentioned that this was no longer a requirement, for merchants to need to upload their URLS to the dev dashboard. Do we know which version this requirement was dropped in?

jcnoriega: @scannillo i think Jon meant for that for NXO. If i dont put it here, i get an error on the browser.

scannillo: Woops, yes that is for NXO!

scannillo (approved):
Changes look good to me!

I have a question though - is there a benefit to documenting that a merchant use a separate return and cancel URL? This would work fine with 1 return URL (for example `<bundleID>://payments`).

sshropshire:
> I have a question though - is there a benefit to documenting that a merchant use a separate return and cancel URL?

Can we simplify the integration on iOS with `ASWebAuthenticationSession`? I know on BT we removed the requirement for the merchant to provide a custom return url scheme. That could be one less step for merchants unless we want 100% feature parity with Android.

jcnoriega:
@sshropshire the `return_url` and `cancel_url` are a requirment of `ordersV2` api for the 3DS flow right? the `CardClient` already uses a `ASWebAuthenticationSession`. I dont think we can escape the setting of the urls, it might be that we can just ask for `return_url` only, and then set the `cancel_url` on `ApplicationContext` internal ourselves. I think what @scannillo points out is true, we might not need the urls to be diferent

sshropshire:
> @sshropshire the `return_url` and `cancel_url` are a requirment of `ordersV2` api for the 3DS flow right?

Yeah definitely. My question was more of a DX one. I think on Braintree SDK the merchant doesn't need to specify either because of the way iOS works. There's no app switching in iOS, so the return and cancel urls can technically be defined by the SDK internally to make the integration more seamless.

jcnoriega:
@sshropshire good point. I will explore if we can set it within the iOS SDK.

jcnoriega:
@sshropshire @scannillo will create a ticket to follow up with the following: try to set `return_url` and `cancel_url` within the SDK, if not possible, just ask for `return_url`.

PR #91, opened 2022-11-03
Initial FPTI Networking Capability

scannillo:
### Reason for changes

- We want Northstar Mobile SDKs to send all analytics to FPTI. This PR sets up a very basic implementation of sending an analytic event to FPTI via `APIClient.sendAnalyticsEvent()`.
- In follow PRs we will address:
    - Batch event uploads
    - Refactoring into separate AnalyticsAPI if necessary / aligning with Android
    - Sending actual events in source code
    - Testing the `sendAnalyticsEvent()` function, once it is actually used 

### Summary of changes

- Add `AnalyticsEventRequest` type which conforms to `APIRequest`
- Add `AnalyticsEventRequest_Tests` for validating metadata sent
- Add encodable models for Analytics JSON post body

### Checklist

- ~Added a changelog entry~ Will update CHANGELOG once events are sent.

### Authors
@mattwylder @scannillo


Sources/PaymentsCore/AnalyticsPayload.swift
```diff
@@ -0,0 +1,93 @@
+import UIKit
+
+struct AnalyticsPayload: Encodable {
```
scannillo: Open to better naming ideas for this top-level type.

Sources/PaymentsCore/PaymentsCoreConstants.swift
```diff
@@ -0,0 +1,6 @@
+/// :nodoc: This class is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+public class PayPalCoreConstants {
+    
+    /// :nodoc: This property is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    public static let paypalSDKVersion: String = "0.0.3"
```
scannillo: Add TODO & JIRA for updating release script to modify this version.

Sources/PaymentsCore/AnalyticsEventRequest.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+struct AnalyticsEventRequest: APIRequest {
+    
+    init(payload: AnalyticsPayload) throws {
```
jaxdesmarais: I think in other files (and on Braintree) we have the properties before the init. Should we update this to match the other file formats?

Sources/PaymentsCore/AnalyticsEventRequest.swift
```diff
@@ -0,0 +1,26 @@
...
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        body = try encoder.encode(payload)
+    }
+    
+    // MARK: - APIRequest
+    
+    typealias ResponseType = EmptyResponse
```
jaxdesmarais, line 13: Just a general question, is this typealias being uses or is it for future implementation

Sources/PaymentsCore/AnalyticsPayload.swift
```diff
@@ -0,0 +1,101 @@
...
+private struct AnalyticsEvent: Encodable {
+    
+    let eventParams: AnalyticsEventParams
+    
+    init(eventName: String, sessionID: String) {
+        eventParams = AnalyticsEventParams(eventName: eventName, sessionID: sessionID)
+    }
+}
+
+private struct AnalyticsEventParams: Encodable {
```
jaxdesmarais: could we update this to either `AnalyticsEventParameters` or `AnalyticsEventDictionary` to make it a bit clearer than params?

Sources/PaymentsCore/AnalyticsPayload.swift
jaxdesmarais: 🤔  would `AnalyticsObject` or `AnalyticsEventData` be clearer? I am also not opposed to Payload, it looks like that's what some other SDKs use as well.

Sources/PaymentsCore/AnalyticsPayload.swift
```diff
@@ -0,0 +1,101 @@
...
+    private let events: AnalyticsEvent
+    
+    init(eventName: String, sessionID: String) {
+        events = AnalyticsEvent(eventName: eventName, sessionID: sessionID)
+    }
+}
+
+private struct AnalyticsEvent: Encodable {
+    
+    let eventParams: AnalyticsEventParams
```
jaxdesmarais: `eventParams` > `eventParameters` or `eventDictionary`

Sources/PaymentsCore/AnalyticsPayload.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    let component = "ppunifiedsdk"
+
+    let deviceManufacturer = "Apple"
+
+    let eventName: String
+
+    let eventSource = "mobile-native"
+
+    let packageManager: String = {
```
jaxdesmarais: General question. Do we have a preference for how we do computed properties? We use both the format here as well as this format in other places:
```
var testProperty: String {
    "my string"
}
```

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -46,4 +46,20 @@ public class APIClient {
         let (response, _) = try await fetch(endpoint: request)
         return response.clientID
     }
+    
+    /// :nodoc: This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    /// - Parameter name: Event name string used to identify this unique event in FPTI.
+    public func sendAnalyticsEvent(name: String) async {
```
jaxdesmarais: On Braintree we do the parameter for name as an anonymous parameter so it the call site we can do `client.sendAnalyticsEvent("my-super.cool.event")`. Do we prefer that over what is here `client.sendAnalyticsEvent(name: "my-super.cool.event")`. I think it reads a little cleaner but curious how y'all feel!

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -46,4 +46,20 @@ public class APIClient {
...
+        let analyticsPayload = AnalyticsPayload(
+            eventName: name,
+            sessionID: sessionID
+        )
```
jaxdesmarais: nit: I think we can single line this

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -46,4 +46,20 @@ public class APIClient {
...
+
+        do {
+            let analyticsEventRequest = try AnalyticsEventRequest(payload: analyticsPayload)
+            let (_, _) = try await fetch(endpoint: analyticsEventRequest)
+        } catch let error {
+            NSLog("[PayPal SDK] Failed to send analytics: %@", error.localizedDescription)
```
jaxdesmarais, line 59: Is our reason to use NSLog here so that we have a timestamp and log to the device as well as the debugger? I wonder if it'd be better to throw an error here directly. Thoughts?

Sources/PaymentsCore/PaymentsCoreConstants.swift
jaxdesmarais: I see in some places we use `paypal` and others `payPal`. We should probably decide on a format and update the rest of the SDK to reflect our preference. Not something for this PR but should we make a follow up ticket for that work?

UnitTests/PaymentsCoreTests/AnalyticsEventRequest_Tests.swift
```diff
@@ -0,0 +1,66 @@
+import XCTest
+@testable import PaymentsCore
+
+// swiftlint:disable force_cast
+// swiftlint:disable implicitly_unwrapped_optional
+// swiftlint:disable force_try
+// swiftlint:disable force_unwrapping
```
jaxdesmarais: take it or leave it comment. I think these can be single lined like `// swiftlint:disable force_unwrapping implicitly_unwrapped_optional` if we wanted

UnitTests/PaymentsCoreTests/AnalyticsEventRequest_Tests.swift
```diff
@@ -0,0 +1,66 @@
...
+        let urlRequest = sut.toURLRequest(environment: .sandbox)
+        XCTAssertEqual(urlRequest?.url?.absoluteString, "https://api.paypal.com/v1/tracking/events")
+    }
+}
+
+extension String {
+    
+    func matches(_ regex: String) -> Bool {
+        return self.range(of: regex, options: .regularExpression) != nil
+    }
```
jaxdesmarais: 🚀  we can remove the return here as well

Sources/PaymentsCore/Networking/APIClient.swift
jcnoriega: i would throw an error and its the responsability of the caller to handle said error. Just to keep it consistent with the behaviour in `APIClient`. I am in favour of a later refactor in a following PR, to remove this responsability from the `APIClient` and into a sort of `FPTIClient` or something.

Sources/PaymentsCore/AnalyticsPayload.swift
jcnoriega: I dont mind `AnaliticsPayload` either

Sources/PaymentsCore/AnalyticsEventRequest.swift
scannillo: Due to how our [`APIRequest` protocol ](https://github.com/paypal/iOS-SDK/blob/main/Sources/PaymentsCore/Networking/Protocols/APIRequest.swift)was designed, it is required to implement an associated, Decodable, response type. So even though this API call has no return details

Doing this PR pointed out a handful of limitations of our current networking layer. Matt and I's plan was to make this initial PR as-is, without changing any of our fundamental networking logic, and just fitting the mold of what currently exists. And then in a follow-up PR, when time permits, make broader changes to the `APIRequest` protocol (or removing it entirely).

For example another weird thing we had to do is on lines 21-23 of this file, to override the environment URL.

Sources/PaymentsCore/AnalyticsPayload.swift
scannillo: Yes, we can change the type names! We used `eventParams` for the variable name so we didn't need to use a custom CodingKey, since `encoder.keyEncodingStrategy = .convertToSnakeCase` converts to the proper format of `event_params`. But we can be more explicit with these names for sure.

scannillo: Another thing, Matt and I messed around with instead of having these multiple, nested struct types, which need to match the FPTI required structure of double nesting the `eventParams`, if we could use custom encoding to achieve the proper JSON format, while using just one `struct` type.

We got that to work and it looked like this:
```
    func encode(to encoder: Encoder) throws {
        var container = encoder.container(keyedBy: ContainerKeys.self)
        var events = container.nestedContainer(keyedBy: EventsKeys.self, forKey: .events)
        var eventParams = events.nestedContainer(keyedBy: EventParamsKeys.self, forKey: .eventParams)
        
        try eventParams.encode(eventName, forKey: .eventName)
    }
```

We decided that this was a little harder to read than just creating new struct types per level. But if others like this, we would only need one top-level struct type, `AnalyticsEventData` or whatever we call it.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: I like this! Fixed

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: > I am in favour of a later refactor in a following PR, to remove this responsability from the APIClient and into a sort of FPTIClient or something.

Agreed! This will come in a follow up, especially once we add batch uploading functionality. We also want to align our networking layer with Android's too which will make going back and forth easier.

> Is our reason to use NSLog here so that we have a timestamp and log to the device as well as the debugger?

We use logging in Braintree when analytics fail (see [flush function](https://github.com/braintree/braintree_ios/blob/b68b41623b5944f1320aa5aeabd825a1ff485a77/Sources/BraintreeCore/BTAnalyticsService.m#L156-L241) and specific example [here](https://github.com/braintree/braintree_ios/blob/b68b41623b5944f1320aa5aeabd825a1ff485a77/Sources/BraintreeCore/BTAnalyticsService.m#L230)). This makes sense to me, since I don't think we want to be returning any errors in our merchant checkout functions just if analytics fail. 

> I would throw an error and its the responsability of the caller to handle said error.

If this function returns an error, than every call-site of `FPTI/APIClient.sendAnalyticsEvent("some-event-name")` will need error handling logic. In Braintree, for example, we call this function 161 times in our source code - so it will be used a lot. The more we can streamline that call-site the better. 

Also, the call-site itself doesn't really care if the API call succeeded or failed. We don't want to return any errors to the merchants in their Checkout flows if Analytics fail.

scannillo: Let me know what you guys think about that!

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Needs unit tests either now or in a TODO when the refactor happens.

Sources/PaymentsCore/AnalyticsEventRequest.swift
scannillo: Good catch! The properties are part of the `APIRequest` protocol implementation, but the `init` is not - so we moved the `init` above the ` // MARK: - APIRequest` to clarify that everything below is part of that protocol. We do this in a few other files too ([example](https://github.com/paypal/iOS-SDK/blob/main/Sources/Card/ConfirmPaymentSourceRequest.swift)).

But I agree it is confusing. The question would be - do we prefer all properties are together, or all protocol stuff is together? I would say the later - but what do you think?

Sources/PaymentsCore/PaymentsCoreConstants.swift
scannillo: DONE! `DTNOR-625` is in the Backlog and references this convo ✅

UnitTests/PaymentsCoreTests/AnalyticsEventRequest_Tests.swift
```diff
@@ -0,0 +1,63 @@
+import XCTest
+@testable import PaymentsCore
+
+// swiftlint:disable force_unwrapping implicitly_unwrapped_optional force_try force_cast
```
scannillo, line 4: This is a sign that our linter for tests needs to be adjusted 🦵 🏈

Sources/PaymentsCore/AnalyticsPayload.swift
mattwylder: This is technically not a computed property, the property is being initialized with a closure. This is because `CodingKey` doesn't support computed variables. Sammy and I also played around with the format below for both `packageManager` and `isSimulator`, but figured the closure syntax was a little less visually cluttered.

```
#if COCOAPODS
let packageManager = "Cocoapods"
#elseif SWIFT_PACKAGE
let packageManager = "Swift Package Manager"
#else
let packageManager = "Carthage or Other"
#endif
```

@scannillo come to think of it - it may be possible to use regular computed variables if we do the custom `encode()` function 🤔

Sources/PaymentsCore/PaymentsCoreConstants.swift
mattwylder: Good catch., I _thiiiink_ the standard is `payPal`

Sources/PaymentsCore/Networking/APIClient.swift
jcnoriega: @scannillo what i meant is that the `APIClient` just sends requests, returns responses and throws errors. Its the future FPTIClient who will be responsable of managing that error. The `sendAnalytics` function will leave in the FPTIClient, something like this:

```swift
class FPTIClient {

    private let apiClient: APIClient

    init(apiClient: APIClient) {
        self.apiClient = apiClient
    }
    public func sendAnalyticsEvent(name: String) async {
        let analyticsPayload = AnalyticsPayload(
           eventName: name,
           sessionID: sessionID
        )

        do {
            let analyticsEventRequest = try AnalyticsEventRequest(payload: analyticsPayload)
            let (_, _) = try await apiClient.fetch(endpoint: analyticsEventRequest)
        } catch let error {
        }
    }
}
```

jcnoriega:
@scannillo @mattwylder most of the comments i see are better suited for the next PR where the refactor comes. This PR is just to have the minimum to make it work. So, i'll give it an approve! the next PR should be interesting

jcnoriega (approved):

Sources/PaymentsCore/AnalyticsEventRequest.swift
jaxdesmarais: Ah, gotcha - I agree I think keeping the protocol stuff together makes sense. Maybe updating protocol conformance marks to `// MARK: - APIRequest conformance` would make this clearer? This is what we've done in Braintree ([example](https://github.com/braintree/braintree_ios/blob/next-major-version/Sources/BraintreeCoreSwift/BTHTTP.swift#L518)) and I think makes it a bit clearer.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Yeh, the `APIClient` right now is a little overloaded. I like how Android has a higher separation of concerns with `API` vs `HTTP`. On iOS, we don't have that separation, and `APIClient` is our only shared service b/w feature clients.

I agree though, we'll definitely want to extract FPTI/Analytics logic into it's own client/class, but for now we figured we would just get a basic infrastructure up and running, given the existing infrastructure.

Sources/PaymentsCore/AnalyticsEventRequest.swift
jaxdesmarais: Gotcha - I think a follow up makes a lot of sense for this! Good for us to find these limitations now so we can improve the networking layer. 🚀

Sources/PaymentsCore/AnalyticsPayload.swift
scannillo: [This](https://www.donnywals.com/using-closures-to-initialize-properties-in-swift/) is a good read. 

I don't see much benefit to us exploring using a custom `encode()` function here, just to use computed properties. Using the closure lets us use `let` and is only computed the first time it's initialized.

Sources/PaymentsCore/AnalyticsEventRequest.swift
scannillo: I like that - especially the name of `APIRequest` isn't very protocol-y (another potential improvement to our networking layer).

Sources/PaymentsCore/AnalyticsPayload.swift
jaxdesmarais: 🤔  up to y'all - I don't feel like that's too difficult to parse but can see how it takes an extra bit of time to reason about if folks aren't familiar with nested containers.

Sources/PaymentsCore/AnalyticsPayload.swift
jaxdesmarais: That's a great read! Thanks for sharing. I had previously only used closure initialization for view based configurations but hadn't thought about it from this perspective.

Sources/PaymentsCore/AnalyticsEventRequest.swift
jaxdesmarais: Yeah agree - I totally missed that it was a protocol during my review just from not looking at the networking layer in a bit 🤦

Sources/PaymentsCore/Networking/APIClient.swift
jaxdesmarais: > We use logging in Braintree when analytics fail (see [flush function](https://github.com/braintree/braintree_ios/blob/b68b41623b5944f1320aa5aeabd825a1ff485a77/Sources/BraintreeCore/BTAnalyticsService.m#L156-L241) and specific example [here](https://github.com/braintree/braintree_ios/blob/b68b41623b5944f1320aa5aeabd825a1ff485a77/Sources/BraintreeCore/BTAnalyticsService.m#L230)). This makes sense to me, since I don't think we want to be returning any errors in our merchant checkout functions just if analytics fail.

This makes sense to me. My question was more around if we prefer `NSLog` over `print` for any particular reason

Sources/PaymentsCore/AnalyticsPayload.swift
scannillo: We opted to use the [custom encoder](https://github.com/paypal/iOS-SDK/pull/91/commits/f4c28d9a8557e38579e208c5e8a051f720b5b819)

Sources/PaymentsCore/AnalyticsEventRequest.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+struct AnalyticsEventRequest: APIRequest {
+    
+    init(payload: AnalyticsEventData) throws {
```
jaxdesmarais: should we update payload to something else in the init?

Sources/PaymentsCore/AnalyticsPayload.swift
```diff
@@ -0,0 +1,114 @@
+import UIKit
+
+struct AnalyticsEventData: Encodable {
```
jaxdesmarais: nit: we will need to update the file name to match

Sources/PaymentsCore/AnalyticsEventRequest.swift
scannillo: Good catch! I updated to `init(eventData: AnalyticsEventData)`

scannillo:
😢 CI keeps failing with the same issue fetching PayPalCheckout as before:

```
❌  /Users/runner/Library/Developer/Xcode/DerivedData/PayPal-ggnhwfenywtizyavwmfixszptapm/Build/Products/Debug-iphonesimulator/PayPalNativeCheckout.framework/Modules/module.modulemap:2:10: header 'PayPalNativeCheckout-Swift.h' not found

  header "PayPalNativeCheckout-Swift.h"
         ^
```

Gonna just re-run for now, but will make a ticket for us to address in another PR.

jaxdesmarais (approved):
This looks great! 👻

scannillo:
The[ latest CI run](https://github.com/paypal/iOS-SDK/actions/runs/3414687826/jobs/5697410225) has all tests passing, but the `swift package resolve` command, which we are using for our code coverage check, is failing to fetch NXO. 

UnitTests also pass locally. So, if 👍 with you all - I am going to merge this PR, and make a separate ticket [DTNOR-627](https://engineering.paypalcorp.com/jira/browse/DTNOR-627) for fixing CI.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Ah! Good question. Yeh for the timestamps, and also `NSLog statements appear in both the device’s console and debugger’s console whereas print only appears in the debugger console` ([stack overflow](https://stackoverflow.com/questions/25951195/swift-print-vs-println-vs-nslog#:~:text=NSLog%20adds%20a%20timestamp%20and,appears%20in%20the%20debugger%20console.)). I think Log works better for merchants who record production app logs.

PR #96, opened 2022-11-30
Remove unused correlationID from all API requests

scannillo:
### Reason for changes

- We are explicitly ignoring the `correlationID` returned on `APIClient.fetch` in every instance it is called. If we don't need it, let's get rid of it.

### Summary of changes

- Remove unused `correlationID`

### Questions

It looks like Android is using `correlationID` more heavily and returning it in merchant-facing errors ([example](https://github.com/paypal/Android-SDK/blob/e8f8aa29e07f1bf5600f8080c7d4046c69b825a0/Core/src/main/java/com/paypal/android/core/APIClientError.kt#L13)). We should align on if this is something we need.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/Card/CardClient.swift
+++ b/Sources/Card/CardClient.swift
@@ -48,7 +48,7 @@ public class CardClient {
         Task {
             do {
                 let confirmPaymentRequest = try ConfirmPaymentSourceRequest(accessToken: config.accessToken, cardRequest: request)
-                let (result, _) = try await apiClient.fetch(endpoint: confirmPaymentRequest)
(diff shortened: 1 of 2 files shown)


jaxdesmarais (approved):
Agree on aligning on how we want to use this value. Could be helpful for us to return it somewhere for troubleshooting/looking up errors in our logs potentially, but if it's not being used, we should nuke it for now until we decide how/if we want to use it.

sshropshire (approved):

scannillo:
I am going to merge this. I made a separate JIRA card for us to spend time deciding how to align the two platforms in regards to `correlationID` - DTNOR-647.

PR #97, opened 2022-11-30
Extract URLSession interaction into HTTP layer

scannillo:
### Reason for changes

- Higher separation of concerns. Currently the `APIClient` is a catch-all.
- Prepare us for a follow-up PR for adding a singleton `AnalyticsClient` layer. 
    - Otherwise, we'd have cyclic dependencies between `API <--> AnalyticsClient`. 
    - This change helps us avoid that, since `AnalyticsClient` will be able to have it's own instance of `HTTP` for networking.
- Alignment with Android. Allows us to isolate our actual `URLSession` interaction.

### Summary of changes

- Move `URLSession.performRequest` call and response handling into new `HTTP` class (took the name `HTTP` from Android, but we can think of something better too).
- Add docstrings to clarify:
    - `HTTP` is for interfacing directly with `URLSession`
    - `APIClient` is for payment method feature clients to execute requests & houses requests shared b/w modules

### Follow Up
- The next PR will be adding a singleton `AnalyticsClient` which will move `sessionID` and `AnalyticsEventData` logic off of `APIClient`.
- In a future PR, I want to remove the default implementation of `APIRequest` ([here](https://github.com/paypal/iOS-SDK/blob/093dd43175d873b5bee592a331e67478fd0c415f/Sources/PaymentsCore/Networking/Protocols/APIRequest.swift#L15-L48)).
    - That logic can move into `HTTP`.
    - `HTTP` can become the place for interacting with all of Apple's networking specific classes (ex: `URLRequest`, `URLResponse`)

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


sshropshire (approved):

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.coreConfig = coreConfig
+    }
+
+    /// For internal use for testing purpose
+    init(urlSession: URLSessionProtocol, coreConfig: CoreConfig) {
+        self.urlSession = urlSession
+        self.coreConfig = coreConfig
+    }
+    
+    func performRequest<T: APIRequest>(endpoint: T) async throws -> (T.ResponseType) {
```
sshropshire, line 15:
```suggestion
    func performRequest<T: APIRequest>(request: T) async throws -> (T.ResponseType) {
```
🧶 : Makes the parameter name match the underlying generic protocol.

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -0,0 +1,39 @@
...
+        guard let request = endpoint.toURLRequest(environment: coreConfig.environment) else {
+            throw APIClientError.invalidURLRequestError
+        }
+        let (data, response) = try await urlSession.performRequest(with: request)
```
jaxdesmarais: nit:

```suggestion
        }

        let (data, response) = try await urlSession.performRequest(with: request)
```

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -0,0 +1,39 @@
...
+        guard let response = response as? HTTPURLResponse else {
+            throw APIClientError.invalidURLResponseError
+        }
+
+        switch response.statusCode {
```
jaxdesmarais, line 25: Should we add a todo to handle the other status codes here? Or do we just want to handle 200-300 the same and everything else as a server response? I think eventually we'd want to handle 400 and 500 level errors differently then have a default for everything else. Thoughts?

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -0,0 +1,39 @@
...
+    private var urlSession: URLSessionProtocol
+    private let decoder = APIClientDecoder()
+
+    init(coreConfig: CoreConfig) {
+        self.urlSession = URLSession.shared
+        self.coreConfig = coreConfig
+    }
+
+    /// For internal use for testing purpose
+    init(urlSession: URLSessionProtocol, coreConfig: CoreConfig) {
```
jaxdesmarais: Instead of having 2 init's since both are internal could we do:
```
    init(coreConfig: CoreConfig, urlSession: URLSessionProtocol = URLSession.shared)
```

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: This is a good call. `APIClient.fetch(endpoint: )` is used pretty heavily throughout the app. I'll make a follow-up PR addressing that for both `APIClient` and `HTTP`, so not to muddy this PR.

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: Good catch!

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: Good question!

[It looks like on Android](https://github.com/paypal/Android-SDK/blob/e8f8aa29e07f1bf5600f8080c7d4046c69b825a0/Core/src/main/java/com/paypal/android/core/HttpResponse.kt#L16), we have the same check (200 to 300) is a success

It might be nice though to have more granular error handling per code, at least 400 v 500. I think this could be a follow-up PR / investigation to see if we think this would add value to our merchants / make our lives easier when troubleshooting. What do you think?

jaxdesmarais: Yeah, I think it makes sense to be a follow up/alignment on both platforms. Agree that it likely would make merchants and our troubleshooting easier.

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: @sshropshire - this feedback was addressed in a separate PR https://github.com/paypal/iOS-SDK/pull/99

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: Ticket added to the backlog - [DTNOR-648](https://engineering.paypalcorp.com/jira/browse/DTNOR-648)

jaxdesmarais (approved):
👻

PR #99, opened 2022-12-05
PR Feedback - rename endpoint to request param

scannillo:
### Reason for changes

- This PR is a result of PR feedback from[ this comment](https://github.com/paypal/iOS-SDK/pull/97#discussion_r1037573302).

### Summary of changes

- Rename `endpoint` param name to `request` to match param type.
    - `APIClient.fetch(endoint: APIRequest)` --> `APIClient.fetch(request: APIRequest)`

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


sshropshire (approved):

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -12,12 +12,12 @@ class HTTP {
         self.coreConfig = coreConfig
     }
     
-    func performRequest<T: APIRequest>(endpoint: T) async throws -> (T.ResponseType) {
-        guard let request = endpoint.toURLRequest(environment: coreConfig.environment) else {
+    func performRequest<T: APIRequest>(_ request: T) async throws -> (T.ResponseType) {
```
jaxdesmarais, line 15: Take it or leave it, but wondering if this would be more readable as:
```
let (_) = try await http.performRequest(with: analyticsEventRequest)
```

This is similar to what we do with the urlSession on line 20 below.

jaxdesmarais (approved):
👻

Sources/PaymentsCore/Networking/HTTP.swift
scannillo: I'm a little confused, which line are you referring to changing?

jaxdesmarais: Ah, yeah GitHub's highlighting is _less than ideal sometimes_. I was referencing this line:
```suggestion
    func performRequest<T: APIRequest>(with request: T) async throws -> (T.ResponseType) {
```

kgangineni (approved):

PR #100, opened 2022-12-05
Add singleton `AnalyticsService`

scannillo:
### Reason for changes

- Follow up to https://github.com/paypal/iOS-SDK/pull/91

### Summary of changes

- Add `AnalyticsService` class
    - This class is a singleton instance and manages the `sessionID`, constructing the appropriate `AnalyticsEventData`, and calling the `HTTP` class to perform the actual request.
    - In the future, this class will also handle the logic for batching multiple events into single requests.
    - This lets us move our Analytics specific logic outside of `APIClient` & `HTTP`.
- Add unit test for confirming singleton instance of `AnalyticsService` as well as single `sessionID` used.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -4,22 +4,42 @@ import Foundation
...
+        set { APIClient._analyticsService = newValue }
+    }
+    private static var _analyticsService: AnalyticsService?
+    
+    // MARK: - Public Initializer
 
     public init(coreConfig: CoreConfig) {
         self.http = HTTP(coreConfig: coreConfig)
         self.coreConfig = coreConfig
+        APIClient._analyticsService = AnalyticsService(http: http)
```
sshropshire: q: should this create a new `AnalyticsService` every time an `APIClient` is created?

Sources/PaymentsCore/Networking/APIClient.swift
```diff
@@ -4,22 +4,42 @@ import Foundation
...
+    
+    /// The `AnalyticsService` instance is static/shared so that only one sessionID is used.
+    /// The "singleton" has to be managed here because `AnalyticsService` has a dependency on `HTTP`.
+    ///
+    /// Exposed for testing.
+    weak var analyticsService: AnalyticsService? {
+        get { APIClient._analyticsService }
+        set { APIClient._analyticsService = newValue }
+    }
+    private static var _analyticsService: AnalyticsService?
```
jcnoriega: Why cant we have a method like `AnalyticsClient.getInstance(http)`? If an instance has already been created, the `getInstance` method sets the new `http` to the instance. 
It feels kinda weird to manage it here. If we need it in another class, we have to repeat this code

kgangineni: @scannillo is this the standard of doing singleton in swift?
since sessionID is now a one time set in `APIClient` we don't need to make `APIClient` to be singleton

PayPal.xcodeproj/project.pbxproj
```diff
@@ -230,6 +231,10 @@
 		803C31B6270E49160067D36E /* Card_Tests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Card_Tests.swift; sourceTree = "<group>"; };
 		803C31B8270E4C560067D36E /* MockURLSession.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockURLSession.swift; sourceTree = "<group>"; };
 		804E62812937EBCE004B9FEF /* HTTP.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = HTTP.swift; sourceTree = "<group>"; };
+<<<<<<< HEAD
```
jaxdesmarais: Looks like a few of these snuck in 🙈

Sources/PaymentsCore/Networking/APIClient.swift
jaxdesmarais:
```suggestion
    }

    private static var _analyticsService: AnalyticsService?
```

jaxdesmarais: Can we group all of the private vars together vs splitting them up?

Sources/PaymentsCore/Networking/AnalyticsService.swift
```diff
@@ -0,0 +1,28 @@
...
+    
+    // MARK: - Internal Methods
+
+    func sendEvent(_ name: String) async {
+        let eventData = AnalyticsEventData(eventName: name, sessionID: sessionID)
+
+        do {
+            let analyticsEventRequest = try AnalyticsEventRequest(eventData: eventData)
+            let (_) = try await http.performRequest(endpoint: analyticsEventRequest)
+        } catch let error {
```
jaxdesmarais: I believe catch provides a local error property automatically so we should just be able to do this without adding `let error` after `catch`:
```
        } catch {
        }
```

PayPal.xcodeproj/project.pbxproj
scannillo: Fixed!

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: I opted to group these together, like we're doing [on BT](https://github.com/braintree/braintree_ios/blob/c54c904b08c9a870fd64c87c7da51046a03f1c9a/Sources/BraintreeCore/BTAPIClient.swift#L44-L52). I think it's easier to understand if they're next to each together since the private static var **only** exists to manage the singleton in the `weak` property above.

That's also why I didn't new-line in between them ([your previous comment](https://github.com/paypal/iOS-SDK/pull/100/files#r1042672815)). I thought it felt more intentional/clear that these two properties only exist to be used together.

But let me know what you think and if there's any way to increase readability!

scannillo: See comment https://github.com/paypal/iOS-SDK/pull/100/files#r1043496137

kgangineni (approved):
LGTM

Sources/PaymentsCore/Networking/APIClient.swift
jaxdesmarais: I think that makes sense for the spacing if we feel like it will cut down on confusion, we have a space on BT currently, but I could see how that could lead to confusion potentially. 

With the grouping, it was more around right now this file goes private > internal > private whereas I think our ideal would be public > internal > private. I think we could move `analyticsService` and `_analyticsService` above the other private vars and still keep these two grouped together to increase readability. Not a big deal either way we go, though!

scannillo: > Why cant we have a method like AnalyticsClient.getInstance(http)?

This is what I was hoping to do, initially too. But, this can't be done in Swift using the [typical singleton implementation](https://cocoacasts.com/what-is-a-singleton-and-how-to-create-one-in-swift](https://www.hackingwithswift.com/example-code/language/what-is-a-singleton)), because of the requirement for a private init. Therefore, you can't have a singleton with dependency injection.

> It feels kinda weird to manage it here. If we need it in another class, we have to repeat this code.

Yeh, I agree it's weird to have to manage the singleton this way. I know Jax and Matt did a lot of work to think of a cleaner way of doing this on BT too ([see thread](https://github.com/braintree/braintree_ios/pull/887#discussion_r1010550537)), but landed on this being the most straightforward. I'll keep messing around with ideas for other solutions.

Though, we shouldn't need to use this in any other place. The intention / design is to only have a single point of entry, via `APIClient.sendEvent`.

jcnoriega: @scannillo my java/kotlin mind is bugging me, but what about something like this:
```swift
class AnalyticsService {

    
    private static let instance = AnalyticsService()
    // MARK: - Internal Properties

    private var http: HTTP?
    let sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")

    // MARK: - Initializer

    private init() {
    }
    
    public static func getInstance(http: HTTP) -> AnalyticsService {
        instance.http = http
        return instance
    }
    
    // MARK: - Internal Methods

    func sendEvent(_ name: String) async {
        let eventData = AnalyticsEventData(eventName: name, sessionID: sessionID)

        do {
            let analyticsEventRequest = try AnalyticsEventRequest(eventData: eventData)
            let (_) = try await http?.performRequest(endpoint: analyticsEventRequest)
        } catch {
        }
    }
}
```

scannillo: > since sessionID is now a one time set in APIClient we don't need to make APIClient to be singleton

This is a good callout. You're right - technically right now, we could solve the single sessionID problem by just using a regular class with a static property for `sessionID`. The singleton setup is preparing us for when we need to take in multiple events from different feature clients, and then combine & batch them into single requests. We could 🏈 punt on it, or do it now.

scannillo: @jcnoriega - this is a cool idea! I'll give this a spin in the PR.

mattwylder: Good discussion here! From my point of view, `AnalyticsService` isn't a "true" singleton and isn't the standard method for writing a singleton in Swift. @jcnoriega your code is a great option to refactor it into a "standard" singleton that manages itself. But I think the issue is that the current setup only has _features_ a singleton: there is only one, but it does not maintain its own shared instance. This was done so we could mock out `AnalyticsService` in tests.

The existing solution isn't the greatest in the world, so a discussion on improving it is great. But I think the "standard" singleton solution that Jose suggested recreates the problems of the current solution but makes it a little less readable (see `init` in the code example below). The `getInstance(http:)` method in Jose's code suggestion is, in my opinion, not very Swifty, and obfuscates how `_analyticsService` is initialized and maintained. The existing code is confusing, and in my opinion the lesser evil is to keep all of the confusing logic in one file vs having some of the logic exist in `AnalyticsService.getInstance(http:)` and some in `APIClient.init(coreConfig:)`.

We originally set this up for dependency injection into `APIClient` for testing. If we were to call `AnalyticsService.instance.sendEvent("eventName")` in every file, that would work well for running the app, but it would couple our tests to real networking logic. Even with a standard singleton solution we would still need the `weak var` with the `get` and `set` in order to mock out analytics.

All that said, with a standard singleton solution I think we would wind up with code that looks very similar to what we have now: 

```swift
    weak var analyticsService: AnalyticsService? {
        get { APIClient._analyticsService }
        set { APIClient._analyticsService = newValue }
    }
    private static var _analyticsService: AnalyticsService?
    
    // MARK: - Public Initializer

    public init(coreConfig: CoreConfig) {
        self.http = HTTP(coreConfig: coreConfig)
        self.coreConfig = coreConfig
        APIClient._analyticsService = AnalyticsService.getInstance(http)
    }
```

Sources/PaymentsCore/Networking/AnalyticsService.swift
```diff
@@ -0,0 +1,28 @@
+import Foundation
+
+class AnalyticsService {
+
+    // MARK: - Internal Properties
+
+    private let http: HTTP
+    let sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
```
mattwylder: I notice there is a sessionID in both `AnalyticsService` and `APIClient`. Should we remove one? Do we want to pass one into `AnalyticsService` from `APIClient`? Or is there some other option?

scannillo: This is an unused variable, forgot to delete. It should live in the AnalyticsService only, now! Removed.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Ah gotcha - I pushed up a change, let me know what you think!

scannillo: Also to clarify, my plan was that the feature clients won't interact with `AnalyticsService` at all, but instead just call `APIClient.sendEvent` ([code](https://github.com/paypal/iOS-SDK/blob/8f8354122809ce9c2eea9e8552b14e48c3fc20a3/Sources/PaymentsCore/Networking/APIClient.swift#L55)).

Also in both the current and Jose's suggested implementation, we'll be able to inject a mock `HTTP` into `AnalyticsService` so we can mock the actual network request happening.

scannillo: > The getInstance(http:) method in Jose's code suggestion is, in my opinion, not very Swifty, and obfuscates how _analyticsService is initialized and maintained.

I agree this method/style isn't Swifty. But there's also just been a lot of general confusion regarding the `_propertyName` syntax with an internal private var like the current implementation. I think Jose's suggestion ends up providing the same solution and is a little easier to wrap your head around.

scannillo: 👋 Hello everyone.

I thought through this PR some more and trialed several things. A few changes:

* After talking with Steven - I think it makes more sense to go with the [original plan](https://github.com/paypal/Android-SDK/pull/104#discussion_r1033756732) of having `1 sessionID: 1 accessToken`, instead of `1 sessionID: entire lifecycle of the app`. 
    * A customer might _never_ close the Uber app, for example, let's say for weeks or months on end. This will make our sessionID basically useless.
    * If we tie to an accessToken, we're now in sync with whatever expiration & refresh preferences the merchant has set on their token.
* I opted for Jose's style of approach to the singleton pattern with added comments and docstrings for clarity.
   * I think it's ultimately just easier to understand and read vs using the `_` private/static wrapper property pattern. It also enforces a true singleton pattern, where the `_` pattern technically does not.
    * **NOTE**: The feature clients _will not_ interact with `AnalyticsService`, so there should not be confusion throughout the code regarding `AnalyticsService.sharedInstance()`. But even so, I added a docstring.
    * Feature clients will call `APIClient.sendEvent()` to fire analytics.
    
Let me know what you think. Happy to move this discussion to stand-up if folks still have thoughts.

Sources/PaymentsCore/Networking/APIClient.swift
scannillo: Technically this line is _not_ creating a new instance every time since it is assigning to the internal (`_`) static property which we have the custom wrapper on.

But this is super hard to decipher IMO. I opted not to use this pattern in the re-write of this PR. Changes pushed.

scannillo:
👋 Hello everyone.

I thought through this PR some more and trialed several things. A few changes:

* After talking with Steven - I think it makes more sense to go with the [original plan](https://github.com/paypal/Android-SDK/pull/104#discussion_r1033756732) of having `1 sessionID: 1 accessToken`, instead of `1 sessionID: entire lifecycle of the app`. 
    * A customer might _never_ close the Uber app, for example, let's say for weeks or months on end. This will make our sessionID basically useless.
    * If we tie to an accessToken, we're now in sync with whatever expiration & refresh preferences the merchant has set on their token.
* I opted for Jose's style of approach to the singleton pattern with added comments and docstrings for clarity.
   * I think it's ultimately just easier to understand and read vs using the `_` private/static wrapper property pattern. It also enforces a true singleton pattern, where the `_` pattern technically does not.
    * **NOTE**: The feature clients _will not_ interact with `AnalyticsService`, so there should not be confusion throughout the code regarding `AnalyticsService.sharedInstance()`. But even so, I added a docstring.
    * Feature clients will call `APIClient.sendEvent()` to fire analytics.
    
Let me know what you think. Happy to move this discussion to stand-up if folks still have thoughts.

Sources/PaymentsCore/Networking/AnalyticsService.swift
```diff
@@ -0,0 +1,45 @@
...
+    private var sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
+    
+    private var http: HTTP? {
+        // This property observer generates a new `sessionID` if the `accessToken`
+        // injected into the `AnalyticsSingleton` ever changes.
+        willSet {
+            if newValue?.coreConfig.accessToken != http?.coreConfig.accessToken {
+                sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
+            }
+        }
```
jcnoriega, line 19: 🔥

jcnoriega (approved):

sshropshire (approved):

mattwylder (approved):

PR #103, opened 2022-12-15
simplify code for getting presentation anchor

kgangineni:
### Reason for changes

1. Simplify code snippet for retrieving presentation anchor
2. couldn't fix runtime warning of running on background thread

### Summary of changes

- 

### Checklist

- [ ] Added a changelog entry

- @kgangineni @scannillo

PR #104, opened 2023-01-05
Remove merchant `ASWebAuthenticationPresentationContextProviding` step

scannillo:
### Reason for changes

We can clean-up our docs story by removing the need for merchants to conform to `ASWebAuthenticationPresentationContextProviding`, and instead place this burden on the PayPal SDK.

### Summary of changes

- Remove `ASWebAuthenticationPresentationContextProviding` merchant step from Card & PayPal web docs
- Refactor internal use of `WebAuthenticationSession` in `CardClient` and `PayPalWebCheckoutClient`
   - Make our testing/mocking strategy more in-line with the rest of the iOS SDK by injecting via internal initializer
   - Existing pattern is more in-line with Android testing setup
- Conform `CardClient` and `PayPalWebCheckoutClient` to `ASWebAuthenticationPresentationContextProviding`

### Checklist

- [X] Added a changelog entry
- [X] [Follow-up PR ](https://github.com/paypal/iOS-SDK/pull/105)to cleanup Demo

### Authors
@scannillo

### DIFF
--- a/Sources/Card/CardClient.swift
+++ b/Sources/Card/CardClient.swift
@@ -35,13 +35,6 @@ public class CardClient: NSObject {
     /// - Returns: Card result
     /// - Throws: PayPalSDK error if approve order could not complete successfully
     public func approveOrder(request: CardRequest) {
-        start(request: request)
-    }
-
-    /// Internal function for testing
-    func start(
-        request: CardRequest
-    ) {
         Task {
             do {
                 let confirmPaymentRequest = try ConfirmPaymentSourceRequest(accessToken: config.accessToken, cardRequest: request)
--- a/Sources/Card/CardClient.swift
+++ b/Sources/Card/CardClient.swift
@@ -4,20 +4,7 @@ import AuthenticationServices
 import PaymentsCore
 #endif
 
-public class CardClient: NSObject, ASWebAuthenticationPresentationContextProviding {
-    
-    // MARK: - ASWebAuthenticationPresentationContextProviding conformance
-
-    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
-        if #available(iOS 15, *) {
-            let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
-            let window = firstScene?.windows.first { $0.isKeyWindow }
-            return window ?? ASPresentationAnchor()
-        } else {
-            let window = UIApplication.shared.windows.first { $0.isKeyWindow }
-            return window ?? ASPresentationAnchor()
-        }
-    }
+public class CardClient: NSObject {
 
     public weak var delegate: CardDelegate?
 
@@ -129,5 +116,21 @@ public class CardClient: NSObject, ASWebAuthenticationPresentationContextProvidi
 
     private func notifyCancellation() {
         delegate?.cardDidCancel(self)
+    }
+}
+
+// MARK: - ASWebAuthenticationPresentationContextProviding conformance
+
+extension CardClient: ASWebAuthenticationPresentationContextProviding {
+
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        if #available(iOS 15, *) {
+            let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
+            let window = firstScene?.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        } else {
+            let window = UIApplication.shared.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        }
     }
 }
--- a/Sources/Card/CardClient.swift
+++ b/Sources/Card/CardClient.swift
@@ -4,24 +4,40 @@ import AuthenticationServices
 import PaymentsCore
 #endif
 
-public class CardClient {
+public class CardClient: NSObject, ASWebAuthenticationPresentationContextProviding {
+    
+    // MARK: - ASWebAuthenticationPresentationContextProviding conformance
+
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        if #available(iOS 15, *) {
+            let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
+            let window = firstScene?.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        } else {
+            let window = UIApplication.shared.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        }
+    }
 
     public weak var delegate: CardDelegate?
 
     private let apiClient: APIClient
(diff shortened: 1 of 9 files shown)


Sources/PaymentsCore/WebAuthenticationSession.swift
```diff
@@ -17,6 +17,8 @@ public class WebAuthenticationSession: NSObject {
         authenticationSession.prefersEphemeralWebBrowserSession = true
         authenticationSession.presentationContextProvider = context
 
-        authenticationSession.start()
+        DispatchQueue.main.async {
+            authenticationSession.start()
+        }
```
scannillo, line 22: This fixes the threading warning discussed [here](https://github.com/paypal/iOS-SDK/pull/103#issuecomment-1355025133)

Sources/Card/CardClient.swift
```diff
@@ -123,3 +111,19 @@ public class CardClient {
         delegate?.cardDidCancel(self)
     }
 }
+
+// MARK: - ASWebAuthenticationPresentationContextProviding conformance
+
+extension CardClient: ASWebAuthenticationPresentationContextProviding {
+
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
```
jaxdesmarais, line 119: Should we provide merchants with a way to pass their own window in the top level class if they want? I think using the first window _should_ cover most merchants but but may not always be the case. In the clients we could have something like we do on BT `public var activeWindow: UIWindow?` to allow merchants to pass in their own optional `UIWindow` to use instead. Then in here we can check if it's nil and if not return the merchant specified window. Thoughts?

scannillo: Great question! 

So I actually made a [ticket](https://paypal.atlassian.net/browse/DTNOR-749) and a TODO to add that functionality after this PR. But after thinking about it more I ended up closing it and leaving a note why it was a "won't do".

SPLUNKing BT logs shows that 98% of all events are on iPhones vs iPads (multitasking is only supporting on iPads). So seeing as we don't get much traffic at all there combined with the fact that I think supporting multi-tasking for a checkout flow would be pretty rare for a merchant, I figured we could leave this as "won't do", until/if a merchant asks us to explicitly support it. 

On the BT side, we added it in for completeness, not based on merchant request. I wonder if we removed it, if anyone would notice 🤷 

What do you think?

jaxdesmarais: Oh dang - that makes a ton of sense, that's for sharing your findings on that! I had assumed it was added because merchants had asked for it previously. I agree in that case that we should not add it for now to clean up the integration path overall and evaluate if merchants ask for explicit support! 🚀

scannillo: Ok sweet. Another idea - we _could_ remove it from the BT V6 beta and see if any merchants make noise about it.

jaxdesmarais: I dig it! Yeah that would be the real test to see if merchants are actually using it. I think we should try to remove it and see if we get any feedback on it. Based on the data you found, this likely would allow us to really clean up the integration pattern with hopefully no merchant issues.

jaxdesmarais (approved):
👻

jcnoriega (approved):
qq: why do `PayPalClient` and `CardClient` need to extend an `NSObject`?

jcnoriega:
also, does this PR invalidate #103 ?

Sources/PayPalWebCheckout/PayPalWebCheckoutClient.swift
```diff
@@ -101,3 +96,19 @@ public class PayPalWebCheckoutClient {
...
+    
+    public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        if #available(iOS 15, *) {
+            let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
+            let window = firstScene?.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        } else {
+            let window = UIApplication.shared.windows.first { $0.isKeyWindow }
+            return window ?? ASPresentationAnchor()
+        }
```
mattwylder, line 112: Could we extract this into a default implementation somewhere to avoid code duplication?

mattwylder (approved):

scannillo:
> also, does this PR invalidate https://github.com/paypal/iOS-SDK/pull/103 ?

Yes, it does! That PR will be closed once these are merged.

Sources/Card/CardClient.swift
mattwylder: Would that mean removing these [lines of code](https://github.com/braintree/braintree_ios/blob/a447ec02ee5f52940f2ab9c53fdd3dbf0f645c31/Sources/BraintreePayPal/BTPayPalClient.swift#L218-L228) and removing `activeWindow` from `BTPayPalRequest`?

scannillo:
> qq: why do PayPalClient and CardClient need to extend an NSObject?

Good question. According to [Apple's docs](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationpresentationcontextproviding?language=objc), `ASWebAuthenticationPresentationContextProviding` inherits from `NSObject`. Without it, we get this build error:

Screen Shot 2023-01-10 at 9 53 22 AM

Another kind of annoying thing, you can't do inheritance in an extension, so we have to have the `: NSObject` inheritance at the actual `CardClient` class definition spot.

Sources/PayPalWebCheckout/PayPalWebCheckoutClient.swift
scannillo: Great callout! 

So I can't do a protocol extension on `ASWebAuthenticationPresentationContextProviding` because it inherits from `NSObject` ([see Apple docs](https://developer.apple.com/documentation/authenticationservices/aswebauthenticationpresentationcontextproviding?language=objc)). Also, even if we could do it, I'm not sure we'd want to because then merchants using `ASWebAuthenticationPresentationContextProviding` in other ways in their app would be stuck with whatever default implementation we need to give it for our SDK.

In terms of trying to re-use (instead of duplicate) this code, I think we'd have to do inheritance or move the logic inside of the `presentationAnchor` into some global/shared func across modules. Since `CardClient` and `PayPalWebCheckoutClient` are merchant integration points, the parent class would need to be public and may cause more confusion to the merchant 🤔.

We could experiment with the later option though. Maybe this can be a new ticket / punt 🏈 ? Also, the `PaymentFlow` module on BT exists to solve this kind of problem. Not needing to redo the logic of switching to a WebView per payment method. Not sure if that is really worthwhile here, but just calling it out.

If you have other ideas please let me know too!

Sources/Card/CardClient.swift
scannillo: Just this [part](https://github.com/braintree/braintree_ios/blob/a447ec02ee5f52940f2ab9c53fdd3dbf0f645c31/Sources/BraintreePayPal/BTPayPalClient.swift#L218-L220), but yes

PR #106, opened 2023-01-10
Add `Bundle.module` to render assets in `PayPalUI` using SPM

jcnoriega:
### Reason for changes
Using `PayPalUI` in a project with SPM, wasnt rendering any assets.

### Summary of changes

- We ask for `Bundle.module` if we are in SPM.
- [Jira ticket](https://paypal.atlassian.net/browse/DTNOR-759)

### Checklist

- [ ] Added a changelog entry

- @jcnoriega 
- @scannillo


Sources/PayPalUI/PaymentButton.swift
```diff
@@ -4,7 +4,12 @@ import UIKit
 public class PaymentButton: UIButton {
 
     // asset identifier path for image and color button assets
+
```
scannillo:
```suggestion
```

scannillo: Nitpick!

scannillo (changes requested):
Changes look good to me, but I think this needs a CHANGELOG entry to call out that we fixed this SPM bug.

scannillo (approved):
🎉 Great find to you and Leon!

Sources/PayPalUI/PaymentButton.swift
```diff
@@ -4,7 +4,11 @@ import UIKit
 public class PaymentButton: UIButton {
 
     // asset identifier path for image and color button assets
+    #if SWIFT_PACKAGE
```
jaxdesmarais, line 7: Just a general question but can we link in the PR description the issue and where the solution was found? More-so just curious and would be great to have it called out somewhere incase we run across a similar thing in the future!

jaxdesmarais (approved):
👻

Sources/PayPalUI/PaymentButton.swift
jcnoriega: @jaxdesmarais i need to create a ticket for it! 
GPS folks wanted to test the sdk, so they simply imported it with SPM and told me that the buttons werent working (you could only see a yellow square for example). 
Adding `Boundle.module` was found in stackoverflow and the apple docs
The idea of puting `#if` was @scannillo !
i'll put more details in the ticket

jaxdesmarais: Awesome! Thanks @jcnoriega!

jcnoriega: here is the [ticket](https://paypal.atlassian.net/browse/DTNOR-759)!

PR #108, opened 2023-01-23
[iOS] PayPalUI: fix label not showing

jcnoriega:
### Reason for changes

GPS reported that the payment buttons dont show any label even when set.
Ticket: [DTNOR-767](https://paypal.atlassian.net/browse/DTNOR-767)

### Summary of changes

- There was an issue with the wordings in the label.
- Corrected the Demo app as well

| Before      | After |
| ----------- | ----------- |
|Simulator Screen Shot - iPhone 14 Pro - 2023-01-23 at 10 09 48 |Screen Shot 2023-01-23 at 09 57 53Screen Shot 2023-01-23 at 09 58 04|

### Checklist

- [x] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -94,3 +94,19 @@ extension PaymentButtonFundingSource {
         PaymentButtonFundingSource.allCases().map { $0.rawValue }
     }
 }
+
+extension PayPalButton.Label {
+    
+    static func allCases() -> [PayPalButton.Label] {
+        [
+            PayPalButton.Label.none,
+            PayPalButton.Label.checkout,
+            PayPalButton.Label.buyNow,
+            PayPalButton.Label.payWith
+        ]
+    }
+    
+    static func allCasesAsString() -> [String] {
+        allCases().map { $0.rawValue }
+    }
+}
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -2,111 +2,77 @@ import PayPalUI
 
 extension PayPalPayLaterButton.Color {
 
-    static func allCases() -> [PayPalPayLaterButton.Color] {
-        [
-            PayPalPayLaterButton.Color.gold,
-            PayPalPayLaterButton.Color.white,
-            PayPalPayLaterButton.Color.black,
-            PayPalPayLaterButton.Color.silver,
-            PayPalPayLaterButton.Color.blue
-        ]
+    public static var allCases: [PayPalPayLaterButton.Color] {
+        [.gold, .white, .black, .silver, .blue]
     }
 
     static func allCasesAsString() -> [String] {
-        allCases().map { $0.rawValue }
+        Self.allCases.map { $0.rawValue }
     }
 }
 
-extension PayPalButton.Color {
+extension PayPalButton.Color: CaseIterable {
 
-    static func allCases() -> [PayPalButton.Color] {
-        [
-            PayPalButton.Color.gold,
-            PayPalButton.Color.white,
-            PayPalButton.Color.black,
-            PayPalButton.Color.silver,
-            PayPalButton.Color.blue
-        ]
+    public static var allCases: [PayPalButton.Color] {
+        [.gold, .white, .black, .silver, .blue]
     }
 
     static func allCasesAsString() -> [String] {
-        allCases().map { $0.rawValue }
+        Self.allCases.map { $0.rawValue }
     }
 }
 
-extension PayPalCreditButton.Color {
+extension PayPalCreditButton.Color: CaseIterable {
 
-    static func allCases() -> [PayPalCreditButton.Color] {
-        [
-            PayPalCreditButton.Color.white,
-            PayPalCreditButton.Color.black,
-            PayPalCreditButton.Color.darkBlue
-        ]
+    public static var allCases: [PayPalCreditButton.Color] {
+        [.white, .black, .darkBlue]
     }
 
     static func allCasesAsString() -> [String] {
-        allCases().map { $0.rawValue }
+        Self.allCases.map { $0.rawValue }
     }
 }
 
-extension PaymentButtonEdges {
+extension PaymentButtonEdges: CaseIterable {
 
-    static func allCases() -> [PaymentButtonEdges] {
-        [
-            PaymentButtonEdges.hardEdges,
-            PaymentButtonEdges.softEdges,
-            PaymentButtonEdges.rounded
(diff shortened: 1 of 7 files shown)


sshropshire (approved):

Sources/PayPalUI/PayPalButton.swift
```diff
@@ -19,18 +19,34 @@ public final class PayPalButton: PaymentButton {
...
         }
+        
+        public static func allCases() -> [Label] {
+            [
+                Label.none,
+                Label.checkout,
+                Label.buyNow,
+                Label.payWith
+            ]
+        }
```
sshropshire: 🧶 : There is a standard library feature that does this, [CaseIterable](https://stackoverflow.com/a/32953212).

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -125,6 +133,12 @@ class PaymentButtonCustomizationViewController: UIViewController {
         if sender == fundingPicker {
             reloadColorPicker()
         }
+        let fundingSource = PaymentButtonFundingSource.allCases()[fundingPicker.selectedSegmentIndex]
+        if fundingSource == .payPal {
+            labelPicker.isHidden = false
+        } else {
+            labelPicker.isHidden = true
+        }
```
jaxdesmarais:
```suggestion
labelPicker.isHidden = fundingSource != .payPal
```

jaxdesmarais: +1 - this enum should be able to be updated to:
```
    public enum Label: String, CaseIterable {
        /// Display no label
        case none
        
        /// Display "Checkout" on the right side of the button's logo
        case checkout = "Checkout"

        /// Display "Buy now" on the right side of the button's logo
        case buyNow = "Buy Now"

        /// Display "Pay with" on the left side of the button's logo
        case payWith = "Pay with"

        var label: PaymentButtonLabel? {
            PaymentButtonLabel(rawValue: rawValue)
        }
        
        public static func allCasesAsString() -> [String] {
            Self.allCases.map { $0.rawValue }
        }
    }
```

jcnoriega: @jaxdesmarais i just realized this shouldnt be here. We had the same for the other enums of the buttons but we removed. This should be in an extension on the Demo app

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -94,3 +94,19 @@ extension PaymentButtonFundingSource {
...
+            PayPalButton.Label.checkout,
+            PayPalButton.Label.buyNow,
+            PayPalButton.Label.payWith
+        ]
+    }
+    
+    static func allCasesAsString() -> [String] {
+        allCases().map { $0.rawValue }
+    }
+}
```
jaxdesmarais: If we wanted, we could still use `CaseIterable` here. Looks slightly different in an extension but still pretty clean imo:
```
extension PayPalButton.Label: CaseIterable {

    public static var allCases: [PayPalButton.Label] {
        [.none, .checkout, .buyNow, .payWith]
    }

    static func allCasesAsString() -> [String] {
        Self.allCases.map { $0.rawValue }
    }
}
```

jcnoriega: @jaxdesmarais oh i tried that but i got a compilation error because the extensions where not on the same file, so i couldnt simply use `allCases`. I'll try this

Demo/Demo/FeatureViewControllers/PaymentButtonCustomizationViewController.swift
```diff
@@ -125,6 +133,9 @@ class PaymentButtonCustomizationViewController: UIViewController {
         if sender == fundingPicker {
             reloadColorPicker()
         }
+        let fundingSource = PaymentButtonFundingSource.allCases[fundingPicker.selectedSegmentIndex]
+        labelPicker.isHidden = fundingSource != .payPal
```
jaxdesmarais: It looks like certain labels are only available with certain sizes. Is this expected (I assume so just want to confirm)? If so, we can probably add something like this under `fundingSource` here to only display the labels when they are applicable:
```
let sizeSource = PaymentButtonSize.allCases[sizePicker.selectedSegmentIndex]
labelPicker.isHidden = fundingSource != .payPal || sizeSource == .mini || sizeSource == .collapsed
```
Can likely do the same in the SwiftUI Demo. Also I noticed that all of the labels apply as expected to the `expanded` and `full` size buttons _except_  `payWith` on the `expanded` button. Is this intentional?

jcnoriega: @jaxdesmarais it is. This is what we copied from NXO. Im guessing it has to do with the logo, because the small ones only have the PayPal logo with the `P`, so it doesnt make sense to add a label, it would look something like `P` `Checkout`, its not readable. 
Same with `payWith`. This reasons are all speculative though, because we never had requirements.

jaxdesmarais: Yeah, I understand why they don't display on the smaller buttons, we should probably call that out somewhere so we don't get inbounds about it.

What I was not sure if is still why `payWith` doesn't display on the expanded button. All of the other labels display as expected on `expanded` and `full` so it seems odd to me that one does not. Again may be expected but worth confirming with NXO if this is a bug or not.

jcnoriega: @jaxdesmarais code wise it is expected. They also did a major refactor of the buttons not so long ago, i havent tested it fully, so we might have diverged from them now. I'll see if i can add it to `expanded` for consistencies sake

jcnoriega: @jaxdesmarais i updated the pickers on both demos, and also added `payLater` to the `expanded size as well. Also added it to the docs

docs/PayPalUI/README.md
```diff
@@ -43,6 +43,7 @@ Each button as a `UKit` and `SwiftUI` implementation as follows:
     | PayPalButton      | PayPalButton.Representable       |
     | PayPalCreditButton   | PayPalCreditButton.Representable        |
     | PayPalPayLaterButton   | PayPalPayLaterButton.Representable        |
+> Note: label customization only applies to `PayPalButton` when its size is `.expanded` or `.full`
```
jaxdesmarais, line 46: ❤️

jaxdesmarais (approved):
Thanks for addressing comments and making those updates! This looks great. 🚀

PR #111, opened 2023-01-26
Rename `PayPalWebCheckout` to `PayPalWebPayments`

jcnoriega:
### Reason for changes
We have to rename to `PayPalWebPayments` because it better reflects the functionality of the module

### Summary of changes

- Rename Target and folder
- Updated podspec
- Rename Unit test folder as well
- Updated `Package.swift`
- Update docs

### Checklist

- [x] Added a changelog entry

- @jcnoriega


.github/workflows/build.yml
```diff
@@ -14,18 +14,18 @@ jobs:
...
+#      - name: Check out repository
+#        uses: actions/checkout@v2
+#      - name: Use Xcode 14
+#        run: sudo xcode-select -switch /Applications/Xcode_14.0.1.app
+#      - name: Use current branch
+#        run: sed -i '' 's/branch = .*/branch = \"'"$GITHUB_HEAD_REF"'\";/' SampleApps/SPMTest/SPMTest.xcodeproj/project.pbxproj
+#      - name: Run swift package resolve
+#        run: cd SampleApps/SPMTest && swift package resolve
+#      - name: Build & archive SPMTest
+#        run: xcodebuild -project 'SampleApps/SPMTest/SPMTest.xcodeproj' -scheme 'SPMTest' clean build archive CODE_SIGNING_ALLOWED=NO
```
mattwylder, line 30: Are we planning on uncommenting this?

mattwylder (approved):
lgtm, but curious about that SPM build step.

.github/workflows/build.yml
jcnoriega: yes, this build was breaking on every module change and i had do some weird merges all the time. It just a test to check that SPM works. I tested every module change manually using SPM, and they work. So i decided to comment it for now, and once all the changes are merged, adding it and updating it

PR #114, opened 2023-01-30
Add HTTP_Tests.swift & stubs for APIClient.fetch()

scannillo:
### Reason for changes

- By extracting URLSession interactions into the `HTTP.swift` layer (PR - #97), we can simplify the testing setup for each feature client.
- Now, feature clients don't need to mock `URLSession` directly.

### Summary of changes

- Add `HTTP_Tests.swift`
    - Move tests from `APIClient_Tests.swift` over that are actually the responsibility of `HTTP`.
- Add functionality to `MockAPIClient` to stub `fetch()` method return values.
- Cleanup `CardClient_Tests.swift`
   - Remove `MockURLSession` references
   - Instead, stub APIClient `fetch()` responses
   
### Future improvements
- Remove `URLSession` dependency in `APIClient`, since only needed at `HTTP` layer
    - Add test that `APIClient` forwards requests properly to `HTTP` layer
- Refactor GraphQL to use same underlying `HTTP` layer as REST requests
- Refactor EligibilityAPI tests not to rely on `MockURLSession` directly

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -34,227 +32,179 @@ class CardClient_Tests: XCTestCase {
...
     }
 
     func testApproveOrder_whenApiCallFails_returnsError() {
-        if let jsonResponse = """
+        mockAPIClient.cannedJSONResponse = """
         {
             "some_unexpected_response": "something"
         }
-        """.data(using: String.Encoding.utf8) {
-            let mockResponse = MockResponse.success(
```
scannillo: Sadly the diff is hideous, but these tests are mostly the same, just indented to the left, since we're force unwrapping the `data(using: String.Encoding.utf8)` now in `MockAPIClient`.

UnitTests/PaymentsCoreTests/HTTP_Tests.swift
```diff
@@ -0,0 +1,139 @@
...
+        
+        sut = HTTP(
+            urlSession: mockURLSession,
+            coreConfig: config
+        )
+    }
+    
+    // MARK: - performRequest()
+
+    func testPerformRequest_withNoURLRequest_returnsInvalidURLRequestError() async {
```
scannillo, line 35: These tests were moved as-is from `APIClient_Tests.swift`

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -34,227 +32,179 @@ class CardClient_Tests: XCTestCase {
...
-                XCTAssertEqual(result.paymentSource?.card.brand, "VISA")
-                XCTAssertEqual(result.paymentSource?.card.lastFourDigits, "7321")
-                XCTAssertEqual(result.paymentSource?.card.type, "CREDIT")
-                expectation.fulfill()
-            }, error: { _, _ -> Void in
-                XCTFail()
-            }, threeDSWillLaunch: { _ -> Void in
-                XCTFail()
-            })
+        let mockCardDelegate = MockCardDelegate(success: {_, result -> Void in
```
jaxdesmarais: nit: we should be able to remove `-> Void` in all of these test closures

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -34,227 +32,179 @@ class CardClient_Tests: XCTestCase {
...
+            XCTAssertEqual(result.orderID, "testOrderId")
+            XCTAssertEqual(result.paymentSource?.card.brand, "VISA")
+            XCTAssertEqual(result.paymentSource?.card.lastFourDigits, "7321")
+            XCTAssertEqual(result.paymentSource?.card.type, "CREDIT")
+            expectation.fulfill()
+        }, error: { _, _ -> Void in
+            XCTFail()
```
jaxdesmarais: Is it worth adding a message to these `XCTFail`s? It could make it easier to debug which we are hitting if we break something.

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -34,227 +32,179 @@ class CardClient_Tests: XCTestCase {
...
+        waitForExpectations(timeout: 10)
     }
 
     func testApproveOrder_withThreeDSecure_userCancelsBrowser() {
-        if let confirmPaymentSourceJsonResponse = CardResponses.confirmPaymentSourceJsonWith3DS.rawValue.data(using: String.Encoding.utf8) {
-            let mockConfirmResponse = MockResponse.success(
-                MockResponse.Success(data: confirmPaymentSourceJsonResponse, urlResponse: successURLResponse)
+        mockAPIClient.cannedJSONResponse = CardResponses.confirmPaymentSourceJsonWith3DS.rawValue
+
+        mockWebAuthSession.cannedErrorResponse = ASWebAuthenticationSessionError(
```
jaxdesmarais: nit, we can do this instead I believe:
```
mockWebAuthSession.cannedErrorResponse = ASWebAuthenticationSessionError(
    .canceledLogin, 
    userInfo: ["Description": "Mock cancellation error description."]
)
```

UnitTests/CardTests/CardClient_Tests.swift
```diff
@@ -34,227 +32,179 @@ class CardClient_Tests: XCTestCase {
...
-
-            cardClient.approveOrder(request: cardRequest)
-
-            waitForExpectations(timeout: 10)
-        } else {
-            XCTFail("Data json cannot be null")
-        }
+        let mockCardDelegate = MockCardDelegate(
+            success: {_, _ -> Void in
+                XCTFail("Flow should not succed")
```
jaxdesmarais:
```suggestion
                XCTFail("Flow should not succeed")
```

UnitTests/PaymentsCoreTests/HTTP_Tests.swift
```diff
@@ -0,0 +1,139 @@
...
+        
+        mockURLSession = MockURLSession()
+        mockURLSession.cannedError = nil
+        mockURLSession.cannedURLResponse = nil
+        mockURLSession.cannedJSONData = nil
+        
+        sut = HTTP(
+            urlSession: mockURLSession,
+            coreConfig: config
+        )
```
jaxdesmarais: I think this was just moved over, but I believe we can single line this and have it be under our liter rule. I also wonder if it'd be clearer if we changed this property name to `mockHTTP` instead of `sut` to match what we use elsewhere in tests. Thoughts?

UnitTests/PaymentsCoreTests/HTTP_Tests.swift
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    func testPerformRequest_whenInvalidData_returnsParseError() async {
+        mockURLSession.cannedURLResponse = successURLResponse
+        mockURLSession.cannedJSONData = "{ \"test\" : \"bad-format\" }"
+
+        do {
+            _ = try await sut.performRequest(fakeRequest)
+            XCTFail()
```
jaxdesmarais: Same question here with `XCTFail`, worth adding a message here?

jaxdesmarais (approved):
Looks great overall! Just some small take it or leave it comments.

UnitTests/PaymentsCoreTests/HTTP_Tests.swift
scannillo: I think using `sut` (subject under test) is a good practice for us. We do it in a handful of other test files (AnalyticsEventRequest_Tests, DeviceInspector_Tests, DataCollector_Tests, and our Button tests). I know on Android they use it heavily too.

It makes very clear which object the file is testing. The `sut` should never be a mock, too.

jcnoriega (approved):

PR #115, opened 2023-01-31
Move test injection of MockURLSession to HTTP layer

scannillo:
### Reason for changes

- Follow up to PR #114. 
    - Move `MockURLSession` injection to `HTTP` layer since that is where it actually is used

### Summary of changes

- Move URLSession injection from `APIClient` to `HTTP`
- Add unit test to `APIClient_Test` that we're forwarding `APIRequests` to `HTTP` class for execution

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


UnitTests/PaymentsCoreTests/APIClient_Tests.swift
```diff
@@ -2,33 +2,38 @@ import XCTest
...
     override func setUp() {
         super.setUp()
-        config = CoreConfig(accessToken: mockAccessToken, environment: .sandbox)
         mockURLSession = MockURLSession()
-        mockURLSession.cannedError = nil
-        mockURLSession.cannedURLResponse = nil
-        mockURLSession.cannedJSONData = nil
-
-        apiClient = APIClient(urlSession: mockURLSession, coreConfig: config)
+        mockHTTP = MockHTTP(urlSession: mockURLSession, coreConfig: config)
```
jcnoriega, line 23: do we still need `MockUrlSession` in the test? `APIClient` now interacts directly with `HTTP`, so the  `URLSession` is an internal component of `HTTP` that `APIClient` doesnt know about.
`MockHTTP` could have its own init with no params that calls super with the corresponding mocks.

scannillo: Yes, that is desired end state! However there is a limitation in our current implementation that prevents us from fully mocking `HTTP.fetch()` without needing to call a super ([here](https://github.com/paypal/iOS-SDK/blob/e376d38bb270bd95f6e80b5e44cb6e7225f3e5b7/UnitTests/TestShared/MockHTTP.swift#L10-L14)). 

This is due to the use of the generic return type `T.ResponseType`. We can't instantiate it, therefore can't make a fake return value in the mock. So - currently, MockHTTP isn't really a _true mock_ since it calls the super and calls the actual method, which is why we end up still needing to stub the URLSession response. 

I was toying with addressing this now, but I think it would lead to a larger refactor of the `APIRequest` protocol that we use heavily. We can make a ticket to 🏈 punt on this.

jcnoriega: yeah totally, thanks for the detailed response! Makes sense. Perhaps we can write a ticket and leave it in the backlog just so we dont forget!

jcnoriega (approved):

UnitTests/PaymentsCoreTests/APIClient_Tests.swift
scannillo: Ticket created! DTNOR-811

jaxdesmarais (approved):
👻

PR #118, opened 2023-01-31
ClientID Caching ability; value required for analytics

scannillo:
### Reason for changes

JIRA [DTNOR-744](https://paypal.atlassian.net/browse/DTNOR-744)

This PR requires a `clientID` be successfully fetched (or cached) before proceeding with any of the 3 payment method flows (card, paypal web, and paypal native). 

### Summary of changes

- Add `withCaching` bool to  `HTTP.performRequest()`; defaults to false
   - Implement caching functionality when `withCaching` true
   - Add `URLCacheTestable` protocol for mocking `URLCache.shared` instance in tests
- Add `partner_client_id` param to `AnalyticsEventData`
- Move `PayPalError.clientIDNotFoundError` --> `CorePaymentsError` type
- Add clientID fetch to PayPalWebCheckoutClient, CardClient, PayPalNativeClient & updated respective unit test files

### Next Steps/Question for the team
We discussed as a team that we want to kick off the `getClientID()` fetch in the background as soon as a user gives us their access token (aka CoreConfig). Given the current SDK architecture, I felt this made more sense to kick off in the `APIClient` init.

This work will be completed in a follow-up PR, but I wanted to get folks thoughts.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/CardPayments/CardClient.swift
```diff
@@ -36,6 +36,13 @@ public class CardClient: NSObject {
     /// - Throws: PayPalSDK error if approve order could not complete successfully
     public func approveOrder(request: CardRequest) {
         Task {
+            do {
+                _ = try await apiClient.getClientID()
```
scannillo: I updated the docstring on `getClientID` to make it clear that the value is fetched from cache, if available. However, we _could_ update this method name to reflect that - if folks find that more clear. Open to name suggestions!

Sources/PaymentsCore/Networking/URLCacheTestable.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+protocol URLCacheTestable {
```
jaxdesmarais: Since we are using the protocol in the real implementation and not just testing do we think it's worth removing the test portion of the name? Or did we mean test in a different way?

Sources/CardPayments/CardClient.swift
jaxdesmarais: Yeah, I feel like updating the name of the method would be appropriate. I can't think of a name that isn't just way too wordy though. 🤔 Something like `fetchNewOrCachedClientID` was my initial though but geez if that isn't an absolute mouthful.

scannillo: Haha right? I keep thinking of `fetchOrReturnRemoteConfiguration` on BT, which functions the same way as this. But that method name was never intuitive to me either. I like your idea more, using the word `cached` is helpful!

How about `fetchCachedOrRemoteClientID`?

Sources/PaymentsCore/Networking/URLCacheTestable.swift
scannillo: So I thought that for protocols that _only_ exist for the purpose of injecting mocks, we could name them `Testable`? We currently have `URLSessionProtocol` which exists for the same purpose, so that we can inject a mock to replace `URLSession.shared`. But the name `URLSessionProtocol` doesn't follow the "adjective ending in -able or -ing" Swift language style guide.

I thought `Testable` might make it more clear what function the protocol serves. But it was more-so a trial to see what y'all think. 

Maybe `Mockable`? Or do you prefer just calling it `Protocol`?

Sources/CardPayments/CardClient.swift
jaxdesmarais: I dig `fetchCachedOrRemoteClientID`!

scannillo: Done!

Sources/PaymentsCore/Networking/URLCacheTestable.swift
jaxdesmarais: Agree that we should follow the Swift language guide (and update `URLSessionProtocol` to match at some point). I think `Mockable` could make sense if we want to differentiate our protocols that can be used to mock vs those that cannot. I guess the other option would be to name this something like `URLCacheable` or `URLCaching` then in our actual tests specify the mock like we do currently in `MockURLCache`?

scannillo: I like that idea! Updated

jcnoriega (approved):

Sources/PaymentsCore/Networking/HTTP.swift
```diff
@@ -5,23 +5,39 @@ class HTTP {
...
     
-    func performRequest<T: APIRequest>(_ request: T) async throws -> (T.ResponseType) {
+    func performRequest<T: APIRequest>(_ request: T, withCaching cachingEnabled: Bool = false) async throws -> (T.ResponseType) {
         guard let urlRequest = request.toURLRequest(environment: coreConfig.environment) else {
             throw APIClientError.invalidURLRequestError
         }
         
+        if cachingEnabled, let response = urlCache.cachedResponse(for: urlRequest) {
+            let decodedData = try decoder.decode(T.self, from: response.data)
+            return (decodedData)
```
jaxdesmarais: Do we need the parenthesis here?

Sources/PaymentsCore/Networking/URLCacheable.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+protocol URLCacheable {
+    
+    func cachedResponse(for request: URLRequest) -> CachedURLResponse?
+
```
jaxdesmarais: can we remove this space between the 2 functions?

jaxdesmarais (approved):

PR #119, opened 2023-02-08
Add environment to analytic events

scannillo:
### Reason for changes

Due to FPTI sandbox tooling limitations, we decided (summarized in [this doc](https://paypal-my.sharepoint.com/:w:/r/personal/scannillo_paypal_com1/_layouts/15/Doc.aspx?sourcedoc=%7B5D9C9200-40CD-4F09-B16D-2C313787AD04%7D&file=FPTI%20Data%20Tooling%20Summary%20%26%20Discussion.docx&nav=eyJjIjoxMTE0NjY2NDQxfQ&action=default&mobileredirect=true)) all analytic events will be sent to `https://api.paypal.com/`.

### Summary of changes

- Add `merchant_app_environment` parameter to distinguish sandbox vs production events.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/CorePayments/AnalyticsEventData.swift
+++ b/Sources/CorePayments/AnalyticsEventData.swift
@@ -18,6 +18,7 @@ struct AnalyticsEventData: Encodable {
         case clientOS = "client_os"
         case component = "comp"
         case deviceManufacturer = "device_manufacturer"
+        case environment = "merchant_app_environment"
         case eventName = "event_name"
         case eventSource = "event_source"
         case packageManager = "ios_package_manager"
@@ -47,6 +48,8 @@ struct AnalyticsEventData: Encodable {
     let eventName: String
 
     let eventSource = "mobile-native"
+    
+    let environment: String
 
     let packageManager: String = {
         #if COCOAPODS
@@ -87,7 +90,8 @@ struct AnalyticsEventData: Encodable {
 
     let tenantName = "PayPal"
     
-    init(eventName: String, clientID: String, sessionID: String) {
+    init(environment: String, eventName: String, clientID: String, sessionID: String) {
+        self.environment = environment
         self.eventName = eventName
         self.clientID = clientID
         self.sessionID = sessionID
@@ -105,6 +109,7 @@ struct AnalyticsEventData: Encodable {
         try eventParameters.encode(clientOS, forKey: .clientOS)
         try eventParameters.encode(component, forKey: .component)
         try eventParameters.encode(deviceManufacturer, forKey: .deviceManufacturer)
+        try eventParameters.encode(environment, forKey: .environment)
         try eventParameters.encode(eventName, forKey: .eventName)
         try eventParameters.encode(eventSource, forKey: .eventSource)
         try eventParameters.encode(packageManager, forKey: .packageManager)
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -33,7 +33,13 @@ class AnalyticsService {
     }
         
     func sendEvent(name: String, clientID: String) async {
+        guard let http else { // Will never occur
+            NSLog("[PayPal SDK]", "Failed to send analytics due to malformed HTTP client.")
+            return
+        }
+        
         let eventData = AnalyticsEventData(
+            environment: http.coreConfig.environment.toString,
             eventName: name,
             clientID: clientID,
             sessionID: sessionID
@@ -41,7 +47,7 @@ class AnalyticsService {
         
         do {
             let analyticsEventRequest = try AnalyticsEventRequest(eventData: eventData)
-            let (_) = try await http?.performRequest(analyticsEventRequest)
+            let (_) = try await http.performRequest(analyticsEventRequest)
         } catch {
             NSLog("[PayPal SDK] Failed to send analytics: %@", error.localizedDescription)
         }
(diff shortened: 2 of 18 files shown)


Sources/CorePayments/AnalyticsEventData.swift
```diff
@@ -18,6 +18,7 @@ struct AnalyticsEventData: Encodable {
         case clientOS = "client_os"
         case component = "comp"
         case deviceManufacturer = "device_manufacturer"
+        case environment = "merchant_app_environment"
```
scannillo, line 21: So, we can't overwrite the `env` param since it gets set by the FPTI service based on the host URL.

I opted to make the name of this really explicit to `merchant_app_environment` so we can distinguish it from the FPTI environment, since we are using FPTI live for both live/sand and prod events.

Would love folks feedback on the best name for this tag to make it extra clear. Either way, this is a new tag we will need to submit for tag approval so the world is our oyster for names.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -33,15 +33,21 @@ class AnalyticsService {
     }
         
     func sendEvent(name: String, clientID: String) async {
+        guard let http else {
+            NSLog("[PayPal SDK]", "Failed to send analytics due to nil HTTP client.")
+            return
+        }
```
scannillo, line 39: This will be removed in a follow-up PR 😈

Sources/CorePayments/AnalyticsEventData.swift
jaxdesmarais: I wonder if "merchant_configuration_environment" would be more clear since it's passed in the `CoreConfig` object?

Sources/CorePayments/Networking/Enums/Environment.swift
```diff
@@ -22,4 +22,13 @@ public enum Environment {
             return URL(string: "https://paypal.com/graphql")!
         }
     }
+    
+    public var toString: String {
```
jaxdesmarais, line 26: General thought - I think we've used a mix of things in these cases `toString` here, `stringValue` in others, and `description` in some places. Not part of this PR but I wonder if it'd be helpful at some point to audit this before the SDK gets too large. I think on BT as we've done the conversions we've generally used `stringValue` but I don't necessarily feel strongly tied to one name vs any other. It would be great to make this consistent across the SDKs though!

UnitTests/PaymentsCoreTests/AnalyticsService_Tests.swift
```diff
@@ -32,6 +32,21 @@ class AnalyticsService_Tests: XCTestCase {
         XCTAssert(mockHTTP.lastAPIRequest is AnalyticsEventRequest)
     }
     
+    func testSendEvent_whenProduction_sendsProperTag() async {
```
jaxdesmarais: Do we want a test for sandbox as well?

jaxdesmarais (approved):

Sources/CorePayments/AnalyticsEventData.swift
scannillo: Yeh, that's a good call. Do you think that would be more clear on the BT side too?

Worth noting - this value will **also** be used for differentiating BT SDK coming from Arachne.

So it'll go:
* Events from https://origin-analytics-prod.sandbox.braintree-api.com --> https://api.paypal.com/v1/tracking + `new_tag_name = sandbox`
* Events from https://origin-analytics-sand.sandbox.braintree-api.com --> https://api.paypal.com/v1/tracking + `new_tag_name = production`

scannillo: We _technically_ can sent `environment = whatever`, but I figured doing that would be confusing since a pre-populated `env` will always be set to `live`.

jaxdesmarais: Yeah I think that makes sense. Are you saying that this event is how we will tell BT Arachne vs BT FPTI events eventually (when we port Arachne to FPTI automatically)? I may have misunderstood that portion of your comment so want to make sure I followed what you were saying.

Sources/CorePayments/Networking/Enums/Environment.swift
scannillo: Yeh, good callout - that would be nice to standardize. Maybe we can look to Apple's APIs to see which language aligns best with their APIs.

Sources/CorePayments/AnalyticsEventData.swift
scannillo: Ah, no. So this value is just a tag that gets included in our analytics metadata payload. Previously (in BT) we don't include `env` in the payload since env is determined by the URL. But, when we move to FPTI we can't use two different URLs (due to the limitations mentioned above). So, we have to differentiate the env in the payload. Does that clear it up?

UnitTests/PaymentsCoreTests/AnalyticsService_Tests.swift
scannillo: Done!

Sources/CorePayments/AnalyticsEventData.swift
scannillo: cc: @DPoplin @debrado

jaxdesmarais: Ah! Gotcha, okay yeah that clears it up. Thanks!

jcnoriega (approved):

PR #123, opened 2023-03-01
Use Xcode 14 for tests & releases

scannillo:
### Reason for changes

- Our README calls out that we required Xcode 14+
- Let's us take advantage of new Swift 5.7 language features

### Summary of changes

- Bump Xcode version in GH Action workflow scripts
- Use new guard let self syntax

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -33,7 +33,7 @@ class AnalyticsService {
     }
         
     func sendEvent(name: String, clientID: String) async {
-        guard let http = http else {
+        guard let http else {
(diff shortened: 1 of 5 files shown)


Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -33,7 +33,7 @@ class AnalyticsService {
     }
         
     func sendEvent(name: String, clientID: String) async {
-        guard let http = http else {
+        guard let http else {
```
jaxdesmarais, line 36: 🚀

jaxdesmarais (approved):

PR #124, opened 2023-03-07
Embed PPRiskMagnes in Demo to enable device building

scannillo:
### Reason for changes

- Our demo needs to run on a device, so we can ensure each payment method flow works on real phones.

### Summary of changes

- Add the `PPRiskMagnes.xcframework` into the Demo app's `Frameworks, Libraries, and Embedded Content` settings

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo 
-


jaxdesmarais (approved):
👻

PR #128, opened 2023-03-16
Add `PayPalNativeCheckoutRequest` to wrap NXO `createOrderAction`

scannillo:
### Reason for changes

DTNOR-866

### Summary of changes

- Replace `PayPalNativeCheckoutClient.start(presentingViewController:createOrder)` with `PayPalNativeCheckoutClient.start(request:presentingViewController:)`
- Add `PayPalNativeCheckoutRequest` type

### Checklist

- [X] Added a changelog entry

### Authors
@scannillo


Sources/PayPalNativePayments/PayPalNativeCheckoutRequest.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+/// Used to configure options for approving a PayPal native order
+public struct PayPalNativeCheckoutRequest {
+
```
scannillo, line 5: NXO exposes "billing agreement token" and "vault approval session ID" we could technically also send. I opted not to, until we spend time fully testing those features.

Raising here in case folks feel differently.

jaxdesmarais: Agree with waiting to fully test those to add them. Would also be nice to know how much/if they are requested by merchants. I wonder if the MXO team can pull this data to see if it will be required as merchants transition from the direct MXO integration.

UnitTests/PayPalNativePaymentsTests/PayPalClient_Tests.swift
```diff
@@ -31,7 +33,7 @@ class PayPalClient_Tests: XCTestCase {
         
         let mockPayPalDelegate = MockPayPalDelegate()
         payPalClient.delegate = mockPayPalDelegate
-        await payPalClient.start { _ in }
+        await payPalClient.start(request: request)
```
jaxdesmarais, line 36: should this be called out in the `docs` or will that be in a follow up PR?

jaxdesmarais (approved):

UnitTests/PayPalNativePaymentsTests/PayPalClient_Tests.swift
scannillo: Follow-up! DTNOR-895

Sources/PayPalNativePayments/PayPalNativeCheckoutRequest.swift
scannillo: That's a great callout for us to keep in mind as we need to migrate MXO direct merchants over to PPCP.

demerino: +1 Keeping the scope tighter makes sense to start.

demerino (approved):

PR #130, opened 2023-03-17
Add optional `PayPalNativeShippingDelegate`

scannillo:
### Reason for changes

DTNOR-866

### Summary of changes

- Add wrapper types for NXO shipping change related classes
    - `PayPalNativeShippingAddress` and `PayPalNativeShippingMethod`
- Add optional `PayPalNativeShippingDelegate`
    - Houses two delegate methods:
        - `onShippingAddressChanged`
        - `onShippingMethodChanged`
     - I opted for 2 separate delegate methods instead of cramming them into 1. This way a merchant doesn't need to do any switch logic on the type of shipping info change.
- Remove `PayPalNativeCheckoutDelegate.onShippingChange` callback that exposed raw NXO callback types

### Checklist

- [x] Added a changelog entry

### Follow-Ups
- Update GH documentation page
- Move NXO to `implementationOnly` import (verify in SPM and CocoaPods apps)
- Dev doc updates
- Equivalent Android update

### Authors
@scannillo @jaxdesmarais


Sources/PayPalNativePayments/PayPalNativeShippingMethod.swift
```diff
@@ -0,0 +1,79 @@
+import PayPalCheckout
+
+// Wrapper for PayPalCheckout.ShippingMethod
+/// Shipping method details for an order via the PayPal Native Checkout UI.
+///
+/// If you want to show shipping options in the PayPal Native Paysheet,
+/// provide `purchase_units[].shipping.options` when creating an orderID with
+/// the [`orders/v2` API](https://developer.paypal.com/docs/api/orders/v2/#definition-purchase_unit) on your server. Otherwise, our Paysheet won't display any shipping options.
+public struct PayPalNativeShippingMethod {
```
jaxdesmarais, line 9: not for this PR but should we change everything from `PayPalNative/NXO` to `PayPalMobile/MXO` at some point before GA?

jaxdesmarais (approved):
🚀

Sources/PayPalNativePayments/PayPalNativeShippingMethod.swift
scannillo: I chose to name this `PayPalNativeShippingMethod` not because of internal naming of MXO/NXO, but because our module name is `PayPal Native Payments`.

We could rename this though if you have other ideas? Another inconsistency is we have an `Address` in `Card` module. I think at this point for simplicity/getting GA out we should keep the types separate per module, just maybe align the naming convention.

jaxdesmarais: Oh I more meant they changed their team name to "PayPal Mobile Checkout/MXO" so I wasn't sure if that was the public name we should be using as "<PRODUCT_NAME>ShippingMethod" for everything in this module (not just just struct, sorry weirdly placed comment)

scannillo: PayPal Native Payments is the module name for the PPCP SDK. There are a few internal comments referencing NXO we can rename to MXO if that's what you mean?

jaxdesmarais: Sure - that sounds good. It was more just a general question around not knowing from MXO what we should be referring to their SDK as and if that changed when their team name changed. I think we originally called it PayPal Native Payments because their product name was PayPal Native and now that it's not does that mean anything to us (the answer may be no, I just hadn't asked and thought about it).

scannillo: We named it `PayPal Native Payments` _without_ regard to the internal name. We had a discussion a bit in[ this HackMD](https://hackmd.io/zgQTvZ51SzSlI6MfHdWJBw), and finalized our names in a mob review session. We then submitted [this proposal ](https://paypal.sharepoint.com/:w:/r/sites/BTSDK/_layouts/15/Doc.aspx?sourcedoc=%7B490D7E50-64F1-4349-A778-0F5F1725F4AD%7D&file=Northstar%20Module%20Rename.docx&action=default&mobileredirect=true&cid=b66651d4-8c9e-4ba6-b650-cb4addad87c8)to product who ✔️ the renames.

> not knowing from MXO what we should be referring to their SDK
I agree though. The names are really similar and confusing across products, and internally!

jaxdesmarais: Gotcha, I was not aware that anything was submitted and approved by Product! I was only aware of the HackMD which was all inner team naming. Thanks for clarifying!

scannillo:
Follow-up PR to remove always-nil fields from `PayPalNativeShippingAddress` --> https://github.com/paypal/iOS-SDK/pull/133

PR #132, opened 2023-03-22
Update import instructions in docs

scannillo:
### Reason for changes

- Our docs had some inaccurate `import` statement instructions, which differ per package manager.
    - These were revealed via GMS/GPS feedback

### Summary of changes

- Remove source code `import` instructions. 
    - This is not something we include in Braintree docs. Merchants using each package manager should know (and defer to package manager documentation) for how to import modules into source code.
- Update incorrect Podfile snippets
- Use consistent wording for SPM instructions

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
👻

PR #137, opened 2023-04-13
add back expiry field for credit card

KunJeongPark:
Added expiry field back for credit card authorization.

### Authors
@KunJeongPark

### DIFF
--- a/Sources/CardPayments/Models/Card.swift
+++ b/Sources/CardPayments/Models/Card.swift
@@ -29,6 +29,11 @@ public struct Card: Encodable {
 
     /// Optional. The billing address
     public var billingAddress: Address?
+    
+    /// The expiration year and month, in ISO-8601 `YYYY-MM` date format.
+    public var expiry: String {
+        "\(expirationYear)-\(expirationMonth)"
+    }
 
     internal var attributes: Attributes?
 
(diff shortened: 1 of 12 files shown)


Sources/CardPayments/Models/Card.swift
```diff
@@ -29,6 +29,11 @@ public struct Card: Encodable {
 
     /// Optional. The billing address
     public var billingAddress: Address?
+    
+    /// The expiration year and month, in ISO-8601 `YYYY-MM` date format.
+    public var expiry: String {
+        "\(expirationYear)-\(expirationMonth)"
+    }
```
scannillo: We don't want to add `expiry` as a public param, since this doesn't maintain parity with Android. Instead, we want the SDK to for the formatting work and send `expiry` to the orders API internally, using the `expirationDate` and `expirationYear` set by the merchant.

scannillo: See Android https://github.com/paypal/Android-SDK/blob/c18d6d675065496e43661fb80eed421f835d6485/CardPayments/src/main/java/com/paypal/android/cardpayments/CardRequestFactory.kt#L13

KunJeongPark: I see. Thank you.

jcnoriega (approved):

scannillo (approved):
Looks good to me!

PR #138, opened 2023-04-14
Fix swift lint errors

jcnoriega:
### Reason for changes
Swiftlint errors were appearing in the pipeline, that did not appear locally. It was due to a swiftlint version mismatch between my computer and what runs on GA. Run brew upgrade swiftlint

### Summary of changes

- Fix swift errors

### Checklist

- [ ] Added a changelog entry

- @jcnoriega

### DIFF
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceResponse.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceResponse.swift
@@ -16,3 +16,4 @@ struct Link: Decodable {
     let href: String?
     let rel, method: String?
 }
+// swiftlint:enable space_after_main_type
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceResponse.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceResponse.swift
@@ -1,11 +1,10 @@
-// swiftlint:disable space_after_main_type
-
 import Foundation
 #if canImport(CorePayments)
 import CorePayments
 #endif
 
 struct ConfirmPaymentSourceResponse: Decodable {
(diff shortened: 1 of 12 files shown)


KunJeongPark (approved):

Sources/CardPayments/APIRequests/ConfirmPaymentSourceResponse.swift
```diff
@@ -16,3 +16,4 @@ struct Link: Decodable {
     let href: String?
     let rel, method: String?
 }
+// swiftlint:enable space_after_main_type
```
sshropshire: q: Is this because there are multiple types in one file?

jcnoriega: now that I see it, it is not needed, ill remove it

jcnoriega: @sshropshire fixed!

Sources/CorePayments/Networking/Enums/Environment.swift
```diff
@@ -32,3 +32,4 @@ public enum Environment {
         }
     }
 }
+// swiftlint:enable force_unwrapping
```
scannillo: I don't think we need to re-enable the rule at the end of the files

jcnoriega: the new version of swiftlint requires it. That is why most of the iOS prs where failing, because we didnt re enable whatever rule we disabled before. Unless we do a `disable:next`, it asks to enable again.

jcnoriega:
changes merge in #137

PR #139, opened 2023-04-17
Add patching capabilites to demo app

jcnoriega:
### Reason for changes
We werent able to test the complete flow of shipping change for NXO, now we can.

### Summary of changes

- Add pacthing capabilities to Demo app.

Selecting a new shipping method:

https://user-images.githubusercontent.com/10046176/232631340-8edd6ad9-7196-4658-8f6e-e37436a97ef2.mov

Selecting a new address, that gets rejected:

https://user-images.githubusercontent.com/10046176/232631800-f662e831-00b6-47af-9811-ef015d6101bf.mov

### Checklist

- [ ] Added a changelog entry

- @jcnoriega


Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -37,13 +34,24 @@ enum OrderRequestHelpers {
             )
         )
     }
+    
+    static func getAmount(value: Double = orderAmount, shipping: Double = 3.99) -> PayPalCheckout.PurchaseUnit.Amount {
+        return PayPalCheckout.PurchaseUnit.Amount(
+            currencyCode: .usd,
+            value: String(value + shipping),
```
sshropshire, line 41: q: Can we move shipping calculations to the merchant server?

jcnoriega: im guessing we could in the future, but i would not do it for this PR. I think it might be a bit more complex as the server will have to keep some sort of internal state of actual the actual order being processed. It will also help us to decouple from MXO value objects

Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -110,15 +111,30 @@ extension PayPalViewModel: PayPalNativeShippingDelegate {
...
-        shippingActions.approve()
+        Task {
+            do {
+                let shippingMethods = OrderRequestHelpers.getShippingMethods(selectedID: shippingMethod.id)
+                let amount = OrderRequestHelpers.getAmount(shipping: Double(shippingMethod.value ?? "0.0") ?? 0.0)
+                let params = UpdateOrderParams(orderID: orderID, shippingMethods: shippingMethods, amount: amount)
+                try await DemoMerchantAPI.sharedService.updateOrder(params)
+                shippingActions.approve()
+            } catch let error {
+                shippingActions.reject()
```
scannillo, line 135: Here is a video of what "reject" looks like if we reject after the shippingMethod change patch order fails.

It's a little odd to me that it reads "Merchant can't ship to this address, choose a different one" when the shipping method was the actual problem. But I think this is the only functionality that NXO exposes right?

https://user-images.githubusercontent.com/35243507/232821844-5f4b353f-d14e-41a4-98fc-bbed66730ae0.mov

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: We could have the Demo app only send the shipping method amount, and then the server fetch the orderID amount and do the math on the server. I think this is what we expect merchants to do. But, we can 🏈 punt on that probably. Maybe add a JIRA ticket.

Demo/Demo/ViewModels/PayPalViewModel.swift
jcnoriega: @scannillo yeap. They dont even offer a why to close the paysheet manually. So this is the only way we can interact with the paysheet if `patch` fails or if we want to display an error

scannillo: Interesting, thanks for checking!

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -10,10 +10,7 @@ enum OrderRequestHelpers {
             intent: .authorize,
             purchaseUnits: [
                 PayPalCheckout.PurchaseUnit(
```
scannillo, line 12: We want to add a ticket to our backlog to remove the Demo app's use of NXO for creating an order, as this is not what a merchant should do.

scannillo (approved):
Looks good to me, let's just make a ticket for moving the addition of order value + shipping value to the server

Demo/Demo/Helpers/OrderRequestHelpers.swift
sshropshire: Yeah this isn't a blocker for me, but duplication of this logic could lead to "works on iOS but not Android" issues and vice versa.

sshropshire (approved):

PR #140, opened 2023-04-18
PayPalNativePayments: Finish unit test and add them to code coverage

jcnoriega:
### Reason for changes
Long overdue ticket: https://paypal.atlassian.net/browse/DTNOR-569?atlOrigin=eyJpIjoiNjUwNWZhN2VlMzk5NDk0Yzk4ZjkwZjczNDNjYWU5MzciLCJwIjoiaiJ9

Now that we have a final PayPalNativePayments interface, we complete unit test and add them to code coverage

### Summary of changes

- Decouple `NativeCheckoutStartable` protocol from MXO classes and callbacks as much as we can. This allows to actually test every possible callback, without having to deal with MXO classes
- Add internal `ShippingActionsProtocol` to test `PayPalNativeShippingAction`
- Add internal `CheckoutProtocol` to mock `Checkout` in order to test `NativeCheckoutProvider`

### Checklist

- [ ] Added a changelog entry

- @jcnoriega


Sources/PayPalNativePayments/PayPalNativePaysheetActions.swift
```diff
@@ -3,16 +3,19 @@ import PayPalCheckout
 /// The actions that can be used to update the Paysheet UI after `PayPalNativeShippingDelegate` methods are invoked.
 public class PayPalNativePaysheetActions {
 
-    private let shippingActions: ShippingChangeAction
+    private let shippingActions: ShippingActionsProtocol?
```
sshropshire:
```suggestion
    private let shippingActions: ShippingActionsProtocol
```

sshropshire (approved):
🌊

Sources/PayPalNativePayments/NativeCheckoutStartable.swift
```diff
@@ -6,16 +6,26 @@ import CorePayments
 #endif
 
 protocol NativeCheckoutStartable {
+    
+    typealias StartabeApproveCallback = (String, String) -> Void
```
scannillo: Looks like a typo `Startabe` --> `Startable` for all of these types

Sources/PayPalNativePayments/PayPalNativePaysheetActions.swift
```diff
@@ -21,5 +24,13 @@ public class PayPalNativePaysheetActions {
...
+    public func approve() { self.shippingActions?.approve() }
 }
+
+protocol ShippingActionsProtocol {
+    
+    func reject()
+    func approve()
+}
+
+extension ShippingChangeAction: ShippingActionsProtocol { }
```
scannillo, line 33: 👏

Sources/PayPalNativePayments/NativeCheckoutProvider.swift
```diff
@@ -7,27 +7,74 @@ import CorePayments
...
+    private func guaranteeMainThread(_ work: @escaping () -> Void) {
+        if Thread.isMainThread {
+            work()
+        } else {
+            DispatchQueue.main.async(execute: work)
+        }
+    }
 }
+
+protocol CheckoutProtocol {
```
scannillo, line 64: I think we can remove `NativeCheckoutStartable` now that we have `CheckoutProtocol` which we use to access `start()` via `checkout.start()` in `NativeCheckoutProvider`

Sources/PayPalNativePayments/NativeCheckoutStartable.swift
scannillo: Take it or leave it coment - I think `Startable` for all the callback protocol wrappers is a bit confusing. We could just follow `<ThingToWrap>Protocol` naming convention, like you did for `CheckoutProtocol`.

I know we generally try to follow the `able` or `ing` adjective syntax recommended by Apple, but this is a weird case.

Sources/PayPalNativePayments/NativeCheckoutStartable.swift
```diff
@@ -6,16 +6,26 @@ import CorePayments
...
+        PayPalNativeShippingMethod?
+    ) -> Void
+    typealias StartabeCancelCallback = () -> Void
+    typealias StartabeErrorCallback = (String) -> Void
 
-    // swiftlint:disable function_parameter_count
+
+    // swiftlint:disable:next function_parameter_count
     func start(
         presentingViewController: UIViewController?,
```
KunJeongPark, line 23: how come we need to explicitly pass in presentingVC? I know in Braintree, we don't pass this param and I think it defaults to root most vc?

jcnoriega: i think in BT it was done for consistency reasons maybe @scannillo? Here if the merchant doesnt pass the VC it will default to root.

scannillo (approved):

PR #141, opened 2023-04-20
Disable lint rules for testing

jcnoriega:
### Reason for changes

We disable some specific lint rules for unit tests
Ticket: [DTNOR-628](https://paypal.atlassian.net/browse/DTNOR-628)

### Summary of changes

- Add .swfitliny.yaml to UnitTest folder to disable rules under that specific path
- Add rules to ignore
- Remove "config_file" from Fastlane, so that it picks up both swiftlint.ymls

### Checklist

- [ ] Added a changelog entry

@jcnoriega 
-

### DIFF
--- a/Sources/CorePayments/CoreConfig.swift
+++ b/Sources/CorePayments/CoreConfig.swift
@@ -5,7 +5,7 @@ import Foundation
 public struct CoreConfig {
 
     public let environment: Environment
-    public let accessToken: String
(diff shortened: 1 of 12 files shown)


sshropshire (approved):

scannillo (approved):
Awesome! Thanks for tackling this, will make our lives much easier

PR #144, opened 2023-05-04
Send `orderID` instead of `sessionID` to FPTI analytics

scannillo:
### Reason for changes

- The analytics team suggests we track and use the `orderID` to link events together, instead of a `sessionID`

### Summary of changes

- Change `AnalyticsService` from singleton to basic struct
- Remove analytics capabilities from `APIClient`
- Each feature client now has it's own `AnalyticsService` instance
    - Move `clientID` fetch requirement to `AnalyticsService` layer
- Send `orderID` instead of `sessionID` in `AnalyticsEventData`
- Rename `merchant_app_environment` to `merchant_sdk_env` to match Arachne --> FPTI tag naming convention

### Checklist

- [x] Added a changelog entry

### Authors
@scannillo

### DIFF
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -1,55 +1,57 @@
 import Foundation
 
 /// Constructs `AnalyticsEventData` models and sends FPTI analytics events.
-class AnalyticsService {
+public class AnalyticsService {
     
     // MARK: - Internal Properties
-
-    private static let instance = AnalyticsService()
     
-    private var sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
-    
-    private var http: HTTP? {
-        // This property observer generates a new `sessionID` if the `accessToken`
-        // injected into the `AnalyticsSingleton` ever changes.
-        willSet {
-            if newValue?.coreConfig.accessToken != http?.coreConfig.accessToken {
-                sessionID = UUID().uuidString.replacingOccurrences(of: "-", with: "")
-            }
-        }
-    }
+    private let coreConfig: CoreConfig
+    private let http: HTTP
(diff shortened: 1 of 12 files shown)


scannillo:
This is an odd CocoaPods failure. `AnalyticsService` has a public and internal version of `sendEvent` (one is for testing). The feature client's can't access internal methods. The Demo can differentiate them just fine, but CocoaPods is losing it's mind.

jcnoriega (approved):

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -1,55 +1,66 @@
...
             let analyticsEventRequest = try AnalyticsEventRequest(eventData: eventData)
             let (_) = try await http.performRequest(analyticsEventRequest)
         } catch {
             NSLog("[PayPal SDK] Failed to send analytics: %@", error.localizedDescription)
         }
     }
+    
+    private func fetchCachedOrRemoteClientID() async throws -> String {
+        let request = GetClientIDRequest(accessToken: coreConfig.accessToken)
+        let (response) = try await http.performRequest(request, withCaching: true)
```
jaxdesmarais: general question, do we need the parenthesis here? Or can we do:
```suggestion
        let response = try await http.performRequest(request, withCaching: true)
```

jaxdesmarais (approved):

scannillo:
> This is an odd CocoaPods failure. AnalyticsService has a public and internal version of sendEvent (one is for testing). The feature client's can't access internal methods. The Demo can differentiate them just fine, but CocoaPods is losing it's mind.

Jax and I resolved this. We had to change the method names, to satisfy CocoaPods. No rhyme or reason, just a CocoaPods issue not recognizing that async-await functions are different signatures.

PR #149, opened 2023-05-08
make sure only cache valid data in HTTP performRequest

KunJeongPark:
### Reason for changes

-We have an issue on both GH and Slack that claim when running on device, the SDKs’s clientID fetch before the NXO flow fails and returns the error: Error fetching clientID. Upon deleting app and reinstalling, there were no more issues.
We suspect that an error response got cached. 

### Summary of changes

- check for valid data when storing and retrieving from cache

### Checklist

### Authors
@KunJeongPark  @scannillo 

-

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -60,7 +60,7 @@ public class CardClient: NSObject {
             } catch let error as CoreSDKError {
                 analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
                 notifyFailure(with: error)
-            } catch {            
+            } catch {
             }
         }
     }
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -61,6 +61,7 @@ public class CardClient: NSObject {
                 analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
                 notifyFailure(with: error)
             } catch {
+                // TODO: notifyFailure and analytics here?
             }
         }
     }
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -60,7 +60,7 @@ public class CardClient: NSObject {
             } catch let error as CoreSDKError {
                 analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
                 notifyFailure(with: error)
-            } catch {                
+            } catch {            
             }
         }
     }
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -60,8 +60,7 @@ public class CardClient: NSObject {
             } catch let error as CoreSDKError {
                 analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
                 notifyFailure(with: error)
-            } catch {
-                // TODO: notifyFailure and analytics here?
+            } catch {                
             }
         }
     }
--- a/Sources/CorePayments/Networking/HTTP.swift
+++ b/Sources/CorePayments/Networking/HTTP.swift
@@ -36,11 +36,12 @@ class HTTP {
         
         switch response.statusCode {
         case 200..<300:
-            let decodedData = try decoder.decode(T.self, from: data)
             if cachingEnabled {
                 let cachedURLResponse = CachedURLResponse(response: response, data: data)
                 urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
             }
+            
+            let decodedData = try decoder.decode(T.self, from: data)
             return (decodedData)
         default:
             let errorData = try decoder.decode(from: data)
--- a/Sources/CorePayments/Networking/HTTP.swift
+++ b/Sources/CorePayments/Networking/HTTP.swift
@@ -24,9 +24,8 @@ class HTTP {
         }
         
         if cachingEnabled, let response = urlCache.cachedResponse(for: urlRequest) {
-            if let decodedData = try? decoder.decode(T.self, from: response.data) {
-                return decodedData
-            }
+            let decodedData = try decoder.decode(T.self, from: response.data)
+            return decodedData
         }
         
         let (data, response) = try await urlSession.performRequest(with: urlRequest)
--- a/Sources/CorePayments/Networking/HTTP.swift
+++ b/Sources/CorePayments/Networking/HTTP.swift
@@ -24,8 +24,9 @@ class HTTP {
         }
         
         if cachingEnabled, let response = urlCache.cachedResponse(for: urlRequest) {
-            let decodedData = try decoder.decode(T.self, from: response.data)
-            return decodedData
+            if let decodedData = try? decoder.decode(T.self, from: response.data) {
+                return decodedData
+            }
         }
         
         let (data, response) = try await urlSession.performRequest(with: urlRequest)
@@ -33,14 +34,13 @@ class HTTP {
             throw APIClientError.invalidURLResponseError
         }
         
-        if cachingEnabled {
-            let cachedURLResponse = CachedURLResponse(response: response, data: data)
-            urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
-        }
-        
         switch response.statusCode {
         case 200..<300:
             let decodedData = try decoder.decode(T.self, from: data)
+            if cachingEnabled {
+                let cachedURLResponse = CachedURLResponse(response: response, data: data)
+                urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
+            }
             return (decodedData)
         default:
             let errorData = try decoder.decode(from: data)
--- a/Sources/CorePayments/Networking/HTTP.swift
+++ b/Sources/CorePayments/Networking/HTTP.swift
@@ -35,12 +35,11 @@ class HTTP {
         
         switch response.statusCode {
         case 200..<300:
+            let decodedData = try decoder.decode(T.self, from: data)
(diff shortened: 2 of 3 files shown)


Sources/CorePayments/Networking/HTTP.swift
```diff
@@ -24,23 +24,23 @@ class HTTP {
         }
         
         if cachingEnabled, let response = urlCache.cachedResponse(for: urlRequest) {
-            let decodedData = try decoder.decode(T.self, from: response.data)
-            return decodedData
+            if let decodedData = try? decoder.decode(T.self, from: response.data) {
+                return decodedData
```
scannillo: We can actually leave this as it was. Any `throw` from the[ decode function here ](https://github.com/paypal/iOS-SDK/blob/c218afc8c8bbe7daefedd363a94c9eb636da2040/Sources/CorePayments/Networking/APIClientDecoder.swift#L12)will get bubbled up and thrown from `performRequest`

Sources/CorePayments/Networking/HTTP.swift
```diff
@@ -24,23 +24,23 @@ class HTTP {
...
-        }
-        
         switch response.statusCode {
         case 200..<300:
             let decodedData = try decoder.decode(T.self, from: data)
+            if cachingEnabled {
+                let cachedURLResponse = CachedURLResponse(response: response, data: data)
+                urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
+            }
             return (decodedData)
```
scannillo, line 43: Nitpick for organization

```suggestion
            if cachingEnabled {
                let cachedURLResponse = CachedURLResponse(response: response, data: data)
                urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
            }
            
            let decodedData = try decoder.decode(T.self, from: data)
            return (decodedData)
```

scannillo:
I also don't think our cache should perform any on-disk caching, only in-memory. I don't think we want cached clientIDs from old app launches. What do y'all think?

The default [`URLCache.shared` ](https://developer.apple.com/documentation/foundation/urlcache/1413377-shared)instance that we're using has default sizes of: 
 - Memory capacity: 4 megabytes (4 * 1024 * 1024 bytes)
 - Disk capacity: 20 megabytes (20 * 1024 * 1024 bytes)

We could do something like this to specify a custom cache size:

```
    /// Initializer
    /// - Parameters:
    ///   - urlSession: Apple's default instance of `URLSession.shared`.
    ///   - urlCache: Has a maximum memory usage of 4 megabytes (4 * 1024 * 1024 bytes). Does not cache to disk.
    ///   - coreConfig: The merchant's ``CoreConfig``.
    init(
        urlSession: URLSessionProtocol = URLSession.shared,
        urlCache: URLCacheable = URLCache(memoryCapacity: 4 * 1024 * 1024, diskCapacity: 0),
        coreConfig: CoreConfig
    ) { ... }
```

Sources/CorePayments/Networking/HTTP.swift
KunJeongPark: Sammy, is that what we want? Do we want it to just throw error or do we just force it to get value again if cached value is not decoded for some reason?

KunJeongPark:
Great point Sammy. With clientId, I don't think we have to worry about memory capacity, are there times that we do? I was looking up eviction policies of URLCache, and it's kind of a black box, "internal algorithm", I think.

scannillo:
> Great point Sammy. With clientId, I don't think we have to worry about memory capacity, are there times that we do? I was looking up eviction policies of URLCache, and it's kind of a black box, "internal algorithm", I think.

If we want to turn off the on-disk caching, we have to set a custom value for `URLCache` which requires us pass `Int` values for the disk and memory size. This is how we can "turn off" disk caching by setting it's max size to 0. 

As for the 4MB size in my example, I pulled that from Apple's `URLCache.shared` docstrings (click into it on Xcode), since that is their default value used. We could use that, or go smaller probably. 

We can do some debugging to see the size of our cache instance after 1 clientID cache and then multiple it by ~X10 or so. (Android caches up to 10 clientIDs)

KunJeongPark:
> We can do some debugging to see the size of our cache instance after 1 clientID cache and then multiple it by ~X10 or so. (Android caches up to 10 clientIDs)

What I meant was are we going to use this function to fetch anything else that might be larger in size?

Sources/CorePayments/Networking/HTTP.swift
scannillo: That's a good question. I think with the change below, we know that we're only putting "success" data into the cache which is good enough.

KunJeongPark: We don't want to worry about some weird case where decoding is not successful? I guess if we try again, likely it would give similar error shortly after.

scannillo:
> What I meant was are we going to use this function to fetch anything else that might be larger in size?

Ah, no - there aren't current plans of using this further beyond clientID caching for analytics.

Eric-Catalyst:
> Great point Sammy. With clientId, I don't think we have to worry about memory capacity, are there times that we do? I was looking up eviction policies of URLCache, and it's kind of a black box, "internal algorithm", I think.

Apple is very good about following RFC-9111, so the algorithms there are what Apple has implemented.

Sources/CorePayments/Networking/HTTP.swift
jaxdesmarais: It looks like in the code Sammy linked above if decoding is not successful we would always throw an error. Unless there is a specific case that you are thinking of that we aren't handling in the decoder?

jaxdesmarais (approved):

Sources/CorePayments/Networking/HTTP.swift
KunJeongPark: Not sure, I thought maybe the decoding fails for some reason but since we are only caching valid data, I think I can revert.

scannillo: @KunJeongPark Let's revert this change, and then I can ✅

KunJeongPark:
@scannillo Do we want to do change from on-disk caching to  in-memory cache on this ticket?
I made another ticket for it. https://paypal.atlassian.net/browse/DTBTSDK-2719

Sources/CorePayments/Networking/HTTP.swift
KunJeongPark: Done

scannillo (approved):
🎸

Sources/CorePayments/Networking/HTTP.swift
```diff
@@ -33,13 +33,13 @@ class HTTP {
...
-        }
-        
         switch response.statusCode {
         case 200..<300:
+            if cachingEnabled {
+                let cachedURLResponse = CachedURLResponse(response: response, data: data)
+                urlCache.storeCachedResponse(cachedURLResponse, for: urlRequest)
+            }
+            
             let decodedData = try decoder.decode(T.self, from: data)
```
KunJeongPark: I think I should decode on top under line 37. Right now we are just screening for bad status codes before caching, not decoding error. Wdyt @scannillo @jaxdesmarais  changed it on last commit: 4c17247d7482d65176830c3bf1d5b1fd99482398

KunJeongPark: I also saw in the CardClient that we are not dealing with all errors like decoding errors. In approveOrder function, we don't send notifyFailure or analytic event in the general catch block and in getOrderInfo only catches CoreSDKError. Is there a reason for this?

Sources/CardPayments/CardClient.swift
```diff
@@ -61,6 +61,7 @@ public class CardClient: NSObject {
                 analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
                 notifyFailure(with: error)
             } catch {
+                // TODO: notifyFailure and analytics here?
```
KunJeongPark: I also saw in the CardClient that we are not dealing with all errors like decoding errors. In approveOrder function, we don't send notifyFailure or analytic event in the general catch block . Is there a reason for this?

scannillo: Can you include links to the lines of code in CardClient you're referring to? It is hard to know what part you mean without a reference

scannillo: We should definitely handle the general `catch` case. I don't think we need two `catch` blocks either, it can all just be the general catch block, something like:

```
} catch let error {
    analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
    notifyFailure(with: error)
}
```

If you're going to leave a TODO, make sure it is actionable and clear and ideally also has a JIRA ticket number. I personally would remove the TODO for now and just make a JIRA with links to address this in the future.

KunJeongPark: I think I will just change it on this ticket as I think it's related and important. What do you think?

jaxdesmarais: We don't even need the let error (catch defines that property for us for free! 👯 ) can be even further simplified to:
```swift
} catch {
    analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:failed")
    notifyFailure(with: error)
}
```

KunJeongPark: Compiler is not happy with out the error on notify line.

KunJeongPark: Yeah, we have to change expected error type on notifyFailure and card delegate function with didFinishWith Error param, type in CardDelegate protocol function. I guess the intention was to send more specific errors and deal with others later. Here is the jira ticket: https://paypal.atlassian.net/browse/DTBTSDK-2724 I will remove TODO comment.

PR #150, opened 2023-05-10
Clarify docs that `PayPalNativeShippingDelegate` is sometimes required

scannillo:
### Reason for changes

- [This GH issue](https://github.com/paypal/iOS-SDK/issues/146#issuecomment-1538689698) raised a question about when a merchant needs to implement the shipping callback. Turns out, it is not entirely optional. There is a case when a merchant must set it to avoid an infinite spinner on their checkout button.
- Special thanks to @laurentbourg for highlighting this area of improvement 👏 

### Summary of changes

- Clarify that merchants using `GET_FROM_FILE` orders must implement `PayPalNativeShippingDelegate`

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/PayPalNativePayments/PayPalNativeCheckoutDelegate.swift
+++ b/Sources/PayPalNativePayments/PayPalNativeCheckoutDelegate.swift
@@ -28,7 +28,7 @@ public protocol PayPalNativeCheckoutDelegate: AnyObject {
     func paypalWillStart(_ payPalClient: PayPalNativeCheckoutClient)
 }
 
-/// A delegate to receive notifcations if the user changes their shipping information.
+/// A delegate to receive notifications if the user changes their shipping information.
 ///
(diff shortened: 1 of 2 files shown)


Sources/PayPalNativePayments/PayPalNativeCheckoutDelegate.swift
```diff
@@ -28,7 +28,9 @@ public protocol PayPalNativeCheckoutDelegate: AnyObject {
     func paypalWillStart(_ payPalClient: PayPalNativeCheckoutClient)
 }
 
-/// An optional delegate to receive notifcations if the user changes their shipping information.
+/// A delegate to receive notifcations if the user changes their shipping information.
```
jaxdesmarais:
```suggestion
/// A delegate to receive notifications if the user changes their shipping information.
```

jaxdesmarais (approved):

wallportillo (approved):
@scannillo these look good to me as well

PR #152, opened 2023-05-14
catch encoding error and general errors in cardClient

KunJeongPark:
### Reason for changes
-CardClient had general catch block with no return to merchant app
-Errors propagated from numerous sources and expecting CoreSDKError type in card delegate function is risky
 without general case handling

### Summary of changes

- Added catching encoding error in ConfirmPaymentRequest
- Added unknownError property in CardClientError
- Send CardClientError.unknownError in general catch block in CardClient's approveOrder function

### Checklist

### DIFF
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -49,6 +49,7 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         // TODO - The complexity in this `init` signals to reconsider our use/design of the `APIRequest` protocol.
         // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
     }
+    
     // MARK: - APIRequest
     
     typealias ResponseType = ConfirmPaymentSourceResponse
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -48,7 +48,7 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
     }
     
-    /// :nodoc: for testing purpose
+    /// For testing purpose
     init(
         accessToken: String,
         cardRequest: CardRequest,
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -18,7 +18,6 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         accessToken: String,
         cardRequest: CardRequest
     ) throws {
-       
         self.jsonEncoder = JSONEncoder()
         var confirmPaymentSource = ConfirmPaymentSource()
         var card = cardRequest.card
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -9,52 +9,20 @@ struct ConfirmPaymentSourceRequest: APIRequest {
     private let orderID: String
     private let pathFormat: String = "/v2/checkout/orders/%@/confirm-payment-source"
     private let accessToken: String
-    var jsonEncoder: JSONEncoder?
+    var jsonEncoder: JSONEncoder
     
     /// Creates a request to attach a payment source to a specific order.
     /// In order to use this initializer, the `paymentSource` parameter has to
     /// contain the entire dictionary as it exists underneath the `payment_source` key.
-    init(
-        accessToken: String,
-        cardRequest: CardRequest
-    ) throws {
-        self.jsonEncoder = JSONEncoder()
-        var confirmPaymentSource = ConfirmPaymentSource()
-        var card = cardRequest.card
-        let verification = Verification(method: cardRequest.sca.rawValue)
-        card.attributes = Attributes(verification: verification)
-            
-        confirmPaymentSource.applicationContext = ApplicationContext(
-            returnUrl: PayPalCoreConstants.callbackURLScheme + "://card/success",
-            cancelUrl: PayPalCoreConstants.callbackURLScheme + "://card/cancel"
-        )
-        
-        confirmPaymentSource.paymentSource = PaymentSource(card: card)
-        
-        self.orderID = cardRequest.orderID
-        self.accessToken = accessToken
-        
-        path = String(format: pathFormat, orderID)
-        
-        jsonEncoder?.keyEncodingStrategy = .convertToSnakeCase
-        do {
-            body = try jsonEncoder?.encode(confirmPaymentSource)
-        } catch {
-            throw CardClientError.encodingError
-        }
-        
-        // TODO - The complexity in this `init` signals to reconsider our use/design of the `APIRequest` protocol.
-        // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
-    }
     
     /// For testing purpose
     init(
         accessToken: String,
         cardRequest: CardRequest,
-        encoder: JSONEncoder?
+        encoder: JSONEncoder = JSONEncoder()
     ) throws {
         // encode with custom encoder that throws error if passed in
-        self.jsonEncoder = encoder ?? JSONEncoder()
+        self.jsonEncoder = encoder
         var confirmPaymentSource = ConfirmPaymentSource()
         var card = cardRequest.card
         let verification = Verification(method: cardRequest.sca.rawValue)
@@ -72,9 +40,9 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         
         path = String(format: pathFormat, orderID)
         
-        jsonEncoder?.keyEncodingStrategy = .convertToSnakeCase
+        jsonEncoder.keyEncodingStrategy = .convertToSnakeCase
         do {
-            body = try jsonEncoder?.encode(confirmPaymentSource)
+            body = try jsonEncoder.encode(confirmPaymentSource)
         } catch {
             throw CardClientError.encodingError
         }
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -36,7 +36,11 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         path = String(format: pathFormat, orderID)
         
         jsonEncoder.keyEncodingStrategy = .convertToSnakeCase
-        body = try jsonEncoder.encode(confirmPaymentSource)
+        do {
+            body = try jsonEncoder.encode(confirmPaymentSource)
+        } catch {
+            throw CardClientError.encodingError
+        }
         
         // TODO - The complexity in this `init` signals to reconsider our use/design of the `APIRequest` protocol.
         // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
--- a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
+++ b/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
@@ -9,15 +9,21 @@ struct ConfirmPaymentSourceRequest: APIRequest {
     private let orderID: String
     private let pathFormat: String = "/v2/checkout/orders/%@/confirm-payment-source"
     private let accessToken: String
-    private let jsonEncoder = JSONEncoder()
+    var jsonEncoder: JSONEncoder?
     
     /// Creates a request to attach a payment source to a specific order.
     /// In order to use this initializer, the `paymentSource` parameter has to
     /// contain the entire dictionary as it exists underneath the `payment_source` key.
     init(
         accessToken: String,
-        cardRequest: CardRequest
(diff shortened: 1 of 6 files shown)


jcnoriega (approved):

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -36,7 +36,11 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         path = String(format: pathFormat, orderID)
         
         jsonEncoder.keyEncodingStrategy = .convertToSnakeCase
-        body = try jsonEncoder.encode(confirmPaymentSource)
+        do {
+            body = try jsonEncoder.encode(confirmPaymentSource)
+        } catch {
+            throw CardClientError.encodingError
+        }
```
jaxdesmarais, line 46: Are we able to add a unit test for this logic?

KunJeongPark: oh the encoding error, I see. A bit tricky. I don't see a way of doing this without modifying the initializer to take in a jsonEncoder overriding encode function in subclass of jsonEncoder to throw an error. Not worth the time and change in code, imo

KunJeongPark: I just added one but I don't think it's worth adding extra code in` ConfirmPaymentSourceRequest` for encoding error unit test

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -73,4 +73,39 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
+    
+    class FailingJSONEncoder: JSONEncoder {
+        
+        override func encode<T>(_ value: T) throws -> Data where T: Encodable {
+            throw TestError.encodeError
+        }
+    }
+
+    func testEncodingFailure_throws_EncodingError() throws {
+        let mockOrderId = "mockOrderId"
```
jaxdesmarais: nit:
```suggestion
        let mockOrderID = "mockOrderID"
```

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -9,15 +9,21 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+    var jsonEncoder: JSONEncoder?
     
     /// Creates a request to attach a payment source to a specific order.
     /// In order to use this initializer, the `paymentSource` parameter has to
     /// contain the entire dictionary as it exists underneath the `payment_source` key.
     init(
         accessToken: String,
-        cardRequest: CardRequest
+        cardRequest: CardRequest,
+        encoder: JSONEncoder? = nil
```
jaxdesmarais: The other option for this is to create a second init for testing with the JSONEncoder. I think we do this in other files to make it super clear which one is for testing.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -9,15 +9,21 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
     ) throws {
+        if encoder != nil {
+            self.jsonEncoder = encoder
+        } else {
+            self.jsonEncoder = JSONEncoder()
+        }
```
jaxdesmarais: I think we could do the following here instead:
```suggestion
        self.jsonEncoder = encoder ?? JSONEncoder()
```

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: That makes sense

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: addressed in commit 4de094c2c79d020a4e248ad4c5677b8d55c02d2e

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: addressed in [commit](4de094c2c79d020a4e248ad4c5677b8d55c02d2e)

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: addressed in [commit](4de094c2c79d020a4e248ad4c5677b8d55c02d2e)

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: addressed in [commit](4de094c2c79d020a4e248ad4c5677b8d55c02d2e)

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -35,13 +36,52 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
         // TODO - The complexity in this `init` signals to reconsider our use/design of the `APIRequest` protocol.
         // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
     }
     
+    /// For testing purpose
+    init(
+        accessToken: String,
+        cardRequest: CardRequest,
+        encoder: JSONEncoder?
+    ) throws {
```
scannillo: If we want to expose an init for testing, it should really just wrap existing functionality. This duplicates everything in the original init. We don't want duplicate code just used for testing, since this won't actually test the actual code a merchant app uses.

KunJeongPark: I am not sure how we would do this since the other initializer instantiates JSONEncoder.
I can't just call the original initializer from the test initializer. 

Originally, I had single initializer with optional encoder parameter and toggled assigning jsonEncoder property depending on whether encoder parameter was provided.

scannillo: We do this extensively throughout the SDK, injecting mocks via dependency injection. [Here is an example](https://github.com/paypal/iOS-SDK/blob/2d2b08f14ac487da68cb1f1278cfac3c17c45dc6/Sources/CorePayments/Networking/HTTP.swift#L11-L15) using default params. [Here is an example](https://github.com/paypal/iOS-SDK/blob/2d2b08f14ac487da68cb1f1278cfac3c17c45dc6/Sources/CardPayments/CardClient.swift#L16-L29) exposing 2 inits.

KunJeongPark: Fixed in this commit. Added default JSONEncoder() in initializer. be9e5acdfbd4a8c9209de3cd5ba073f4f76ff494 and deleted "for testing purposes" comment in next commit efac2b7ecd532f13e91a44c9f2b2436b9957969a.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -36,12 +40,15 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+        do {
+            body = try jsonEncoder.encode(confirmPaymentSource)
+        } catch {
+            throw CardClientError.encodingError
+        }
         
         // TODO - The complexity in this `init` signals to reconsider our use/design of the `APIRequest` protocol.
         // Existing pattern doesn't provide clear, testable interface for encoding JSON POST bodies.
     }
-
```
jaxdesmarais: I think we want to keep this line break

KunJeongPark: addressed in this commit 6409a89f4d28cb87a1025b5f579f2e4e4a2e1210

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -73,4 +73,39 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
+            securityCode: "222"
+        )
+        let cardRequest = CardRequest(orderID: mockOrderID, card: card)
+        
+        let failingEncoder = FailingJSONEncoder()
+        
+        XCTAssertThrowsError(try ConfirmPaymentSourceRequest(
+            accessToken: "fake token",
+            cardRequest: cardRequest,
+            encoder: failingEncoder)) { error in
```
jaxdesmarais: I believe this spacing should throw some sort of warning. We should update it to:
```swift
        XCTAssertThrowsError(
            try ConfirmPaymentSourceRequest(
                accessToken: "fake token",
                cardRequest: cardRequest,
                encoder: failingEncoder
            )
        ) { error in
```

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -73,4 +73,39 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
+    enum TestError: Error {
+        case encodeError
+    }
+    
+    class FailingJSONEncoder: JSONEncoder {
+        
+        override func encode<T>(_ value: T) throws -> Data where T: Encodable {
+            throw TestError.encodeError
+        }
+    }
```
scannillo: Nitpick - I would move this all into a `FailingJSONEncoder.swift` file in this test module

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
scannillo: 👏 That looks awesome!

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -9,15 +9,19 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
     /// In order to use this initializer, the `paymentSource` parameter has to
     /// contain the entire dictionary as it exists underneath the `payment_source` key.
+    
     init(
         accessToken: String,
-        cardRequest: CardRequest
+        cardRequest: CardRequest,
+        encoder: JSONEncoder = JSONEncoder()
     ) throws {
+        // encode with custom encoder that throws error if passed in
```
scannillo:
```suggestion
```

Don't think this comment is necessary. What would be more appropriate is adding a comment next to the default param like

```
        encoder: JSONEncoder = JSONEncoder() // exposed for test injection
```

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: Thank you Sammy.

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: Addressed here: 7f57a8067d0624219df2ff69d2edf49deaf397c5

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: Addressed here: 7f57a8067d0624219df2ff69d2edf49deaf397c5
I put it under Test Shared and imported that module.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: Addressed here: 7f57a8067d0624219df2ff69d2edf49deaf397c5

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -73,4 +74,30 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
+        )
+        let cardRequest = CardRequest(orderID: mockOrderID, card: card)
+        
+        let failingEncoder = FailingJSONEncoder()
+        
+        XCTAssertThrowsError(
+            try ConfirmPaymentSourceRequest(
+            accessToken: "fake token",
+            cardRequest: cardRequest,
+            encoder: failingEncoder)
```
jaxdesmarais:
```suggestion
            try ConfirmPaymentSourceRequest(
                accessToken: "fake token",
                cardRequest: cardRequest,
                encoder: failingEncoder
            )
```

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
scannillo: Awesome! And thank you for linking to the commit, super easy to find and read

scannillo (approved):

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: addressed in 8276e5e070ada929f494281e93582e54d75475a8

PR #157, opened 2023-06-02
Add `HTTPResponseParser` & `HTTPResponse` types

scannillo:
### Reason for changes

Screenshot 2023-06-02 at 1 57 28 PM

In the diagram above, the green lines represent the actual dependency graph. 

Due our `HTTP` [layer returning ](https://github.com/paypal/iOS-SDK/blob/17eb5bd1658fcc9d313a0e322380e9be9a8387de/Sources/CorePayments/Networking/HTTP.swift#L21)the [`T.ResponseType`](https://github.com/paypal/iOS-SDK/blob/17eb5bd1658fcc9d313a0e322380e9be9a8387de/Sources/CorePayments/Networking/Protocols/APIRequest.swift#L4), we were forced to mock URLSession to mimic the `HTTP` responses we needed. 

This PR adds a `HTTPResponse` and `HTTPResponseParser` layer (similar to Android) to allow us to actually mock `HTTP` responses in isolation.

### Main changes
- Update `HTTP` layer to vend `HTTPResponse` instead of generic `T.ResponseType`
    -  `performRequest<T: APIRequest>(_ request: T) async throws -> (T.ResponseType)` --> `performRequest(_ request: any APIRequest) async throws -> HTTPResponse`
    - Allows us to mock `HTTP` without needing to also mock `URLSession`. Returning the generic `T.ResponseType` made mocking really hard because a test file cannot instantiate this type.
    - Sets us up to remove `APIClient` in favor of `PpassAPI` types[ like Android](https://github.com/paypal/Android-SDK/pull/138)
- Move response parsing logic out of `HTTP` and into `HTTPResponseParser`
    - Isolates `HTTP`s responsibility solely to networking

### Other changes
- Delete `APIClientDecoder` file
- Add `HTTPResponse` & `HTTPResponseParser` files
- Rename `dataParsingError` --> `jsonDecodingError` for clarity
- Remove client ID caching
    - It made this PR a lot easier, and I know we will be doing it anyway

### Next Steps
- Refactor GraphQL layer to use `HTTP` instead of it's own `URLSession` instance
- Remove `APIClient` in favor of `PpassAPI` classes that house request creation, fetch, and response parsing

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/CorePayments/Networking/HTTPResponse.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+struct HTTPResponse {
+    
+    let status: Int
+    let body: Data?
+    
+    var isSuccessful: Bool { (200...300).contains(status) }
```
jaxdesmarais: 🥳  this makes it so much clearer and easier to add response parsing as the networking layer grows

jaxdesmarais (approved):
👏  this makes the HTTP layer so much clearer already along with the tests being way easier to understand. 🚀

Sources/CorePayments/Networking/HTTPResponse.swift
sshropshire:
```suggestion
    var isSuccessful: Bool { (200..<300).contains(status) }
```
Slight 🧶 we probably won't get many 300s but this will make sure to keep [HTTP 300](https://httpstatusdogs.com/300-multiple-choices) out of the successful response range.

sshropshire (approved):
🌊 🙌

Sources/CorePayments/Networking/HTTPResponse.swift
scannillo: Good catch! Updated in this commit (634b59a12de8878414ba6c892723b9fb0350c71d) as well as added unit tests for the `isSuccessful` bool.

PR #158, opened 2023-06-02
Flatten GraphQL project files

scannillo:
### Reason for changes

- The GraphQL files were nested more than necessary. This PR flattens them for easier visualization as we prep for a larger GraphQL-layer refactor to utilize the same underlying `HTTP` as REST calls
- This not the ideal end state, just a temp cleanup 

### Summary of changes
- Remove unused `APIResult.swift`
- Consolidate model/struct type files where possible
- Clean up file structure
### Photos
| Before      | After |
| ----------- | ----------- |
| Screenshot 2023-06-02 at 4 22 44 PM      | Screenshot 2023-06-02 at 4 22 09 PM       |

### Checklist

- [ ] Added a changelog entry

-


Sources/CorePayments/Networking/GraphQL/GraphQLError.swift
```diff
@@ -3,3 +3,8 @@ public struct GraphQLError: Codable, Error {
     let message: String
     let extensions: [Extension]?
 }
+
+struct Extension: Codable {
+
+    let correlationId: String
```
jaxdesmarais, line 9:
```suggestion
    let correlationID: String
```

sshropshire (approved):

Sources/CorePayments/Networking/GraphQL/GraphQLError.swift
scannillo: Gonna leave this for a separate PR, because it might affect the JSON parsing when converting from snake case & we don't have tests for it right now to cover it.

jaxdesmarais (approved):
I am like 6 years late here, didn't realize I didn't actually approve this ever

PR #161, opened 2023-06-12
Return Deep Link URL from 3DS Verification

KunJeongPark:
### Summary of changes

- Added url from 3DS authentication to CardResult returned to the merchant in notifySuccess

### Checklist

- [x] Added a changelog entry

- @KunJeongPark @sshropshire

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -50,9 +50,7 @@ public class CardClient: NSObject {
                 } else {
                     analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:succeeded")
                     
-                    let cardResult = CardResult(
-                        orderID: result.id
-                    )
+                    let cardResult = CardResult(orderID: result.id, deepLinkURL: nil)
                     notifySuccess(for: cardResult)
                 }
             } catch let error as CoreSDKError {
@@ -86,7 +84,7 @@ public class CardClient: NSObject {
                     self?.analyticsService?.sendEvent("card-payments:3ds:challenge-presentation:failed")
                 }
             },
-            sessionDidComplete: { _, error in
+            sessionDidComplete: { url, error in
                 self.delegate?.cardThreeDSecureDidFinish(self)
                 if let error = error {
                     switch error {
@@ -99,9 +97,7 @@ public class CardClient: NSObject {
                     }
                 }
                 
-                let cardResult = CardResult(
-                    orderID: orderId
-                )
+                let cardResult = CardResult(orderID: orderId, deepLinkURL: url)
                 self.notifySuccess(for: cardResult)
             }
         )
--- a/Sources/CardPayments/Models/CardResult.swift
+++ b/Sources/CardPayments/Models/CardResult.swift
@@ -9,6 +9,8 @@ public struct CardResult {
     /// The order ID associated with the transaction
     public let orderID: String
 
-    ///  The deep link url returned from 3DS authentication
+    //// :nodoc: This is the deep link url returned from 3DS authentication
+    ///
+    // TODO: parse contents of this URL once we are clear on values returned from 3ds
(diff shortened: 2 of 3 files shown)


Sources/CardPayments/Models/CardResult.swift
```diff
@@ -8,4 +8,7 @@ public struct CardResult {
 
     /// The order ID associated with the transaction
     public let orderID: String
+
+    ///  The deep link url returned from 3DS authentication
+    public let deepLinkURL: URL?
```
scannillo, line 14: I know Android exposes this - but I am curious what value it gives to the merchant?

jaxdesmarais: ➕ was just typing up the same. How do we expect merchants to use this?

sshropshire: For this, the format is inconsistent from what I've seen. We return it as an unparsed string so the Demo app can display it in the UI. It'll help us debug the return URL until we can figure out if it has a set schema definition somewhere that we could use to write tests against.

KunJeongPark: I think it's a temporary solution. I'm in conversation with HeliosNodeWeb team to make sure we are getting the values we need from the return url.

jaxdesmarais: If it's for internal use/temporary we should remove it from our CHANGELOG and add a `nodoc` for it. And likely add a `TODO` so we don't forget to come back to removing it once we can parse the URL into an expected format.

KunJeongPark: Addressed in 5d58df71bc2bbc65604e50fcafe0dcffdd8c5dc6

jaxdesmarais (approved):

Sources/CardPayments/CardClient.swift
```diff
@@ -50,9 +50,7 @@ public class CardClient: NSObject {
                 } else {
                     analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:succeeded")
                     
-                    let cardResult = CardResult(
-                        orderID: result.id
-                    )
+                    let cardResult = CardResult(orderID: result.id, deepLinkURL: nil)
```
KunJeongPark, line 53: We want to be able to return order status here, right? I think the issue was that after 3ds challenge, we didn't have updated order status so we just deleted that field in CardResult for now. I think we definitely want to add that order status field back, and assume some status, or return "UNKNOWN" or something after 3DS challenge without error. Should we add a TODO here?

I think the top comment above approveOrder needs to be updated. We are not returning payment source to the merchant anymore.

jaxdesmarais: Looks like both `paymentSource` and `status` was removed on the [Android PR](https://github.com/paypal/Android-SDK/pull/146). We should do the same on iOS so the platforms are aligned. There was a conversation here about this decision: https://github.com/paypal/Android-SDK/pull/146#discussion_r1210503077

jaxdesmarais: Agree, we should always update docstrings to reflect the current status of what our SDK is returning

scannillo (approved):

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -8,4 +8,8 @@ public struct CardResult {
 
     /// The order ID associated with the transaction
     public let orderID: String
+
+    //// :nodoc: This is the deep link url returned from 3DS authentication
+    // TODO: parse contents of this URL once we are clear on values returned from 3ds
+    public let deepLinkURL: URL?
```
scannillo, line 14: We'll want to update this repo to use the new `@_documentation(visibility: private)` syntax introduced in Xcode 14.3 (see BT PR - https://github.com/braintree/braintree_ios/pull/1034)

PR #162, opened 2023-06-12
Vaulting with Purchase

KunJeongPark:
### Summary of changes
- add optional `Vault` object with optional `customerID` of String type in `CardRequest`
- move `Attributes` from Card to new `EncodedCard` struct that formats contents of `CardRequest `
  for `confirm-payment-source` Orders API request. 

### Checklist

- [x] Added a changelog entry

- @KunJeongPark 
 <p float="left">
 <image src = "https://github.com/paypal/iOS-SDK/assets/9273272/0f516f8f-0913-4f7a-b1be-1b1a9f2f0df4" width="400" />
 <image src ="https://github.com/paypal/iOS-SDK/assets/9273272/5b0e02af-e55b-43d3-ad89-f8a5db198a10" width="400" />
</p>

### No Vault Request:
**Method**: POST **URL**: https://api.sandbox.paypal.com/v2/checkout/orders/0EL07199CK987522C/confirm-payment-source
**Body**:
```
{
  "application_context": {
    "return_url": "sdk.ios.paypal://card/success",
    "cancel_url": "sdk.ios.paypal://card/cancel"
  },
  "payment_source": {
    "card": {
      "number": "4344511979394247",
      "security_code": "123",
      "billing_address": null,
      "name": null,
      "attributes": {
        "verification": {
          "method": "SCA_ALWAYS"
        }
      },
      "expiry": "2025-01"
    }
  }
}
```

**Response After Authorize**
```
{
  "id": "0EL07199CK987522C",
  "status": "COMPLETED",
  "payment_source": {
    "card": {
      "last_digits": "4247",
      "expiry": "2025-01",
      "brand": "VISA",
      "type": "UNKNOWN"
    }
  }
```

### Vault with Purchase Request:
**Method**: POST **URL**: https://api.sandbox.paypal.com/v2/checkout/orders/3FB45835KB971130X/confirm-payment-source

**Body**:
```
{
  "application_context": {
    "return_url": "sdk.ios.paypal://card/success",
    "cancel_url": "sdk.ios.paypal://card/cancel"
  },
  "payment_source": {
    "card": {
      "number": "4344511979394247",
      "security_code": "123",
      "billing_address": null,
      "name": null,
      "attributes": {
        "vault": {
          "store_in_vault": "ON_SUCCESS"
        },
        "verification": {
          "method": "SCA_ALWAYS"
        },
        "customer": {
          "id": "test567"
        }
      },
      "expiry": "2025-01"
    }
  }
}

```
**Response After Authorize**:
```
{
  "id": "3FB45835KB971130X",
  "status": "COMPLETED",
  "payment_source": {
    "card": {
      "last_digits": "4247",
      "expiry": "2025-01",
      "brand": "VISA",
      "type": "UNKNOWN",
      "attributes": {
        "vault": {
          "id": "3fb3mfxf",
          "status": "VAULTED",
          "customer": {
            "id": "test567"
          }..
```


UnitTests/CardPaymentsTests/Card_Tests.swift
```diff
@@ -35,4 +37,82 @@ class Card_Tests: XCTestCase {
...
+        // swiftlint:enable line_length
+        
+        let expectedCarData = expectedCardJSON.data(using: .utf8)!
+        
+        
+        let actualCardDict = try JSONSerialization.jsonObject(with: encodedCard, options: []) as! [String: Any]
+        let expectedCardDict = try JSONSerialization.jsonObject(with: expectedCarData, options: []) as! [String: Any]
+        
+        XCTAssertEqual(actualCardDict as NSDictionary, expectedCardDict as NSDictionary)
+    }
```
KunJeongPark: This one was strange. For some reason, adding the customer field in Attributes made the test fail on just comparing json strings because attribute properties got encoded in different order as soon as I added customer field, causing the test to fail. I opted to just compare the dictionary objects.

Demo/Demo/SwiftUIComponents/CheckBoxView.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+
+struct CheckBoxView: View {
```
jaxdesmarais: Are you able to add a screen shot of what this looks like in the PR description?

KunJeongPark: Sure. This is a draft so I wanted to check with Steven to see if this is what we want.
My textfield for vault customer id only appears if vault option is checked.
I put this here to make it easier for me to test vaulting result but I will put it up.
It'd be helpful to have feedback with WIP

jaxdesmarais: Looking through Apple's [human interface guidelines](https://developer.apple.com/design/human-interface-guidelines/toggles#Best-practices) it seems to indicate that unless used elsewhere in our design the preference is to use toggles when choosing 2 opposing values. It seems like this would be an appropriate use case for a toggle and would follow their guidelines for this type of UI. I think like other platform specific things we should lean into the design guidelines for each platform where possible.

UnitTests/CardPaymentsTests/Card_Tests.swift
KunJeongPark: Simulator Screen Shot - iPhone 14 Pro - 2023-06-14 at 06 12 04
29115f)

Demo/Demo/SwiftUIComponents/CheckBoxView.swift
KunJeongPark: Simulator Screen Shot - iPhone 14 Pro - 2023-06-14 at 06 13 32

KunJeongPark: I saw a checkbox on the Android demo app, so I was trying to replicate with toggle option on the checkbox but I am not tied to one thing or other, whatever is more intuitive works for me. I do think checkbox is prettier but toggle switch is what people are used to on iOS.

jaxdesmarais: Yeah I think with UI we should follow the platform specific guidelines for UI since we don't have a designer for our demo apps.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -23,8 +23,19 @@ struct ConfirmPaymentSourceRequest: APIRequest {
         var confirmPaymentSource = ConfirmPaymentSource()
         var card = cardRequest.card
         let verification = Verification(method: cardRequest.sca.rawValue)
-        card.attributes = Attributes(verification: verification)
+        
+        if cardRequest.shouldVault {
+            var customer: Customer?
+            let vault = Vault(storeInVault: "ON_SUCCESS")
```
jaxdesmarais: Can we extract `ON_SUCCESS` out into an enum with all possible values?

Sources/CardPayments/Models/Card.swift
```diff
@@ -66,10 +66,22 @@ public struct Card: Encodable {
...
+    let vault: Vault?
     let verification: Verification
 }
 
 struct Verification: Codable {
 
     let method: String
 }
+
+struct Customer: Codable {
```
jaxdesmarais: Will we use this struct for other flows or should we consider adding a `customerID` to the Vault object? I am not sure how we plan to grow this object in the future.

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,20 @@ public struct CardRequest {
     /// 3DS authentication launch option
     public let sca: SCA
     
+    public let shouldVault: Bool
+    
+    public let customerID: String?
```
jaxdesmarais: We should add docstrings for these

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,20 @@ public struct CardRequest {
...
+    
     /// Creates an instance of a card request
     /// - Parameters:
     ///    - orderID: The order to be approved
     ///    - card: The card to be charged for this order
     ///    - sca: Specificy to always launch 3DS or only when required. Defaults to `scaWhenRequired`.
```
jaxdesmarais, line 21: Will need to make sure to add these properties to the docstring here as well. I know at one time we talked about having a separate vault object instead of adding it to card but can't remember where we landed on that. @scannillo do you remember?

Sources/CardPayments/Models/Card.swift
KunJeongPark: I wanted to keep it in the same format that the confirm-payment-source request expects in the body.
Sometimes the customer id is not entered by the merchant and PayPal endpoint generates it for the merchant.

```
"card": {
      ..
      "attributes": {
        "vault": {
          "store_in_vault": "ON_SUCCESS"
        },
        "verification": {
          "method": "SCA_ALWAYS"
        },
        "customer": {
          "id": "test654"
        }
      }
```,

File: Sources/CardPayments/Models/CardRequest.swift (hunk 6 above)
Comment (KunJeongPark) [reply Sources/CardPayments/Models/CardRequest.swift:21]: They're both mentioned as possibilities in Jose's research. Steven and I decided to keep it less nested.

File: Sources/CardPayments/Models/CardRequest.swift (hunk 5 above)
Comment (KunJeongPark) [reply Sources/CardPayments/Models/CardRequest.swift]: Addressed in commit eb8ccc19b6d47f49f3cc6bdf5f7f6e18a9871ce4

File: Sources/CardPayments/Models/CardRequest.swift (hunk 6 above)
Comment (KunJeongPark) [reply Sources/CardPayments/Models/CardRequest.swift:21]: DocStrings addressed in commit eb8ccc19b6d47f49f3cc6bdf5f7f6e18a9871ce4

File: Demo/Demo/SwiftUIComponents/CheckBoxView.swift (hunk 2 above)
Comment (KunJeongPark) [reply Demo/Demo/SwiftUIComponents/CheckBoxView.swift]: Addressed in commit a1d9a4758c29e5f82cd6fb3234b7332427433c4f and bc71f8f58d464dca4ba99e7886751515ad94b63d

File: Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift (hunk 3 above)
Comment (KunJeongPark) [reply Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift]: Our API documentation lists just this one option. I will put it in enum just in case they decide to add more.

Comment (KunJeongPark) [reply Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift]: Addressed in commit 06b7171b03cb4363f01007db483c395267394245

File: Demo/Demo/ViewModels/BaseViewModel.swift [hunk 7]
```diff
@@ -102,14 +102,23 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
         let cardClient = CardClient(config: config)
         cardClient.delegate = self
-        let cardRequest = CardRequest(orderID: orderID, card: card, sca: .scaAlways)
+        
+        var passedCustomerID: String?
+        if customerID.isEmpty {
+            passedCustomerID = nil
+        } else {
+            passedCustomerID = customerID
+        }
```
sshropshire:
```suggestion
        let passedCustomerID = customerID.isEmpty ? nil : customerID
```
🧶 : this should be equivalent. It's a bit more concise and eliminates the `var`.

UnitTests/CardPaymentsTests/Card_Tests.swift
sshropshire: We should do this for all JSON assertions to make the tests more stable.

sshropshire (approved):

Demo/Demo/ViewModels/BaseViewModel.swift
KunJeongPark: Addressed in commit 6fd6e1bb5e4234f8e0fc2d77a3ca77670eac6ee3

Demo/Demo/Models/Order.swift
```diff
@@ -1,5 +1,39 @@
 struct Order: Codable {
 
+    enum CodingKeys: String, CodingKey {
+        case id
+        case status
+        case paymentSource = "payment_source"
+    }
```
scannillo, line 7: We don't need custom coding keys if we use the `JSONEncoder.convertToSnakeCase` setting ([see Apple docs](https://developer.apple.com/documentation/foundation/jsonencoder/keyencodingstrategy/converttosnakecase))

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -192,15 +195,25 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
         Task {
             let captureResult = try? await DemoMerchantAPI.sharedService.caputureOrder(orderID: result.orderID)
             let status = captureResult?.status ?? ""
-            updateTitle("Order ID:\(result.orderID) status: \(status)")
+            let vault = captureResult?.paymentSource?.card.attributes?.vault
+            let orderInfo = "Order ID: \(result.orderID) \n Status: \(status)"
+            let vaultInfo = vault != nil ? "Vault Status: \((vault?.status) ?? "") \n Vault ID: \((vault?.id) ?? "")" : ""
+            let customerInfo = vault != nil ? "Customer ID: \((vault?.customer.id ?? ""))" : ""
+            
+            updateTitle("\(orderInfo) \n \(vaultInfo)\n \(customerInfo)")
```
scannillo: This block of code is duplicated b/w the capture & authorize functions. We can consolidate this into a helper function so we don't have to duplicate. Maybe something like `displayResult(orderID: String, vaultID: String, customerID: String)`.

Sources/CardPayments/Models/Card.swift
```diff
@@ -66,10 +66,26 @@ public struct Card: Encodable {
...
+    let id: String
+}
+
+enum StoreInVault: String, Codable {
+    case onSuccess = "ON_SUCCESS"
+}
+
+struct Vault: Codable {
+    
+    let storeInVault: StoreInVault
```
scannillo: Since this will always be `ON_SUCCESS` do we need an enum to store this string? Why not just `let storeInVault = "ON_SUCCESS"`?

Are we planning to add more cases?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -23,8 +23,19 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+            var customer: Customer?
+            let vault = Vault(storeInVault: .onSuccess)
+            if let customerID = cardRequest.customerID {
+                customer = Customer(id: customerID)
+            }
             
+            card.attributes = Attributes(customer: customer, vault: vault, verification: verification)
+        } else {
+            card.attributes = Attributes(customer: nil, vault: nil, verification: verification)
+        }
```
scannillo: I think this could be simplified a bit - see a suggestion I made in this [branch diff](https://github.com/paypal/iOS-SDK/compare/vaulting_with_purchase...vault-wo-purchase-suggestion). Let me know what you think! Feel free to merge that branch into yours directly - the tests will likely need to be updated accordingly.

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,24 @@ public struct CardRequest {
     /// 3DS authentication launch option
     public let sca: SCA
     
+    /// The option to vault with purchase
+    public let shouldVault: Bool
+    
+    /// The customerID for vault with purchase option. If left blank, PayPal endpoint will auto-generate one
```
scannillo: The backing PayPal service will only generate a customerID if `shouldVault` is set to `true`, right? We should clarify this.

UnitTests/CardPaymentsTests/Card_Tests.swift
```diff
@@ -35,4 +37,82 @@ class Card_Tests: XCTestCase {
...
+       
+        let encodedCard = try JSONEncoder().encode(card)
+
+        // swiftlint:disable line_length
+        let expectedCardJSON = """
+        {"number":"4111111111111111","billingAddress":{"admin_area_2":"Test City","addressLine1":"Test Line 1","countryCode":"Test Country","addressLine2":"Test Line 2","admin_area_1":"Test State","postalCode":"Test Zip"},"securityCode":"123","name":"Test Name","attributes":{"customer":{"id":"test_id"},"vault":{"storeInVault":"ON_SUCCESS"},"verification":{"method":"SCA_ALWAYS"}},"expiry":"2031-01"}
+        """
+        // swiftlint:enable line_length
+        
+        let expectedCarData = expectedCardJSON.data(using: .utf8)!
```
scannillo:
```suggestion
        let expectedCardData = expectedCardJSON.data(using: .utf8)!
```

UnitTests/CardPaymentsTests/Card_Tests.swift
scannillo: Yeh, agreed!

scannillo: Also, we can start with `expectedCardJSON` being in a Dictionary format from the start, so we don' have to do the conversion to `Data` -> `JSON` -> `Dict`

UnitTests/CardPaymentsTests/Card_Tests.swift
```diff
@@ -35,4 +37,82 @@ class Card_Tests: XCTestCase {
...
+        )
+        
+        let customer = Customer(id: "test_id")
+        card.attributes = Attributes(
+            customer: customer,
+            vault: Vault(storeInVault: .onSuccess),
+            verification: Verification(
+                method: "SCA_ALWAYS")
+        )
+
```
scannillo:
```suggestion
```

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,24 @@ public struct CardRequest {
     /// 3DS authentication launch option
     public let sca: SCA
     
+    /// The option to vault with purchase
```
scannillo: What does "vault with purchase" mean? Can we be a bit more descriptive?

Sources/CardPayments/Models/CardRequest.swift
KunJeongPark: I am not sure if it is internally generated upon capture/authorize without vault option, but vault field is not returned in capture/authorize result if shouldVault in request is false,

Sources/CardPayments/Models/Card.swift
KunJeongPark: As we confirmed in stand, for vault with purchase, we are only using ON_SUCCESS value but there are other values we may need later.

Demo/Demo/Models/Order.swift
KunJeongPark: I ran into trouble when decoding when I removed the coding keys.

PayPal.xcodeproj/project.pbxproj
```diff
@@ -1568,6 +1576,7 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
+				MACOSX_DEPLOYMENT_TARGET = "$(RECOMMENDED_MACOSX_DEPLOYMENT_TARGET)";
```
KunJeongPark: I am not really understanding these large diffs in pbxproj file. I did update to Xcode 14.3.1 last week but didn't see all these changes before. I was prompted one time in Xcode to make some recommended changes to file settings but I reverted those changes. I initially had a lot of problems with running unit tests but I resolved it by unchecking TestShared as target in PayPalNativePaysheetActions. I think these changes in pbxproj file are from Xcode upgrade but I wanted to make sure that these don't  cause problems in pipeline.

UnitTests/CardPaymentsTests/Card_Tests.swift
KunJeongPark: Addressed in commit 6057b4e6155b57c8c2fa4f361c0410eb0f29f1e3 for ServiceCard_Tests.swift. I didn't think we need test for a simple Card struct that is not encoded directly.

Demo/Demo/ViewModels/BaseViewModel.swift
KunJeongPark: Addressed in commit 6057b4e6155b57c8c2fa4f361c0410eb0f29f1e3

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: Addressed in commit 6057b4e6155b57c8c2fa4f361c0410eb0f29f1e3 Modified it a bit.

Sources/CardPayments/Models/CardRequest.swift
KunJeongPark: Addressed in commit 6057b4e6155b57c8c2fa4f361c0410eb0f29f1e3

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,24 @@ public struct CardRequest {
...
+    public let shouldVault: Bool
```
jaxdesmarais: We should update this PR to match the [implementation decided on Android](https://github.com/paypal/Android-SDK/pull/157#discussion_r1263037902). So instead of 2 separate properties, we want a [Vault object](https://github.com/paypal/Android-SDK/blob/main/CardPayments/src/main/java/com/paypal/android/cardpayments/Vault.kt) with an optional `customerID`.

CHANGELOG.md
```diff
@@ -3,6 +3,14 @@
 
 ## unreleased
 * Breaking Changes
+  * Card 
+    * Update `Attributes` field, adding `customer` and `vault` properties
+      `customer` is of optional `Customer` type, containing property `id` of String type
+      `vault` is of optional `Vault` type, containing `storeInVault` field of `StoreInVault` enum type
+       with one value, `.onSuccess`
```
jaxdesmarais: The `Attributes` object is internal correct? If so we do not need a changelog entry for it.

Sources/CardPayments/Models/CardRequest.swift
jaxdesmarais: Ope, I think you already made this change and my GH was out of date for some reason.

CHANGELOG.md
```diff
@@ -2,8 +2,14 @@
...
 
 ## unreleased
+* Breaking Changes
+  * CardPayments
+    * Add `vault`, optional `Vault` type that contains `customerID`, an optional String property  
 * PayPalNativePayments
   * Bump `PayPalCheckout` to `1.0.0`
+* CardPayments
+    * Remove `attributes` from `Card`
+    * Add `ServiceCard` and added `attributes`
```
jaxdesmarais: This is internal so doesn't need to go in the CHANGELOG.

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -184,11 +190,20 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
             let authorizeResult = try? await DemoMerchantAPI.sharedService.authorizeOrder(
                 orderID: result.orderID, selectedMerchantIntegration: selectedMerchantIntegration
             )
-            let status = authorizeResult?.status ?? ""
-            updateTitle("Order ID:\(result.orderID) status: \(status)")
+            displayResult(cardResult: result, orderResult: authorizeResult)
         }
     }
     
+    func displayResult(cardResult: CardResult, orderResult: Order?) {
```
jaxdesmarais: since the first property in the method signature is result I think we can do:
```suggestion
    func displayResult(_ cardResult: CardResult, orderResult: Order?) {
    // OR
    func display(_ cardResult: CardResult, orderResult: Order?) {
```

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,19 @@ public struct CardRequest {
...
     public let sca: SCA
     
+    /// Vault is a struct with optional customerID that indicates option to vault a card in checkout
+    public let vault: Vault?
+    
     /// Creates an instance of a card request
     /// - Parameters:
     ///    - orderID: The order to be approved
     ///    - card: The card to be charged for this order
     ///    - sca: Specificy to always launch 3DS or only when required. Defaults to `scaWhenRequired`.
```
jaxdesmarais, line 21: We want to add `vault` to this init's parameters docstring

Sources/CardPayments/Models/ServiceCard.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+
+/// ServiceCard is used for formatting contents of `CardRequest` for `ConfirmPaymentSourceRequest`
+struct ServiceCard: Encodable {
+
```
KunJeongPark: I've thought about whether to make this struct as part of ConfirmPaymentSourceRequest struct since it is only used to format Card portion of the body of the ConfirmpaymentSourceRequest but separated it out since it is long. Maybe separating it out in Models folder is confusing since it is only used internally?

Sources/CardPayments/Models/ServiceCard.swift
```diff
@@ -0,0 +1,79 @@
+import Foundation
+
+/// ServiceCard is used for formatting contents of `CardRequest` for `ConfirmPaymentSourceRequest`
+struct ServiceCard: Encodable {
```
jaxdesmarais: Is the naming of this supposed to be similar to a factory class on Android? I think generally we would use `Card` on iOS but curious to hear @scannillo thoughts as naming of these classes/structs is not something we've discussed much on iOS from an architecture point of view.

Sources/CardPayments/Models/ServiceCard.swift
```diff
@@ -0,0 +1,79 @@
...
+
+struct Attributes: Codable {
+
+    var customer: Customer?
+    var vault: CardVault?
+    let verification: Verification
+    
+    init(vault: Vault? = nil, verificationMethod: String) {
+        self.verification = Verification(method: verificationMethod)
+        if let vault = vault {
```
jaxdesmarais:
```suggestion
        if let vault {
```

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+public struct Vault {
+    
+    public let customerID: String?
+    
+    public init(customerID: String? = nil) {
+
```
jaxdesmarais:
```suggestion
```

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
```diff
@@ -15,38 +15,170 @@ class ConfirmPaymentSourceRequest_Tests: XCTestCase {
...
+            "application_context": [
+                "return_url": "sdk.ios.paypal://card/success",
+                "cancel_url": "sdk.ios.paypal://card/cancel"
+            ],
+            "payment_source": [
+                "card": [
+                    "number": "4032036247327321",
+                    "security_code": "222",
+                    "billing_address": NSNull(),
+                    "name": NSNull(),
```
jaxdesmarais: Is there a reason we need to use `NSNull` here instead of `nil`?

jaxdesmarais: Ah, I see I think this line just needs to be updated to an optional `Any`:
```swift
let expectedPaymentSourceDict: [String: Any] = [
// to
let expectedPaymentSourceDict: [String: Any?] = [
```

Sources/CardPayments/Models/ServiceCard.swift
KunJeongPark: I have `Card` as merchant facing class that just has fields that the customer enters. Originally this class had `Attributes` which I found inconvenient because of having to reflect exactly the format of API call. `ServiceCard` exists simply for the flexibility of formatting it for confirm-payment-source Orders API request and incorporating contents of the `CardRequest` which now includes the `vault` object. 

I am now thinking that `ServiceCard` should be part of `ConfirmPaymentSourceRequest` since it is only used to format `Card` for the request.

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: Because I am converting it to NSDictionary to be able to do equality on nested dictionaries, I thought I needed to use NSNull(). I think I got syntax error trying it with `nil` initially.

Sources/CardPayments/Models/ServiceCard.swift
KunJeongPark: I am looking at original Android refactor PR for refactor for parity with Android, removing the general APIClient class, having separate API classes for individual APIRequests that include formatting, making requests and parsing results. I thought that would be too long for this PR which was started a while ago. I think the plan was for that refactor to go in before next phase of vaulting.

CHANGELOG.md
KunJeongPark: Addressed in commit b36e66bd4cb8409036e00132543e30d8d21b6119

Sources/CardPayments/Models/ServiceCard.swift
KunJeongPark: Addressed in commit b36e66bd4cb8409036e00132543e30d8d21b6119

Sources/CardPayments/Models/Vault.swift
KunJeongPark: Addressed in commit b36e66bd4cb8409036e00132543e30d8d21b6119

Demo/Demo/ViewModels/BaseViewModel.swift
KunJeongPark: I thought it might be clarifying since there are two parameters with results. Addressed in commit 6bbd03f18af9ba2c3e8afb67ca2178f233be8734

CHANGELOG.md
```diff
@@ -2,8 +2,11 @@
 # PayPal iOS SDK Release Notes
 
 ## unreleased
+* Breaking Changes
+  * CardPayments
+    * Add `vault`property in `CardRequest`, optional `Vault` type that contains `customerID`, an optional String property  
 * PayPalNativePayments
-  * Bump `PayPalCheckout` to `1.0.0`
+  * Bump `PayPalCheckout` to `1.0.0`
```
jaxdesmarais: We can revert this additional space

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -83,14 +83,21 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
 
     func checkoutWith(
         card: Card,
-        orderID: String
+        orderID: String,
+        shouldVault: Bool = false,
+        customerID: String = ""
```
jaxdesmarais: I think we can make this optional and default to nil so we can remove `passedCustomerID` below. In `SwiftUICardDemo` we'd want to make this change at the callsite instead: 
```swift
await baseViewModel.checkoutWith(
    card: card,
    orderID: orderID,
    shouldVault: shouldVaultSelected,
    customerID: vaultCustomerID.isEmpty ? nil : vaultCustomerID
)
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -83,14 +83,21 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
...
             return
         }
         let cardClient = CardClient(config: config)
         cardClient.delegate = self
-        let cardRequest = CardRequest(orderID: orderID, card: card, sca: .scaAlways)
+        let passedCustomerID = customerID.isEmpty ? nil : customerID
+        var vault: Vault?
+        if shouldVault {
+            vault = Vault(customerID: passedCustomerID)
+        }
```
```suggestion
        let vault: Vault? = shouldVault ? Vault(customerID: passedCustomerID) : nil
```

Demo/Demo/ViewModels/BaseViewModel.swift
```diff
@@ -184,11 +190,20 @@ class BaseViewModel: ObservableObject, PayPalWebCheckoutDelegate, CardDelegate {
             let authorizeResult = try? await DemoMerchantAPI.sharedService.authorizeOrder(
                 orderID: result.orderID, selectedMerchantIntegration: selectedMerchantIntegration
             )
-            let status = authorizeResult?.status ?? ""
-            updateTitle("Order ID:\(result.orderID) status: \(status)")
+            displayResult(result, orderResult: authorizeResult)
         }
     }
     
+    func displayResult(_ cardResult: CardResult, orderResult: Order?) {
```
jaxdesmarais: can this be private?

Sources/CardPayments/Models/CardRequest.swift
```diff
@@ -11,14 +11,19 @@ public struct CardRequest {
     /// 3DS authentication launch option
     public let sca: SCA
     
+    /// Vault is a struct with optional customerID that indicates option to vault a card in checkout
```
jaxdesmarais:
```suggestion
    /// The details used when vaulting a payment method 
```

Sources/CardPayments/Models/ServiceCard.swift
jaxdesmarais: I'm not sure I full understand the reasoning for moving this out of the card model.

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+public struct Vault {
+    
+    public let customerID: String?
+
+    public init(customerID: String? = nil) {
```
jaxdesmarais: These will all need docstrings

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
jaxdesmarais: You shouldn't get a syntax error. Just tested and using `nil` only requires updating `as [String: Any]` to `as [String: Any?]` since with `nil` the value in the dictionary is now optional.

Sources/CardPayments/Models/ServiceCard.swift
KunJeongPark: The Card struct with Encodable was used to get Card info from the customer and also included as part of  CardRequest which now accepts optional vault object to form a request body for confirm-payment-source endpoint. 

The Card struct to collect data has no need for attribute section. I created ServiceCard to be able to format the request for the endpoint. The request body needs the SCA and vault object to create attributes portion as part of Card section, which goes under payment_source. 

```
{
  "application_context": {
    "return_url": "sdk.ios.paypal://card/success",
    "cancel_url": "sdk.ios.paypal://card/cancel"
  },
  "payment_source": {
    "card": {
      "number": "4344511979394247",
      "security_code": "123",
      "billing_address": null,
      "name": null,
      "attributes": {
        "vault": {
          "store_in_vault": "ON_SUCCESS"
        },
        "verification": {
          "method": "SCA_ALWAYS"
        },
        "customer": {
          "id": "test567"
        }
      },
      "expiry": "2025-01"
    }
  }
}

```

KunJeongPark: It was much cleaner to construct the request body from a struct. It made sense for me to decouple collecting card info from customer and exact format needed to create an API request.

Demo/Demo/ViewModels/BaseViewModel.swift
sshropshire:
```suggestion
        let vault = shouldVault ? Vault(customerID: passedCustomerID) : nil
```
We could also get rid of the explicit type declaration here.

Sources/CardPayments/Models/Address.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 
 /// The postal address.
-public struct Address: Codable {
+public struct Address: Codable, Equatable {
```
sshropshire, line 4: q: curious is `Equatable` conformance needed here?

Sources/CardPayments/Models/ServiceCard.swift
sshropshire: 2 cents from Android:

- The PayPal API has a lot of nested JSON structures that don't map nicely to `Codable`
- We had a noticeable amount of class bloat associated with making requests to the sample merchant server
-  Working with dictionaries and converting to JSON may be a better approach to give us more flexibility in the Demo app
- We do end up duplicating a lot of types to make a simple JSON request. Type safety is nice overall, but in this case it has felt like unnecessary boilerplate.

KunJeongPark: In Swift, I've seen use of structs or class rather than raw dictionaries. I think it seems safer, less error prone and easier to read.

sshropshire: Yeah I agree on the less error prone part no doubt. I guess JavaScript has the advantage here since JSON is basically `Codable` for JS.

sshropshire: spicy take: I like `Codable` for the SDK but not as much for the Demo app 😅

Sources/CardPayments/Models/Address.swift
KunJeongPark: I do comparison of the Addresses in a unit test and I didn't want
to have to compare each components of the address.

sshropshire: I think you may be able to declare an extension like

```swift
extension Address: Equatable {}
```
in the unit test if needed. May be OD, but it avoids modifying the `Address` struct that gets shipped with the SDK.

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
KunJeongPark: addressed in commit 552e6ef914120af69d81f61a2c023fe85a62f809

CHANGELOG.md
KunJeongPark: addressed in commit 552e6ef914120af69d81f61a2c023fe85a62f809

Demo/Demo/ViewModels/BaseViewModel.swift
KunJeongPark: addressed in commit 552e6ef914120af69d81f61a2c023fe85a62f809

Demo/Demo/ViewModels/BaseViewModel.swift
KunJeongPark: addressed in commit 552e6ef914120af69d81f61a2c023fe85a62f809

Sources/CardPayments/Models/CardRequest.swift
KunJeongPark: addressed in commit 552e6ef914120af69d81f61a2c023fe85a62f809

Sources/CardPayments/Models/Vault.swift
KunJeongPark: Addressed 730bb4d09f4bebf0dc15e83998a6bafe30144c2b

Sources/CardPayments/Models/Address.swift
KunJeongPark: I was thinking it might be something that can be done commonly. Like is this old address same as new address, etc.

sshropshire: Yeah it's probably fine. I'm thinking of a dev best practice I saw a while back about how test code can influence production code. [This article](https://8thlight.com/insights/dont-mix-test-code-with-production-code) holds a similar sentiment. But like I said i'm being OD lol. Total take or leave from me.

PayPal.xcodeproj/project.pbxproj
```diff
@@ -2573,7 +2580,7 @@
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
 				DRIVERKIT_DEPLOYMENT_TARGET = 19.0;
-				EMBEDDED_CONTENT_CONTAINS_SWIFT = YES;
+                EMBEDDED_CONTENT_CONTAINS_SWIFT = YES;
```
jaxdesmarais: Looks like this spacing got a bit wonky

jaxdesmarais: There are several similar ones below where the spacing is off

Sources/CardPayments/Models/CardRequest.swift
jaxdesmarais: Looks like this was changed in the init but not here

Sources/CardPayments/Models/ServiceCard.swift
jaxdesmarais: I am fine if it's the path we want to go down for decoupling the merchant facing pieces vs request building pieces. I'm not opposed to this as a solution for that but want to ensure:
1. That we have a name convention decided on - I would want @scannillo to weigh in on this as it will determine the architecture of the SDK and this is a divergence from what we've done previously
2. A plan to tackle the other models since right now we would be leaving some in the current pattern and moving card to this new architecture pattern

I understand the reasoning and the differences in the merchant facing part of the SDK vs the API, just want to make sure that we are following a common architecture pattern across the entire SDK.

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+/// This class is used to configure a CardRequest for vaulting
+/// CustomerID is optional. When set, this property will be used to associate  a payment method with vault of customer with this ID
+/// If customerID is not specified, one will be generated by the PayPal API
+public struct Vault {
+    
+    public let customerID: String?
+
+    public init(customerID: String? = nil) {
```
jaxdesmarais: These are still missing docstrings

PayPal.xcodeproj/project.pbxproj
KunJeongPark, line 2583: I have no idea why this diff shows up like this. These lines look identical in Xcode and I've even tried copying same line from main and it shows no diff from what I have.

UnitTests/CardPaymentsTests/ConfirmPaymentSourceRequest_Tests.swift
jaxdesmarais: Looks like not all of these were captured in that commit update, we should also update: `testEncodingPaymentSource_withValidCard_andVaultWithPurchase_nilCustomerID` and `testEncodingPaymentSource_withValidCard_andVaultWithPurchase_noCustomerID`.

PayPal.xcodeproj/project.pbxproj
KunJeongPark: Fixed it with TextEditor. Thanks to Steven. Xcode was inserting spaces for some reason.

Sources/CardPayments/Models/Vault.swift
KunJeongPark: Isn't this it? 
/// This class is used to configure a CardRequest for vaulting
/// CustomerID is optional. When set, this property will be used to associate  a payment method with vault of customer with this ID
/// If customerID is not specified, one will be generated by the PayPal API

jaxdesmarais: That's for the struct, these public declarations also need docstrings:
```
public let customerID: String?

public init(customerID: String? = nil)
```
See [`cardRequest`](https://github.com/paypal/iOS-SDK/blob/main/Sources/CardPayments/Models/CardRequest.swift) as an example - though that is missing a docstring for the top level. We should create a follow up ticket to ensure everything public is documented properly, but in this PR we should ensure everything new is documented as expected.

KunJeongPark: addressed in commit 02c21178d2f5a602e9592d8c5c6a57c3239c0d06

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 # PayPal iOS SDK Release Notes
 
 ## unreleased
+* Breaking Changes
+  * CardPayments
+    * Add `vault`property in `CardRequest`, optional `Vault` type that contains `customerID`, an optional String property
```
jaxdesmarais:
```suggestion
    * Add `vault` property in `CardRequest`, this type that contains a `customerID`, if a customer ID is not provided one will be generated
```

Sources/CardPayments/Models/Card.swift
scannillo: If there's values we _know_ we will add soon, I understand preparing for it. But we try to follow YAGNI and not add things we don't need, since it often muddies up the code.

KunJeongPark: That property is in "ServiceCard" now.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -21,16 +21,15 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+        let serviceCard = ServiceCard(cardRequest: cardRequest)
+        
         confirmPaymentSource.applicationContext = ApplicationContext(
             returnUrl: PayPalCoreConstants.callbackURLScheme + "://card/success",
             cancelUrl: PayPalCoreConstants.callbackURLScheme + "://card/cancel"
         )
-        
-        confirmPaymentSource.paymentSource = PaymentSource(card: card)
+
+        confirmPaymentSource.paymentSource = PaymentSource(card: serviceCard)
```
scannillo: This `PaymentSource` is currently [public](https://github.com/paypal/iOS-SDK/blob/b0e61d6f27d6d094354f1322f8df2310f15c4ad9/Sources/CardPayments/Models/PaymentSource.swift#L6). I don't think it needs to be.

scannillo: Actually, it looks like `ConfirmPaymentSource` uses [this `PaymentSource`](https://github.com/paypal/iOS-SDK/blob/b0e61d6f27d6d094354f1322f8df2310f15c4ad9/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift#L77-L80). Do we even need the one linked above?

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// This class is used to configure a CardRequest for vaulting
```
jaxdesmarais:
```suggestion
/// Used to configure Vault request
```

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,15 @@
...
+public struct Vault {
+    
+    /// Property used to associate  a payment method with vault of customer with this ID
```
jaxdesmarais: Nit to match the [Android docs](https://github.com/paypal/Android-SDK/blob/772220ada6aed8456e5d146131d0bfe311079688/CardPayments/src/main/java/com/paypal/android/cardpayments/Vault.kt#L9-L10) more closely: 
```suggestion
    /// Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. If not specified a customer ID will be generated.
```

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,15 @@
...
+    public let customerID: String?
+
+    /// Creates an instance of a vault object
+    /// - Parameters:
+    ///   - customerID: The identifier for customer. If not specified, PayPal will create one
```
jaxdesmarais:
```suggestion
    ///   - customerID: Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. If not specified a customer ID will be generated.
```

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: Yes, it's used to format the request body.
```
{
  "application_context": {
    "return_url": "sdk.ios.paypal://card/success",
    "cancel_url": "sdk.ios.paypal://card/cancel"
  },
  "payment_source": {
    "card": {
      "number": "4344511979394247",
      "security_code": "123",
      "billing_address": null,
      "name": null,
      "attributes": {
        "vault": {
          "store_in_vault": "ON_SUCCESS"
        },
        "verification": {
          "method": "SCA_ALWAYS"
        },
        "customer": {
          "id": "test567"
        }
      },
      "expiry": "2025-01"
    }
  }
}

```

CHANGELOG.md
KunJeongPark: I had a question on this. It doesn't actually break anyone's implementation since it is an optional property set to nil by default in initializer.
Do I still put it as a breaking change?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: I see what you mean. We are not using it anywhere now. 

I think it was used to return results after capture/authorize. If we have plans to get the entire authenticationResult back from getOrderInfo lite(?) we would need something like this. I think CardResult used to have this included.

It is in the ConfirmPaymentSourceResponse but we are not getting back PaymentSource from this endpoint. I don't know whether that was a result of using clientID only. I'm looking it up now.

Sources/CardPayments/Models/ServiceCard.swift
scannillo: Another option is to leverage Swift's namespacing to create an internal `Card` type, and it will not conflict with the public facing one. Here's a branch with a simplified diff example off of `main` --> https://github.com/paypal/iOS-SDK/compare/demo-internal-card-type. The internal type in the example would be accessed like `ConfirmPaymentSource.PaymentSource.Card` due to the way it's nested. We can flatted things as well with custom encoders if you want, but IMO this route if using Swift namespacing is easier to understand than another `Service` layer.

Thoughts? cc: @jaxdesmarais

KunJeongPark: I like that. I was considering putting the ServiceCard within the `ConfirmPaymentRequest` struct. It makes more sense the way you have it. I thought having another Card struct might be confusing. I like the way you can specify within the PaymentSource struct itself `PaymentSource.Card`, I didn't know that.

jaxdesmarais: I dig this approach as well. I think it achieves @KunJeongPark goal of moving the API layer out of the public facing interface while also keeping our interface cleaner and more maintainable vs creating a `Service` layer. 🚀

KunJeongPark: I put the new private Card struct at the ConfirmPaymentSourceRequest level since I got "at most 1 level deep" warning when I embedded it in the PaymentSource.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -74,8 +75,80 @@ struct ConfirmPaymentSourceRequest: APIRequest {
...
+                    self.customer = Customer(id: id)
+                }
+            }
+        }
+    }
+    
     private struct PaymentSource: Encodable {
         
-        let card: Card
+        var card: EncodedCard
```
jaxdesmarais: Take it or leave it, calling the property the same name may make this more clear - it can also be a constant
```suggestion
        let encodedCard: EncodedCard
```

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Used to configure Vault request
+public struct Vault {
+    
+    /// Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. If not specified a customer ID will be generated.
```
jaxdesmarais: Take it or leave it to break this out onto two lines since it's a bit long:
```suggestion
    /// Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. 
    /// If not specified a customer ID will be generated.
```

Sources/CardPayments/Models/Vault.swift
```diff
@@ -0,0 +1,15 @@
...
+    public let customerID: String?
+
+    /// Creates an instance of a vault object
+    /// - Parameters:
+    /// - customerID: Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. If not specified a customer ID will be generated.
```
jaxdesmarais: Same here, but again not a blocker:
```suggestion
    /// - customerID: Optional. When set, the value of this property will be used to associate a payment method with the vault of a customer with this ID. 
    /// If not specified a customer ID will be generated.
```

jaxdesmarais (approved):
👏  thanks for being so willing to address all the feedback! This is going to make the SDK better for our merchants as well as more maintainable for us.

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
KunJeongPark: Yes, I was just looking at that. You are right.

KunJeongPark: I appreciate all of your comments and feedback!

PR #163, opened 2023-06-13
Remove Reference to AccessToken in Integration Guides

KunJeongPark:
### Summary of changes

- Change instruction for instantiating CoreConfig to use clientID instead of accessToken in README and /docs

### Checklist

- @KunJeongPark

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,10 +31,7 @@ This SDK supports:
 
 ## Client ID
 
-The PayPal SDK uses client ID for authentication.
-
-> The following example can be adapted to any server-side language/framework of your choice. We use command-line curl requests to demonstrate the overall composition of the Access Token HTTP request.
-
+The PayPal SDK uses a client ID for authentication.
 
 ## Modules
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ To accept a certain payment method in your app, you only need to include that pa
 ## Sample Code
 
 ```swift
-// STEP 0: Fetch an CLIENT_ID and ORDER_ID from your server.
+// STEP 0: Enter a CLIENT_ID and fetch ORDER_ID from your server.
 
 // STEP 1: Create a CoreConfiguration object
 paymentConfig = CoreConfig(clientID: CLIENT_ID, environment: .sandbox)
--- a/README.md
+++ b/README.md
@@ -29,48 +29,12 @@ This SDK supports:
 * UIKit
 * SwiftUI
 
-## Access Token
+## ClientID
 
-The PayPal SDK uses access tokens for authentication.
+The PayPal SDK uses clientID for authentication.
 
 > The following example can be adapted to any server-side language/framework of your choice. We use command-line curl requests to demonstrate the overall composition of the Access Token HTTP request.
 
-To create an access token:
-
-1. Follow the steps in [Get Started](https://developer.paypal.com/api/rest/#link-getstarted) to obtain a client ID and secret from the PayPal Developer site.
-1. Make an HTTP request with Basic Authentication using client ID and secret to fetch an access token:
-
-**Request**
-```bash
-# for LIVE environment
-curl -X POST https://api.paypal.com/v1/oauth2/token \
--u $CLIENT_ID:$CLIENT_SECRET \
--H 'Content-Type: application/x-www-form-urlencoded' \
--d 'grant_type=client_credentials&response_type=token&return_authn_schemes=true'
-
-# for SANDBOX environment
-curl -X POST https://api.sandbox.paypal.com/v1/oauth2/token \
--u $CLIENT_ID:$CLIENT_SECRET \
--H 'Content-Type: application/x-www-form-urlencoded' \
--d 'grant_type=client_credentials&response_type=token&return_authn_schemes=true'
-```
-
-:warning:&nbsp;Make sure the environment variables for `CLIENT_ID` and `CLIENT_SECRET` are set.
-
-**Response**
-
-```json
-{
-  "scope": "...",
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -29,48 +29,12 @@ This SDK supports:
...
-  "scope": "...",
-  "access_token": "<ACCESS_TOKEN>",
-  "token_type": "Bearer",
-  "app_id": "...",
-  "expires_in": 32400,
-  "nonce": "..."
-}
-```
-
-Use the value for `access_token` in the response to create an instance of `CoreConfig` to use with any of the SDK's feature clients.
```
KunJeongPark, line 35: Do we still need this so that merchants can to make calls to create, authorize/capture orders on their servers?

jaxdesmarais: I don't think we need to include it in our docs. We can rely on this being documented in the PayPal docs for the server side authorize/capture call. It was only documented here since it was necessary to use the SDK previously.

README.md
```diff
@@ -86,10 +50,10 @@ To accept a certain payment method in your app, you only need to include that pa
 ## Sample Code
 
 ```swift
-// STEP 0: Fetch an ACCESS_TOKEN and ORDER_ID from your server.
+// STEP 0: Fetch an CLIENT_ID and ORDER_ID from your server.
```
scannillo:
```suggestion
// STEP 0: Fetch a CLIENT_ID and ORDER_ID from your server.
```

Do they have to fetch a client ID? They can also have one safely hard-coded on their client as well

KunJeongPark: In demo app, we have a merchant server call to fetch client id but they can just enter it manually as well. 
Should I say "fetch or enter"?

README.md
```diff
@@ -29,48 +29,12 @@ This SDK supports:
 * UIKit
 * SwiftUI
 
-## Access Token
+## Client ID
 
-The PayPal SDK uses access tokens for authentication.
+The PayPal SDK uses client ID for authentication.
 
 > The following example can be adapted to any server-side language/framework of your choice. We use command-line curl requests to demonstrate the overall composition of the Access Token HTTP request.
```
scannillo:
```suggestion
```

README.md
jaxdesmarais: I don't expect most merchants will make an API request to fetch the client ID. I think they will probably hardcode it as Sammy said.

KunJeongPark: Addressed in 47abc2204e62fc9738a73e258d1931f9181907ea

README.md
```diff
@@ -29,48 +29,12 @@ This SDK supports:
 * UIKit
 * SwiftUI
 
-## Access Token
+## Client ID
 
-The PayPal SDK uses access tokens for authentication.
+The PayPal SDK uses client ID for authentication.
```
jaxdesmarais:
```suggestion
The PayPal SDK uses a client ID for authentication.
```

KunJeongPark: Addressed in 44250f07b9791b6cac9cca79803900a17b1d10ee

README.md
KunJeongPark: Addressed in 44250f07b9791b6cac9cca79803900a17b1d10ee

README.md
```diff
@@ -86,10 +47,10 @@ To accept a certain payment method in your app, you only need to include that pa
 ## Sample Code
 
 ```swift
-// STEP 0: Fetch an ACCESS_TOKEN and ORDER_ID from your server.
+// STEP 0: Enter a CLIENT_ID and fetch ORDER_ID from your server.
```
scannillo:
```suggestion
// STEP 0: Fetch a CLIENT_ID and ORDER_ID from your server.
```

scannillo: Do we have a sample cURL or link to a cURL for how merchants can perform a clientID fetch on their server?

KunJeongPark: That's a good point, I don't think we have one for the function Steven wrote on for the merchant server. I can write it in as a comment until we have it in the integration guide that we can link. What do you think?

sshropshire: A typical merchant won't need to fetch a client ID. It will most likely be injected at compile time using an environment (or hardcoded) variable for most, unless they need to dynamically set their client ID.

sshropshire: We'll eventually support multiple merchants in the Demo app e.g. to test connected partner flows, so we need to set it dynamically for PPCP e2e testing.

sshropshire (approved):
🌊

README.md
jaxdesmarais: If we expect merchants to hardcode this value, we should document the merchants expected path to enter a client ID vs fetching it on their server, imo

sshropshire: [On Android](https://github.com/paypal/Android-SDK/tree/main/docs/CardPayments#2-initiate-the-payments-sdk), we only mention that it's needed to construct a `CoreConfig` object. It is technically up to the merchant, and `clientID` doesn't need to be protected like the `clientSecret` does. Maybe we can do something similar here and eliminate step 0?

scannillo: Do we need this snippet in the README at all? We have this integration covered in our `docs/` section per module.

sshropshire: We should be able to remove:
```swift
let paymentConfig = CoreConfig(clientID: "<CLIENT_ID>", environment: .sandbox)
```
is probably all that's needed.

scannillo: I mean this entire section under `## Sample Code` in the README ([link](https://github.com/paypal/iOS-SDK#sample-code)). IMO it can be removed. We don't have an equivalent in Android and the sample code for each flow already lives in `docs/`.

sshropshire: 👍 agreed.

PR #167, opened 2023-07-06
Remove UIKit Option in Demo App

KunJeongPark:
### Summary of changes

- Removed UIKit option in demo app in settings 
- Remove UIKit ViewControllers used for UIKit option
- Removed enum DemoUIFramework 

### Checklist

### Authors
- @KunJeongPark


jaxdesmarais (commented):
This can be a follow up PR or in this PR but we should convert `FeatureBaseViewController`, `ViewController`, and remove `FeatureBaseViewControllerRepresentable`. Those 3 were all used for interoperability and sharing components across SwiftUI and UIKit. Since we want to take a SwiftUI first approach we should move these all into SwiftUI vs mixing the two.

scannillo:
Just pulled this down, it looks awesome 👏.

I see that we still are specifying `UIKit` in this [UI Test](https://github.com/paypal/iOS-SDK/blob/b3dbc6523f323003b4ccd10400d258dc06e84375/Demo/DemoUITests/CardUITests.swift#L11). We should update this to SwiftUI - but it looks like this UI test isn't even running on CI, it's probably pretty outdated.

scannillo (approved):
Looks great! Left a small comment about a UI test

jaxdesmarais (approved):

KunJeongPark:
> Just pulled this down, it looks awesome 👏.
> 
> I see that we still are specifying `UIKit` in this [UI Test](https://github.com/paypal/iOS-SDK/blob/b3dbc6523f323003b4ccd10400d258dc06e84375/Demo/DemoUITests/CardUITests.swift#L11). We should update this to SwiftUI - but it looks like this UI test isn't even running on CI, it's probably pretty outdated.

Thank you Sammy! Good call. I saw that the test is failing on main as well with SwiftUI option and for UIKit option. I will make a separate ticket for this UI test.

PR #169, opened 2023-07-11
Update MXO to Version 1.0.0

KunJeongPark:
### Summary of changes

- Bump `PayPalCheckout` to 1.0.0 in Podspec and Package.swift files

### Checklist

- [x] Added a changelog entry

### Authors
- @KunJeongPark

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,11 +34,8 @@ Pods
 Carthage/
 
 # SPM
-Package.resolved
 .swiftpm
 .build
-*/xcshareddata/swiftpm
-SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
 
 # Fastlane
 /fastlane/report.xml
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ Carthage/
 # SPM
 .swiftpm
 .build
+SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
 
 # Fastlane
 /fastlane/report.xml
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+* PayPalNativePayments
+  * Bump `PayPalCheckout` to `1.0.0`
 ## 0.0.9 (2023-06-23)
 * Breaking Changes
   * CardPayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ## unreleased
 * PayPalNativePayments
   * Bump `PayPalCheckout` to `1.0.0`
+  
 ## 0.0.9 (2023-06-23)
 * Breaking Changes
   * CardPayments
--- a/Demo/Demo.xcodeproj/project.pbxproj
+++ b/Demo/Demo.xcodeproj/project.pbxproj
@@ -775,7 +775,7 @@
 			repositoryURL = "https://github.com/paypal/paypalcheckout-ios";
 			requirement = {
 				kind = exactVersion;
-				version = 0.112.1;
+				version = 1.0.0;
 			};
 		};
 		BE1766D526FA7AC3007EF438 /* XCRemoteSwiftPackageReference "InAppSettingsKit" */ = {
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,0 +1,14 @@
+{
+  "pins" : [
+    {
+      "identity" : "paypalcheckout-ios",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/paypal/paypalcheckout-ios",
+      "state" : {
+        "revision" : "773568fbbffd54f900d6d78be7793ae1871d3b35",
+        "version" : "1.0.0"
+      }
+    }
+  ],
+  "version" : 2
+}
--- a/Package.swift
+++ b/Package.swift
@@ -35,7 +35,7 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .exact("0.112.1"))
+        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .exact("1.0.0"))
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
--- a/PayPal.podspec
+++ b/PayPal.podspec
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
   s.subspec "PayPalNativePayments" do |s|
    s.source_files  = "Sources/PayPalNativePayments/**/*.swift"
    s.dependency "PayPal/CorePayments"
-   s.dependency "PayPalCheckout", "0.112.1"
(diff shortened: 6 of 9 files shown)


sshropshire (approved):

Package.swift
```diff
@@ -35,7 +35,7 @@ let package = Package(
     ],
     dependencies: [
         // Dependencies declare other packages that this package depends on.
-        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .exact("0.112.1"))
+        .package(name: "PayPalCheckout", url: "https://github.com/paypal/paypalcheckout-ios", .exact("1.0.0"))
```
jaxdesmarais, line 38: We should consider moving PPCP to use a `binaryDependency` instead of an explicit dependency. Right now when CI runs for the `SPMTest` app it runs an old version of NXO since SPMTest points to the main branch of our SDK. Moving to a binary dependency would allow us to pull in the NXO dependency as part of our package graph on version bump PRs vs needing to bump the sample app in a follow up PR.

It would also allow us to obfuscate this module from merchants since right now if a merchant uses SPM they will _always_ see `PayPalCheckout` pulled in as a dependence even if they aren't using the `PayPalNativePayments` module.

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -775,7 +775,7 @@
 			repositoryURL = "https://github.com/paypal/paypalcheckout-ios";
 			requirement = {
 				kind = exactVersion;
-				version = 0.112.1;
+				version = 1.0.0;
```
jaxdesmarais, line 778: @KunJeongPark can you create a follow up ticket to remove `PayPalCheckout` as a package dependency in the demo app? It's used because original we thought we wanted merchants to pull in `PayPalCheckout` directly. Since we have moved away from this we should mock the NXO classes in our demo app vs using their classes directly.

KunJeongPark: Jax, I am not 100% sure what you mean by "mock the NXO classes" in the demo app

jaxdesmarais: Instead of using their classes directly we need to create our own classes to use - our classes will likely be a replication of the NXO classes we are using currently.

KunJeongPark: I am not sure why we need to do this Jax. Don't we want to see how actual PayPalCheckout behaves in the demo app?

Package.swift
jaxdesmarais: No - SPMTest is the equivalent to how a merchant would pull in our SDK. Merchants should not pull in `PayPalCheckout` directly, in fact they shouldn't need to know that we need the underlying module at all. SPMTest does not pull it in today it's pulled in via the `dependencies` line in our `Package.swift`, that app does not currently pull `PayPalCheckout` directly.

Demo/Demo.xcodeproj/project.pbxproj
jaxdesmarais: Merchants shouldn't know about `PayPalCheckout` at all and right now our demo app displays that they should both use our `PayPalNativePayments` _and_  pull in `PayPalCheckout`. This is not a path that we want merchants to take so we should remove it from our demo app. Our demo app should display how we expect merchant to use `PayPalNativePayments`.

scannillo: Jax is right - a merchant should never integrate with PayPalCheckout directly. Only the PPCP SDK should.

We have parts of the demo app that fetch orders via PayPalCheckout, when it should instead use our sample server.

Package.swift
scannillo: Check out this PR for some history/reference - https://github.com/braintree/braintree_ios/pull/882

.gitignore
```diff
@@ -34,11 +34,8 @@ Pods
 Carthage/
 
 # SPM
-Package.resolved
 .swiftpm
 .build
-*/xcshareddata/swiftpm
-SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```
scannillo, line 41: I think we can keep this one in the gitignore

Demo/Demo.xcodeproj/project.pbxproj
KunJeongPark: Sammy, we do create order in our sample server for NativeCheckout in our demo app. 

I created ticket to remove PayPalCheckout as Package Dependency in demo app. DTNOR-1008

Package.swift
KunJeongPark: I created ticket for creating binary dependency instead of explicit dependency for PayPalCheckout.  DTNOR-296

.gitignore
KunJeongPark: @jaxdesmarais We removed this because it was not getting updated using "swift package resolve" command right?

jaxdesmarais: I think we want `*/xcshareddata/swiftpm` but @scannillo is right it looks like we have `SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` included in the BT `.gitignore`. We should revert the removal of `SampleApps/SPMTest/SPMTest.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved`.

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+* PayPalNativePayments
+  * Bump `PayPalCheckout` to `1.0.0`
 ## 0.0.9 (2023-06-23)
```
scannillo, line 8:
```suggestion
## 0.0.9 (2023-06-23)

```

.gitignore
KunJeongPark: I reverted deletion of SampleApps/../Package.resolved in gitIgnore file in the commit  => 99c10102b0d453532d041b140ff717f9a09e9720

CHANGELOG.md
KunJeongPark: Oh you mean line between version notes?

KunJeongPark: Added a line after *Bump .. in commit a1522a11c1bc0864228b890e618f53be524596f4

jaxdesmarais (approved):

.gitignore
scannillo: Awesome!

Package.resolved
```diff
@@ -0,0 +1,14 @@
+{
```
scannillo, line 1: I also think we can delete this file. We don't have a top level `Package.resolved` in [braintree_ios](https://github.com/braintree/braintree_ios). The other `package.resolved` files that you committed in this PR will cover the Demo app & PayPal xcode proj.

jaxdesmarais: I think this is actually generated when we have the `PayPalCheckout` as a dependency in our `Package.swift`. When that was done on BT initially we also had a top level `Package.resolved` ([PR here](https://github.com/braintree/braintree_ios/pull/827/files#diff-3b22d20b6ae7cdf1e2591c647bcae94a3db9ab955f497aeb9d0715f468e6e6ce)). When we move PPCP to also use a binary dependency we should be able to remove this file. 

@KunJeongPark can you please make sure removing this file is in the ticket to move to a binary dependency? Just want to make sure removing this doesn't get missed when we make that change.

KunJeongPark: Done.

PR #171, opened 2023-07-28
Send 'production' over 'live' to FPTI

scannillo:
### Reason for changes

- While creating analytics dashboards, we realized that the BT & PPCP SDKs differ in the values we are sending for the `merchant_sdk_env` key to FPTI
    - BT Mobile SDKs are sending: production & sandbox, PPCP mobile SDKs are sending live & sandbox

### Summary of changes

- Send `production` over `live` in FPTI post body

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -45,7 +45,7 @@ public struct AnalyticsService {
             let clientID = coreConfig.clientID
             
             let eventData = AnalyticsEventData(
-                environment: http.coreConfig.environment.toString,
+                environment: http.coreConfig.environment.analyticsString,
                 eventName: name,
                 clientID: clientID,
                 orderID: orderID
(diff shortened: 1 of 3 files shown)


scannillo:
We technically could change the BT SDKs to send `live`, but Jax mentioned that PayPal uses the term "production" as well. I personally don't feel strongly about which SDK gets this change. Happy to do it in BT instead if folks prefer keeping this as-is.

jaxdesmarais (approved):
Since this SDK is not GA yet, I think it makes sense to make the change here vs the BT SDK that is GA.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -45,7 +45,7 @@ public struct AnalyticsService {
             let clientID = coreConfig.clientID
             
             let eventData = AnalyticsEventData(
-                environment: http.coreConfig.environment.toString,
```
jaxdesmarais, line 48: Can we remove `toString` from the `Environment` enum all together? Looks like this is the only place it was used.

scannillo:
Closing in favor of - https://github.com/braintree/braintree_ios/pull/1083

PR #173, opened 2023-08-01
Revert Vault with Purchase on Client

KunJeongPark:
### Summary of changes

- Reverting implementation of vault with purchase on the client. 
   As of now, plan is for merchant to implement this on their server side 
  because of issues with a merchant integration pattern that requires authentication to be
  passed from server side

### Checklist

- [ x] Added a changelog entry
  
### Authors

- @KunJeongPark


Holsta0822 (commented):
👍

sshropshire (approved):

jaxdesmarais (commented):
Do we also want to remove the [`Vault.swift`](https://github.com/paypal/paypal-ios/blob/52a6218c0fccc655d5807cad1663a0b2292df86a/Sources/CardPayments/Models/Vault.swift) file as well as the vault objects from [`ConfirmPaymentSourceRequest`](https://github.com/paypal/paypal-ios/blob/52a6218c0fccc655d5807cad1663a0b2292df86a/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift)

KunJeongPark:
Hi @jaxdesmarais I think vault objects are removed from `ConfirmPaymentSourceRequest`. Good catch on the `Vault.swfit`
Addressed in commit ddb7e64c0c2a32ba963b894cc18baeb807e87e7c

jaxdesmarais (approved):

PR #175, opened 2023-08-03
UI for Vault with Purchase

KunJeongPark:
### Summary of changes

- Disabled toggle button UI for Vault with Purchase for now.
- It was created for now reverted client side Vault with Purchase
- Exposed optional Customer ID for testing Vault without Purchase which is being implemented. 
- Plan is to turn this toggle button on when that implementation of passing in vault options for create order 
   is done for server side, sample merchant server.
   We have work to do to provide correct parameters, headers for different integrations for create order, setup token.
- We are waiting on current issues on that implementation to be resolved. 
- Final UI will be more complex with selections for different integration patterns including C4, vault without purchase flow vs checkout flow with/without vault option.
<p float="left">
<image src = "https://github.com/paypal/paypal-ios/assets/9273272/d24b991b-db1c-4052-9791-6a1fe64154dd" width = "500" />
</p>

### Checklist

### Authors
- @KunJeongPark

### DIFF
--- a/Demo/Demo/SwiftUIComponents/SwiftUICardDemo.swift
+++ b/Demo/Demo/SwiftUIComponents/SwiftUICardDemo.swift
@@ -35,12 +35,12 @@ struct SwiftUICardDemo: View {
                         cvvText = cardFormatter.formatFieldWith(newValue, field: .cvv)
                     }
                 HStack {
-                    Toggle("Should Vault", isOn: $shouldVaultSelected)
+                    Toggle("Should Vault with Purchase", isOn: $shouldVaultSelected)
+                    // TODO: turn on if vault with purchase on sample server is implemented
+                        .disabled(true)
                     Spacer()
                 }
-                if shouldVaultSelected {
-                    FloatingLabelTextField(placeholder: "Vault Customer ID (Optional)", text: $vaultCustomerID)
-                }
+                FloatingLabelTextField(placeholder: "Vault Customer ID (Optional)", text: $vaultCustomerID)
                 Button("\(DemoSettings.intent.rawValue.capitalized) Order") {
                     guard let card = baseViewModel.createCard(
                         cardNumber: cardNumberText,


Demo/Demo/SwiftUIComponents/SwiftUICardDemo.swift
```diff
@@ -35,12 +35,12 @@ struct SwiftUICardDemo: View {
                         cvvText = cardFormatter.formatFieldWith(newValue, field: .cvv)
                     }
                 HStack {
-                    Toggle("Should Vault", isOn: $shouldVaultSelected)
+                    Toggle("Should Vault with Purchase", isOn: $shouldVaultSelected)
+                    // TODO: turn on if vault with purchase on sample server is implemented
+                        .disabled(true)
```
jaxdesmarais, line 40: Instead of `.disabled(true)` I think it makes more sense to use `.hidden()` to remove this from the UI completely until it can be turned back on. What do you think?

KunJeongPark: I left that there because I thought it would be strange to have optional customerID field in there by itself. 
Right now, I am using that field for testing Vault without Purchase flow with the button for Vault without Purchase right below. This is temporary while I do some refactor on the UI for demo app. I think there was actually a ticket to have UI for vault with purchase and mention in stand about having this UI but disabled.

I am working on making opening screen for having three sets of options, vault without purchase (with options for payment methods) and checkout with payment options and integration type. I think having App struct replacing SceneDelegate/AppDelegate is straight forward, I tried commenting out those files and having App file as entry point and it seems to be fine.

There are also selections from Settings screen which may need to be moved since first set of selections is 1. vault without purchase vs checkout and 2. integration type. 
Difficulty is having many options but making it the flow easy, removing UI elements not needed for flows selected.

I want to extract out the Card view from current demo screen so I can reuse it for the Vault without Payment flow. 
I am working on moving Card.swift and Address.swift to Core module so they can be reused.

jaxdesmarais: > I left that there because I thought it would be strange to have optional customerID field in there by itself.

I think either option is fine since this is internal. To me though a disabled element typically indicates there is an action that can be done to set the state from disabled to enabled whereas hidden indicates that it cannot should be used in it's current state.

I think before we get too far into building UI it's worth tackling removing UIKit completely from the Demo app. This will make it much easier to add and update demo app features. A few of the views are still tightly coupled to UIKit views which will make it much harder to adapt the new views you want.

> I am working on moving Card.swift and Address.swift to Core module so they can be reused.

I don't think we want to move these into core unless they can be used across multiple modules. Especially card since merchants using PayPal or NXO should not ever need to interact with the card model. I may be missing some context that you and @sshropshire have already discussed though, but imo doesn't make sense to move those into core.

KunJeongPark: I agree about removing UIKit. That's why I was investigating removing SceneDelegate/AppDelegate.

For Card.swift, for right now, it is a form that we use to collect card info to be passed into SDK to build encodable structs for PaymentSource which looks different for different requests. I wanted a way to use this for Demo app for Vault without Purchase flow and GraphQL request ( which lives in Core Module ) for update setup token. 

The problem I am running into right now is having to pass in some kind of Card struct from the CardClient to instantiate GraphQLQuery which lives in CorePayment for updateSetupToken call.

 I will discuss it further with @sshropshire

KunJeongPark: That makes sense what you are saying about PayPal and NXO though. Let me think on it and discuss with Steven.

jaxdesmarais: That sounds good! I am sure y'all will come up with something that makes sense for both platforms while keeping things DRY where possible. 😄

scannillo (approved):

sshropshire (approved):

PR #172, opened 2023-08-01
Vault without Purchase

KunJeongPark:
### Summary of changes

- Create setup token with empty card object in demo app, simulating merchant server's behavior and output
- Use graphQL endpoint to update setup token with payment method.
- Create VaultDelegate protocol and delegate methods to return setup token id and vault status to merchant app
- Implement creating actual payment token using setupToken in demo app via sample merchant server call to PayPal API
- Fix unit tests, add new ones from changes

There is currently a refactor ticket (https://github.com/paypal/paypal-ios/pull/177) on the network layer that includes replacing GraphQLClient. There are some TODO comments to mark some changes to address after the refactor.

**TODO for another PR handling 3DS contingency:** 
- Use GET request for getting setupToken to get minimum PaymentSource information and for 3DS contingency, updated 
   status, 3Ds results.

  We are not fetching minimum paymentSource from UpdateSetupToken graphQL endpoint. (following JS implementation)
  I am assuming that this is because of cards with 3DS contingency where GET request is needed
  for updated status. Using a separate GET request lets me use objects internal to CardPayments module. 

- Not clear on 3DS contingency => I am not getting 3DS url (helios link) from setup token endpoints (from setup token create with or without card info, update setup token endpoint) with many of the test cards that I've   
tried. These test cards did trigger 3DS contingency in checkout flow.

I am confused about whether 3DS contingency gets kicked off at moment of standalone vault or during checkout process with the vaulted method. I think it's the former but wanted to make sure I know all the steps involved in creating 3DS contingency correctly for standalone vaulting.

### Checklist

- [x] Added a changelog entry

### Authors
- @KunJeongPark

PR #176, opened 2023-08-08
Remove EligibiltyAPI 

scannillo:
### Reason for changes

This removal is to solely to make the future GraphQL refactor PRs easier.

Currently the GraphQL implementation is entirely separate from the REST (and other HTTP) implementation. It was added and currently only used with the Eligibility API. The Eligibility API will be replaced with a different API in the future and is not exposed to merchants. So to me, it feels OK to remove to allow for easier refactoring.

### Notes

In future PRs, we will make a few updates to the networking layer:
- Simplify the `APIRequest` protocol 
    - Remove it's associated decodable type for easier testing
    - Move Data encoding out of `APIRequest` initializers ([example](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift#L39-L44))
    - Move `URLRequest` construction out of `APIRequest` protocol extension ([example](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CorePayments/Networking/Protocols/APIRequest.swift#L26-L47))
- Create `RestRequest` and `GraphQLRequest` protocols
- Use single HTTP layer for REST & GraphQL requests

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo 
-


PayPal.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
```
scannillo:
```suggestion
	objectVersion = 52;
```

scannillo:
```suggestion
	objectVersion = 52;
```

scannillo:
Actually, I realized [@KunJeongPark's draft PR](https://github.com/paypal/paypal-ios/pull/172/files) is currently using this.

I will close this for now and we can remove this after the refactor.

PR #177, opened 2023-08-09
[PART 1] Networking Refactor - Remove `APIRequest` usage and add `PPassAPI.swift` layers

scannillo:
### Reason for changes

This PR ...
- Sets the stage for having a GraphQL layer that uses the same underlying `URLSession` as REST requests
- Follows [Android](https://github.com/paypal/paypal-android/pull/138) by using `<PPaaS>API.swift` layers to abstract networking out of each feature client
- Addresses several longstanding limitations of the `APIRequest` protocol

### Summary of changes

- Add `TrackingEventsAPI.swift` & `CheckoutOrdersAPI.swift`
    - Update `CardClient` to interface with `CheckoutOrdersAPI` instead of `APIClient`
    - Update `AnalyticsService` to interface with `TrackingEventsAPI` instead of `APIClient`
    - These API.swift files handle creating the encodable post body & constructing a `RESTRequest`
- Remove usage of `APIRequest` protocol in favor of inline `RESTRequest` object creation
    - This offers a lot of wins:
        - Moved `URLRequest` creation out of [protocol extension](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CorePayments/Networking/Protocols/APIRequest.swift#L15-L48) (impossible to test in isolation) into the `HTTP` layer
        - Remove [`associatedtype`](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CorePayments/Networking/Protocols/APIRequest.swift#L4) to represent Decodable. Makes testing our flows where we don't decode the response (like analytics) much easier. Will also make mocking any API request & response simpler.
        - Move post body encoding out of [request `init`](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift#L39-L44) and into `<PPaSS>API` layers ([example](https://github.com/paypal/paypal-ios/blob/ffd9b10baee0acaf9a4c8b5d6da88509b1ea98c9/Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift#L27-L30)) (left a TODO for moving this into own class). This was also a pain to test.
        - This inline request creation is more readable IMO
- Simplify `ConfirmPaymentSourceRequest` Encodable
    - Implement [custom encoder](https://github.com/paypal/paypal-ios/blob/ffd9b10baee0acaf9a4c8b5d6da88509b1ea98c9/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift#L74-L106) versus nesting structs
- `APIClient` changes
    - Replace `fetch<T: APIRequest>(request: T)` generic func with `fetch(request: RESTRequest)`
    - Move `URL` construction out of `APIRequest` [protocol extension](https://github.com/paypal/paypal-ios/blob/257df315bb8ab2e323df39bf17a5af48e36498a1/Sources/CorePayments/Networking/Protocols/APIRequest.swift#L27-L36) and into this function
- `HTTP` changes
    - Replace `performRequest(_ request: any APIRequest)` generic func with `performRequest(_ request: HTTPRequest)`

### Next Steps

- Add GraphQL equivalent of [`APIClient.fetch()` func](https://github.com/paypal/paypal-ios/blob/e5285f7687797b4e22d718d7aae6a411e1a345c6/Sources/CorePayments/Networking/APIClient.swift#L31)
    - This function will perform similar actions to the REST equivalent function, but will take in a `GraphQLRequest` type and transform it into a `HTTPRequest`
- Unit tests in a separate PR (wanted to make this PR easy to digest)
- Delete `APIRequest` file
- Address TODOs
- Reorganize file structure

### Note

Tests are expected to fail. This work will have several subsequent PRs.

### Checklist

- ~Added a changelog entry~

### Authors
- @scannillo


Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -4,132 +4,89 @@ import CorePayments
...
+    // MARK: - Initializer
+    
+    init(cardRequest: CardRequest) {
+        self.cardRequest = cardRequest
     }
     
-    private struct PaymentSource: Encodable {
+    // MARK: - Custom Encoder
+    
+    func encode(to encoder: Encoder) throws {
```
jaxdesmarais, line 65: 👏 imo this is soooo much easier to read and reason about vs a bunch of nested structs

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -4,132 +4,89 @@ import CorePayments
...
+        try card.encode("\(cardRequest.card.expirationYear)-\(cardRequest.card.expirationMonth)", forKey: .expiry)
+        try card.encode(cardRequest.card.cardholderName, forKey: .name)
+
+        if let cardBillingInfo = cardRequest.card.billingAddress {
+            var billingAddress = card.nestedContainer(keyedBy: BillingAddressKeys.self, forKey: .billingAddress)
+            try billingAddress.encode(cardBillingInfo.addressLine1, forKey: .addressLine1)
+            try billingAddress.encode(cardBillingInfo.addressLine2, forKey: .addressLine2)
+            try billingAddress.encode(cardBillingInfo.postalCode, forKey: .postalCode)
+            try billingAddress.encode(cardBillingInfo.region, forKey: .adminArea1)
+            try billingAddress.encode(cardBillingInfo.locality, forKey: .adminArea2)
```
jaxdesmarais: Since these are optional do we want to use `encodeIfPresent` instead?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -4,132 +4,89 @@ import CorePayments
...
-                billingAddress: card.billingAddress,
-                name: card.cardholderName,
-                expiry: "\(card.expirationYear)-\(card.expirationMonth)",
-                attributes: Attributes(verificationMethod: scaType.rawValue)
-            )
+        var paymentSource = topLevel.nestedContainer(keyedBy: PaymentSourceKeys.self, forKey: .paymentSource)
+        var card = paymentSource.nestedContainer(keyedBy: CardKeys.self, forKey: .card)
+        try card.encode(cardRequest.card.number, forKey: .number)
+        try card.encode("\(cardRequest.card.expirationYear)-\(cardRequest.card.expirationMonth)", forKey: .expiry)
+        try card.encode(cardRequest.card.cardholderName, forKey: .name)
```
jaxdesmarais: Same question here on using `encodeIfPresent` instead potentially?

Sources/CorePayments/Networking/Protocols/APIRequest.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 
+// TODO: - Remove this protocol once tests are updated 🎉
```
jaxdesmarais, line 3: 🥳

PayPal.xcodeproj/project.pbxproj
```diff
@@ -549,6 +554,7 @@
 			children = (
 				06CE009F26F3DF100000CC46 /* ConfirmPaymentSourceRequest.swift */,
 				06CE00A226F3E32A0000CC46 /* ConfirmPaymentSourceResponse.swift */,
+				80E2FDBD2A83528B0045593D /* CheckoutOrdersAPI.swift */,
```
jaxdesmarais, line 535: nit to sort by name, looks like this new file is out of order

PayPal.xcodeproj/project.pbxproj
```diff
@@ -691,8 +697,11 @@
 				E646231928369B71008AC8E1 /* GraphQL */,
 				BEA100E626EF9EDA0036A6A5 /* APIClient.swift */,
 				804E628529380B04004B9FEF /* AnalyticsService.swift */,
+				80E2FDBF2A8353550045593D /* TrackingEventsAPI.swift */,
```
jaxdesmarais, line 678: Same here, looks like these are maybe in a weird order, but I think there are also nested directories in here so not sure how that fits into sorting by name

scannillo: Yeh - agreed. IMO the current structure of `CorePayments` is messy / unorganized. I was planning to reorganize everything at the end of this refactor to make the diffs easier to read. Thoughts?

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
```diff
@@ -0,0 +1,49 @@
...
+                .contentType: "application/json", .acceptLanguage: "en_US",
+                .authorization: "Basic \(base64EncodedCredentials)"
+            ],
+            queryParameters: nil,
+            body: body
+        )
+        
+        let httpResponse = try await apiClient.fetch(request: restRequest)
+        return try HTTPResponseParser().parse(httpResponse, as: ConfirmPaymentSourceResponse.self)
+    }
```
KunJeongPark, line 42: So this class is strictly for confirm-payment-source endpoint, right?

PayPal.xcodeproj/project.pbxproj
jaxdesmarais: Yeah this can totally be a 🏈 if it makes more sense to be included in one of your follow up PRs!

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
jaxdesmarais: Based on the class description I think this would be anything under `v2/checkout/orders` in PPaaS. Right now that's just `confirm-payment-source` in our SDK but it seems like that's the intent of how we expect this class to grow. Is my  understanding correct @scannillo?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
scannillo: Yay! Yeh the nested structs were getting unruly

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
scannillo: Great idea! Done in this commit bc2213bc5e842825037ba0a83ff32dc597c626a3

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
scannillo: I updated cardholderName to use `encodeIfPresent` in this commit - bc2213bc5e842825037ba0a83ff32dc597c626a3

PayPal.xcodeproj/project.pbxproj
scannillo: Awesome, updated the PR description with that as a follow-up PR 👍

PayPal.xcodeproj/project.pbxproj
scannillo: See https://github.com/paypal/paypal-ios/pull/177#discussion_r1290526289

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
scannillo: Exactly what Jax said! Huge kudos to @sshropshire for coming up with the idea of separate classes per API 🎊

Sources/CorePayments/Networking/HTTPRequest.swift
```diff
@@ -0,0 +1,23 @@
...
+    let method: HTTPMethod
+    let body: Data?
+    let headers: [HTTPHeader: String]
+    
+    /// :nodoc:
+    public init(
+        url: URL,
+        method: HTTPMethod,
+        body: Data?,
+        headers: [HTTPHeader: String]
```
jaxdesmarais: General q, do we have a preference for order in these cases? Since `body` is optional I wonder if it'd read better to be last. I generally prefer required > optional, but don't know that we've ever solidified a style.

jaxdesmarais (approved):
😍  thanks for tackling the beast of this refactor - this is going to make adding to our networking layer in the subsequent PRs so much easier.

Sources/CorePayments/Networking/HTTPRequest.swift
scannillo: Yeh, I'm with you - optional last feels nice to me. I updated it to be alphabetical with optionals last in this commit c435d0a5cfafc2626b8768b8f50ba04a3fcd9d54

scannillo: Yeh we could add this to our style guide too

jaxdesmarais: Yeah I think it'd be good to style guide! We've been all over the place with it on both BT and PPCP it would be great to have a ⭐ way we wanted to do it

sshropshire: This is in [Swift API guidelines](https://www.swift.org/documentation/api-design-guidelines/#parameter-with-defaults-towards-the-end) too!

Screenshot 2023-08-10 at 3 07 29 PM

sshropshire (approved):

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
```diff
@@ -0,0 +1,49 @@
...
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        let body = try encoder.encode(confirmData)
+        
+        let restRequest = RESTRequest(
+            path: "/v2/checkout/orders/\(cardRequest.orderID)/confirm-payment-source",
+            method: .post,
+            headers: [
+                .contentType: "application/json", .acceptLanguage: "en_US",
+                .authorization: "Basic \(base64EncodedCredentials)"
```
sshropshire: q: Is there a way we can make Auth header assignment more DRY?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
```diff
@@ -4,132 +4,89 @@ import CorePayments
...
     }
     
-    private struct PaymentSource: Encodable {
+    // MARK: - Custom Encoder
+    
+    func encode(to encoder: Encoder) throws {
+        var topLevel = encoder.container(keyedBy: TopLevelKeys.self)
         
-        let card: EncodedCard
+        var applicationContext = topLevel.nestedContainer(keyedBy: ApplicationContextKeys.self, forKey: .applicationContext)
```
sshropshire, line 68: qq: What does a custom encoder for `ConfirmPaymentSourceRequest` do vs. the default implementation provided by the standard library?

scannillo: It's mainly just for readability. If we use the default, then we have to format our structs & properties to be nested in the same exact format that we want the request to be encoded as ([see previous implementation](https://github.com/paypal/paypal-ios/blob/main/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift)).

Whereas with the custom encoder style ([see file](https://github.com/paypal/paypal-ios/blob/construct-url-request-in-http/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift)), we can basically flatten all of our properties & keys and IMO this init is easier to digest the format of the POST versus the nested structs.

It helps to view the files side-by-side, the diff in the PR here is a little hard to decipher.

scannillo: Also, we were doing the custom encoder already to format the[ analytics post body ](https://github.com/paypal/paypal-ios/blob/main/Sources/CorePayments/AnalyticsEventData.swift), for the same reason to avoid needing to nest a bunch of structs and properties just to satisfy the API format requirement.

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
scannillo: Great question - we definitely can. Though there is one case where the PPass API type does influence the headers, and that is to _not_ send `Basic clientID` for REST post requests to `/v1/tracking`. All other REST requests should send that header. (But you know, PPass gonna PPass, so who knows what's coming out way lol)

Do you think it would be cleaner to include a param on the `RESTRequest` to disable sending that header instead of how it is currently?

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
sshropshire: Yeah it makes sense for the most part. This [implementation](https://github.com/paypal/paypal-ios/blob/main/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift) has some additional logic also besides `Codable` conformance. I'm wondering if taking out all of the default value initialization would allow us to keep the default codable implementation. That would be a win if we could write less code.

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
sshropshire: I think the best solution is allow API classes to make requests without having to think about auth–that'll make it easy to create new API classes in the future.

That is interesting about the tracking API though. On Android, for TrackingAPI we do still send the auth header. It isn't perfect since that header isn't required, but sending the header doesn't seem to cause any issues and it keeps the design clean. Most likely it will be ignored by the server.

scannillo: I agree, will update this PR!

Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift
scannillo: Yep - all of that extra logic was removed from [`ConfirmPaymentSourceRequest`](https://github.com/paypal/paypal-ios/blob/construct-url-request-in-http/Sources/CardPayments/APIRequests/ConfirmPaymentSourceRequest.swift) (in a future PR we're gonna nuke that `APIRequest` protocol entirely 🔥. I only left it in to make refactoring the tests a little easier. But it isn't being used in this PR anymore. It was the one responsible for giving too much responsibility to the Request types). So in this PR we now have data objects that _only_ represent the model which is nice. It doesn't do anything more.

If we want the flat structure though, and to avoid those nested structs, we will need to implement `encode()`.

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
scannillo: Addressed in commit 311ac819705d274487edbf873db3d6a0321ebce4

scannillo:
These UML Class Diagrams show a visual representation of the redesign introduced in this PR. I will create a formal write-up describing these diagrams & the benefits to the new structure once all PRs are complete. 

| Before      | After |
| ----------- | ----------- |
| PPCP iOS Networking Before     |     PPCP iOS Networking After |

PR #178, opened 2023-08-09
Remove unneeded APIRequest conformance from Demo

scannillo:
### Reason for changes

Make it clear that the Demo does not rely on the SDK's internal `APIRequest` protocol. 

_We're also planning to remove this protocol entirely (👀  #177)_

### Summary of changes

- Remove unneeded conformance to `APIRequest` on `ClientIDRequest` in demo app

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Demo/Demo/Models/ClientIDRequest.swift
+++ b/Demo/Demo/Models/ClientIDRequest.swift
@@ -1,7 +1,7 @@
 import Foundation
 import CorePayments
 
-struct ClientIDRequest: APIRequest {
+struct ClientIDRequest {
 
     typealias ResponseType = ClientIDResponse
 


jaxdesmarais (approved):

Demo/Demo/Models/ClientIDRequest.swift
```diff
@@ -1,7 +1,7 @@
 import Foundation
 import CorePayments
 
-struct ClientIDRequest: APIRequest {
+struct ClientIDRequest {
```
KunJeongPark, line 5: Sammy, on my draft PR for vault without purchase, ( https://github.com/paypal/paypal-ios/blob/de43d87c98acc75c5fcb094a62b58d94c248850b/Demo/Demo/Networking/DemoMerchantAPI.swift#L185) I had reused this createUrlRequest for setup token call with APIRequest protocol. I have to have some other protocol so we can take in different request structs to create urlRequest for demo app.

What I can do right now is just create a separate createUrlRequest functions for both REST calls

KunJeongPark: I addressed it in latest commit in the PR. It's redundant but I will consolidate if we have another protocol.

scannillo: The Demo should never have been using `APIRequest`. Awesome, I'll check out your new updates!

PR #182, opened 2023-08-11
[PART 2] Networking Refactor - Unify `REST` & `GraphQL` networking implementation

scannillo:
### Reason for changes

Follow up to #177.

This PR:
- Standardizes the interface for making REST & GraphQL requests from the feature client's perspective
- Unifies REST & GraphQL requests to use the same underlying HTTP, URLRequest, and URLSession implementations
- Repurposes `APIClient` (which previously just passed requests along to HTTP) into a true `NetworkingCoordinator` that transforms REST & GraphQL --> HTTP.

### Summary of changes

- Add `fetch(request: GraphQLRequest)` method to `APIClient`
   - This class now has only 2 public functions, an equivalent for REST & GraphQL
- Remove `GraphQLClient` and other no-longer-used GraphQL files
- Rename `GraphQLQueryResponse` --> `GraphQLHTTPResponse`

### Next
- Rename `APIClient` to `NetworkingClient`
- Unit Tests
- Add `VaultPaymentMethodTokensAPI` class (waiting on Victoria's #172 PR to merge)
- Update parsing layer for GraphQL (see TODOs)

### Note
Tests are expected to fail.

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -51,12 +52,26 @@ public class APIClient {
         return try await http.performRequest(httpRequest)
     }
     
-    // TODO: - Add GraphQL equivalent request type & function
-    // public func fetch(request: GraphQLRequest) async throws -> HTTPResponse { }
+    /// :nodoc:
+    public func fetch(request: GraphQLRequest) async throws -> HTTPResponse {
+        let url = try constructGraphQLURL(queryName: request.queryNameForURL)
+
```
jaxdesmarais, line 51: not sure if we need this extra space but up to you:
```suggestion
```

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -51,12 +52,26 @@ public class APIClient {
...
+        let postBody = GraphQLHTTPPostBody(query: request.query, variables: request.variables)
+        let postData = try JSONEncoder().encode(postBody)
+                
+        let httpRequest = HTTPRequest(
```
jaxdesmarais: If we single line this is it above our linter line length? May be nice to just group all of the properties together but not sure if this is more readable this way.

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -51,12 +52,26 @@ public class APIClient {
...
+            body: postData
+        )
+        
+        return try await http.performRequest(httpRequest)
+    }
     
     // MARK: - Private Methods
     
-    private func constructURL(path: String, queryParameters: [String: String]) throws -> URL {
+    private func constructRestURL(path: String, queryParameters: [String: String]) throws -> URL {
```
jaxdesmarais: I think REST is typically all caps right?

sshropshire (approved):
🔥

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -70,4 +85,16 @@ public class APIClient {
         
         return url
     }
+    
+    private func constructGraphQLURL(queryName: String? = nil) throws -> URL {
+        guard let queryName else {
+            return coreConfig.environment.graphQLURL
+        }
+        
+        if let url = URL(string: coreConfig.environment.graphQLURL.absoluteString + "?" + queryName) {
```
jaxdesmarais: Above in the `constructRestURL` method we use a guard then return the URL. Should we use that here instead to keep the methods consistent vs an `if let`?

Sources/CorePayments/Networking/GraphQL/GraphQLErrorResponse.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// Used to parse error message details out of GraphQL HTTP response body
+struct GraphQLErrorResponse: Decodable {
+    
+    let error: String
+    let correlationId: String?
```
jaxdesmarais: I think we want to do `ID` for the style guide. Not sure if that would require us to use a custom key though, but should be possible?
```suggestion
    let correlationID: String?
```

Sources/CorePayments/Networking/APIClient.swift
scannillo: It is, I had `constructRESTURL()` before, but that looked odd to me. Which do you think is better?

Sources/CorePayments/Networking/APIClient.swift
scannillo: It does not violate our line length, but I thought it was a little cleaner. Happy to go either way though!

Sources/CorePayments/Networking/APIClient.swift
scannillo: Done in 2d5a36a4e3bb0ce8031c7a298cab85d5f73e8bf7!

Sources/CorePayments/Networking/GraphQL/GraphQLErrorResponse.swift
scannillo: It would require the custom keys, but it's possible! Addressed in 4ef956b78d8340d6eff9894965817d81e3be1482

Sources/CorePayments/Networking/APIClient.swift
scannillo: I did it intentionally to group `postBody` and `postData` together, but down to remove this newline if it's harder to read

Sources/CorePayments/Networking/APIClient.swift
jaxdesmarais: Either way is fine, was more just curious. If it makes more sense to have things grouped together a particular way that's fine too.

Sources/CorePayments/Networking/APIClient.swift
jaxdesmarais: I think we have something similar with `constructGraphQLURL` as well. Agree both look a little awkward but not sure if there is a way around it. 🤷 It's private so not a super strong preference.

scannillo: Ok, I updated to `constructRESTURL()` d3fe18806d8df1cd523662e998ecb2ff46263ba3

Sources/CorePayments/Networking/APIClient.swift
scannillo: Not sure what you mean by "group all of these properties together"? You mean single line it?

scannillo: Ok I single-lined both the `HTTPRequest()` definitions (see commit 7417e153d6284d7ffb41c7a14487b812e743239f)

jaxdesmarais: > Not sure what you mean by "group all of these properties together"? You mean single line it?

Ah, sorry was referencing [this comment](https://github.com/paypal/paypal-ios/pull/182#discussion_r1296358473)

scannillo: That link takes me back to:

> If we single line this is it above our linter line length? May be nice to just group all of the properties together but not sure if this is more readable this way.

Do  you mean to remove the newlines between [these variables](https://github.com/paypal/paypal-ios/blob/ff3ed9b993134432f620d30c407e34e570f51e4c/Sources/CorePayments/Networking/APIClient.swift#L37-L45)?

jaxdesmarais: Weird, idk why GH is linking to that. This comment above is the one I was trying to reference:

> I did it intentionally to group postBody and postData together, but down to remove this newline if it's harder to read

Not a blocker though, whatever you feel is best

jaxdesmarais (approved):
🚀

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -29,34 +30,36 @@ public class APIClient {
...
             headers[.contentType] = "application/json"
         }
         
-        let httpRequest = HTTPRequest(
-            headers: headers,
-            method: request.method,
-            url: url,
-            body: request.body
-        )
+        let httpRequest = HTTPRequest(headers: headers, method: request.method, url: url, body: request.body)
```
KunJeongPark, line 44: Very cool!

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -29,34 +30,36 @@ public class APIClient {
...
         return try await http.performRequest(httpRequest)
     }
     
-    // TODO: - Add GraphQL equivalent request type & function
-    // public func fetch(request: GraphQLRequest) async throws -> HTTPResponse { }
+    /// :nodoc:
+    public func fetch(request: GraphQLRequest) async throws -> HTTPResponse {
+        let url = try constructGraphQLURL(queryName: request.queryNameForURL)
+
+        let postBody = GraphQLHTTPPostBody(query: request.query, variables: request.variables)
```
KunJeongPark, line 52: 👏🏼

PR #183, opened 2023-08-15
Vault without Purchase UI and View Model

KunJeongPark:
### Summary of changes

- Create new feature in settings menu for `CardVault`
- Create new view model, `CardVaultViewModel` for standalone vaulting 
- Create SwiftUI `CardVaultView` and subviews. 
- Create ErrorView to display errors on SwiftUI views.
- Make resultViews observe error or success state to toggle success view or error view.
  Too many condition checking on main view caused SwiftUI compiler error. 
   I moved logic to display success or error on the resultViews themselves.
- Make `CardVaultState` equatable so changes can be observed when scrolling to bottom with each new resultView
- Diffs on last 4 files are mostly spacing issues from resolving merge conflict with parent branch PR that got merged. 
   They're just removal of trailing white spaces and adding extra line. 

Notes:
- I like this UI below for purpose of standalone vaulting, adapted from Android, for displaying all of the results that can be 
  viewed on one scrollable page, comparing setup token and customer ID and card information.

https://github.com/paypal/paypal-ios/assets/9273272/b49d0e33-abe6-455c-a2d2-f405807c2022

Example of an error view for `UpdateSetupTokenResultView` on the bottom:
 <p float="left">
 <image src = "https://github.com/paypal/paypal-ios/assets/9273272/82ec5469-5602-4b78-98a8-b7701d1c557d" width="400" />
</p>

- There will be a PR up for taking out UIKit layers since it limits navigation from SwiftUI views.
   SwiftUI will make changes to UI with additional functionalities and testing requirements easier.

TODO for next PR for refactoring Demo app:
- Starting point should be main SwiftUI view, rather than a settings menu.
  First set of selections should be feature, merchant integration type, environment
- Have separate viewModels, main views for each feature.
- Have alternate error states, views for each result view.
  This allows customization of the views rather than having simple text in status bar or Xcode console.
- Question about `BaseViewModel` and possibly just keeping `CoreConfig` (environment and clientID), 
   integration type, published properties here that the other view models can inherit.
  I think these property selections should be at top level and inherited by all the feature view models.
  Make different viewModels for each feature (payment method, vaulting) views would allow customization of
  each view and changes easier to manage, understand.
- `SwiftUINativeCheckoutDemo` view uses toggling views in the body depending on state in `PayPalViewModel`. 
    All of the views are defined in this file.
   `PayPalViewModel` has its own `getClientID` function rather than having reference to `BaseViewModel`.
    We need to think about passing in different parameters for Native Checkout order creation 
    as well to support different integrations. We might want to add order creation view so we can enter 
    custom values rather than autogenerating for testing purposes.
    Initial `ClientIDView` is wrapped in `NavigationView` there but it doesn't do anything.

### Checklist

### Authors

- @KunJeongPark


Demo/Demo/ViewModels/CardVaultViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+        let cleanedCardNumber = cardNumber.replacingOccurrences(of: " ", with: "")
+        let cleanedExpirationDate = expirationDate.replacingOccurrences(of: " / ", with: "")
+
+        let enabled = cleanedCardNumber.count >= 15 && cleanedCardNumber.count <= 19
+        && cleanedExpirationDate.count == 4 && cvv.count >= 3 && cvv.count <= 4
+        return enabled
+    }
+
+    func updateUpdateSetupTokenResult(vaultResult: CardPayments.CardVaultResult) {
+        DispatchQueue.main.async {
```
KunJeongPark, line 87: This function could probably use a better name. The viewModel state needs to be updated so the SwiftUI view gets updated with result view and only way we get the "done" status in the demo app from the SDK is through the delegate function.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                    if let paymentTokenResponse = cardVaultViewModel.state.paymentTokenResponse {
+                        PaymentTokenResultView(paymentTokenResponse: paymentTokenResponse)
+                            .id("paymentTokenResultView")
+                        Button("Reset") {
+                            cardVaultViewModel.resetState()
+                        }
+                        .foregroundColor(.white)
+                        .padding()
+                        .frame(maxWidth: .infinity)
```
sshropshire: qq: What does maxWidth `.infinity` do? I'm n 👀 b status with SwiftUI.

KunJeongPark: It expands the view as much as possible to its container view with the padding. Without it, sometimes you get it just big enough for the content the view contains.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    private let cardFormatter = CardFormatter()
+
+    // MARK: Views
+
+    var body: some View {
+        ScrollView {
+            ScrollViewReader { scrollView in
+                VStack(spacing: 16) {
+                    CreateSetupTokenView(baseViewModel: baseViewModel, cardVaultViewModel: cardVaultViewModel)
```
sshropshire: nit: There may be a way to provide an `onSubmit` action to `CreateSetupTokenView` as an `@escaping` closure. Then we could remove the view model parameters from the constructor.

sshropshire: Similar to [this post](https://developer.apple.com/forums/thread/718378). I'm not sure if this is kosher for SwiftUI, but it's a common delegation pattern we use in Compose for Android.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
```diff
@@ -0,0 +1,71 @@
...
+                    if let setupTokenResponse = cardVaultViewModel.state.setupTokenResponse {
+                        SetupTokenResultView(setupTokenResponse: setupTokenResponse)
+                        let setupToken = setupTokenResponse.id
+                        UpdateSetupTokenView(baseViewModel: baseViewModel, cardVaultViewModel: cardVaultViewModel, setupToken: setupToken)
+                            .id("updateSetupTokenView")
```
sshropshire: another noob q: Does the `id` modifier help prevent re-renders?

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: Yes, we can do that but I still need the CardVaultViewModel to get the loading state (which I plan to move to viewModel) or error state from the CreateSetupTokenViews and other views as well. Feels like it would be a lot on the parent view.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
```diff
@@ -0,0 +1,59 @@
...
+                    }
+                }
+            }
+            .foregroundColor(.white)
+            .padding()
+            .frame(maxWidth: .infinity)
+            .background(.blue)
+            .cornerRadius(10)
+        }
+        .padding()
```
sshropshire, line 44:
```suggestion
```
Duplicate modifier.

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
```diff
@@ -0,0 +1,40 @@
...
+            Text("\(paymentTokenResponse.paymentSource.card.brand ?? "")")
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("Card Last 4")
+                .bold()
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("\(paymentTokenResponse.paymentSource.card.lastDigits)")
+                .frame(maxWidth: .infinity, alignment: .leading)
+        }
+        .frame(maxWidth: .infinity)
+        .padding()
```
sshropshire, line 35:
```suggestion
```
+1 potential duplicate.

Demo/Demo/SwiftUIComponents/CardVaultViews/SetupTokenResultView.swift
```diff
@@ -0,0 +1,35 @@
...
+            Text("\(setupTokenResponse.customer?.id ?? "")")
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("Status")
+                .bold()
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("\(setupTokenResponse.status)")
+                .frame(maxWidth: .infinity, alignment: .leading)
+        }
+        .frame(maxWidth: .infinity)
+        .padding()
```
sshropshire, line 33:
```suggestion
```
+1

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenResultView.swift
```diff
@@ -0,0 +1,30 @@
...
+            Text("\(updateSetupTokenResponse.id)")
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("Status")
+                .bold()
+                .frame(maxWidth: .infinity, alignment: .leading)
+            Text("\(updateSetupTokenResponse.status)")
+                .frame(maxWidth: .infinity, alignment: .leading)
+        }
+        .frame(maxWidth: .infinity)
+        .padding()
```
sshropshire, line 31:
```suggestion
```
Not trying to be annoying lol.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,82 @@
...
+                // handing off control to CardClient's vault function.
+				isLoading = false
+            }
+            .foregroundColor(.white)
+            .padding()
+            .frame(maxWidth: .infinity)
+            .background(.blue)
+            .cornerRadius(10)
+        }
+        .padding()
```
sshropshire, line 64:
```suggestion
```
Want to make sure we get all of these.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,82 @@
...
+                            print("Error getting Config or Card")
+                            isLoading = false
+                        }
+                    }
+                }
+            }
+            .onChange(of: cardVaultViewModel.state.updateSetupToken?.id) { _ in
+                // Updating is done through CardVaultDelegate function after
+                // handing off control to CardClient's vault function.
+				isLoading = false
```
sshropshire: Indentation is off some.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,82 @@
...
+                cardNumberText: $cardNumberText,
+                expirationDateText: $expirationDateText,
+                cvvText: $cvvText
+            )
+
+            let card = baseViewModel.createCard(
+                cardNumber: cardNumberText,
+                expirationDate: expirationDateText,
+                cvv: cvvText
+            )
```
sshropshire, line 38: nit: This feels like it may belong in the constructor?

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,82 @@
...
+                    }
+                }
+            }
+            .onChange(of: cardVaultViewModel.state.updateSetupToken?.id) { _ in
+                // Updating is done through CardVaultDelegate function after
+                // handing off control to CardClient's vault function.
+				isLoading = false
+            }
+            .foregroundColor(.white)
+            .padding()
```
sshropshire:
```suggestion
```
+1. I'm marking these. If you want to change them in a separate commit that's cool too.

Demo/Demo/SwiftUIComponents/SwiftUICardDemo.swift
```diff
@@ -22,18 +22,7 @@ struct SwiftUICardDemo: View {
...
-                    }
-                FloatingLabelTextField(placeholder: "Expiration Date", text: $expirationDateText)
-                    .onChange(of: expirationDateText) { newValue in
-                        expirationDateText = cardFormatter.formatFieldWith(newValue, field: .expirationDate)
-                    }
-                FloatingLabelTextField(placeholder: "CVV", text: $cvvText)
-                    .onChange(of: cvvText) { newValue in
-                        cvvText = cardFormatter.formatFieldWith(newValue, field: .cvv)
-                    }
+                CardFormView(cardNumberText: $cardNumberText, expirationDateText: $expirationDateText, cvvText: $cvvText)
```
sshropshire, line 25: 🔥

Demo/Demo/ViewModels/CardVaultViewModel.swift
sshropshire: On Android we use a Kotlin feature similar to Swift's computed properties, that could be a good fit here as well.

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
KunJeongPark: Oh, maybe that's why it gets weirdly fat after I press it. no that's not it. But good catch.

sshropshire (approved):
🌊 Awesome PR! My nit's / q's are mostly out of curiosity. It's nice to see that Compose has a lot in common with SwiftUI.

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
sshropshire: Could be that's interesting.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: It was strictly for the purpose of scrolling to those positions so the new views added are visible.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
sshropshire: Nice yeah I'm sure it'll be a mixture of different stuff. I saw SwiftUI has some concept of an environment object too that looks cool.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenResultView.swift
KunJeongPark: Not at all. great catch

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
sshropshire: Ah yeah I saw that shortly after I made this common in a web search. I wish there was a `scrollToBottom` method that would be legit, but if ID's work that's the wave 🌊 🏄 .

KunJeongPark:
Thank you. Yes, I was really surprised how similar they looked when I saw your PRs.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: Yeah, they're really easy to use but probably use with caution for work. I should just probably pass in merchant integration instead of reference to baseVM into CreateSetupTokenView though.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: Yeah, maybe you can do that with Geometry Reader. I found a way to scroll to bottom with a blank Text item. It didn't work with an EmptyView.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
KunJeongPark: This was interesting because this is actually padding from outer edge of VStack and its contents and the .padding(5) is around the background rectangle and the parent view, so I do need both.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: I thought that since I am also using baseVM's `getCoreConfig` in this view, I could just have it here so it doesn't clutter up the main `CardVaultView`. I feel like with `CardFormView` here, card being created where the card information is entered makes sense.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
sshropshire: Ok I'm not as familiar with the `padding` modifier, but I'll say the part that jumped out to me wasn't necessarily that it was a duplicate, but the fact that it's called with no args e.g. `padding()`.

KunJeongPark: There is a default number, I think it's 10.

sshropshire: Oh ok cool.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -44,7 +47,7 @@ final class DemoMerchantAPI {
                 environment: DemoSettings.environment,
                 selectedMerchantIntegration: selectedMerchantIntegration
 			)
-		let data = try await data(for: urlRequest)
+			let data = try await data(for: urlRequest)
```
jaxdesmarais: I think we can revert this indent

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
```diff
@@ -0,0 +1,56 @@
...
+    public init(cardVaultViewModel: CardVaultViewModel, selectedMerchantIntegration: MerchantIntegration, setupToken: String) {
+        self.cardVaultViewModel = cardVaultViewModel
+        self.selectedMerchantIntegration = selectedMerchantIntegration
+        self.setupToken = setupToken
+    }
+
+    var body: some View {
+        VStack(spacing: 16) {
+            HStack {
+                Text("Create a Permanent Payment Method Token")
```
jaxdesmarais: nit: is there a name that PayPal uses in their public docs for this? It seems like maybe "create payment token" would match the public API more but not sure if you've seen something different used in the docs.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
```diff
@@ -0,0 +1,56 @@
...
+                            print("Error in getting setup token. \(error.localizedDescription)")
+                        }
+                    }
+                }
+            }
+            .foregroundColor(.white)
+            .padding()
+            .frame(maxWidth: .infinity)
+            .background(.blue)
+            .cornerRadius(10)
```
jaxdesmarais: I see we use a lot of the same modifiers across our views. Is there a way to create a custom modifier that applies things that we use often vs having to add some of these to every view? Not sure if this is common in SwiftUI but could be worth exploring.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
```diff
@@ -0,0 +1,56 @@
...
+        }
+        .padding()
+        .background(RoundedRectangle(cornerRadius: 10).stroke(Color.gray, lineWidth: 2))
+        .padding(5)
```
jaxdesmarais: Do we need the `padding` modifier twice here?

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
```diff
@@ -0,0 +1,51 @@
...
+    @ObservedObject var cardVaultViewModel: CardVaultViewModel
+
+    var body: some View {
+        switch cardVaultViewModel.state.paymentTokenResponse {
+        case .idle, .loading:
+            EmptyView()
+        case .loaded(let paymentTokenResponse):
+            getSucessView(paymentTokenResponse: paymentTokenResponse)
+        case .error(let message):
+            ErrorView(errorText: message)
```
jaxdesmarais: nit: it might be clearer to rename `errorText` to `message` or `errorMessage`. Then this would read as:
```swift
ErrorView(message: message)
// OR
case .error(let errorMessage):
    ErrorView(errorMessage: errorMessage)
```
What do you think?

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
```diff
@@ -0,0 +1,51 @@
...
+    func getSucessView(paymentTokenResponse: PaymentTokenResponse) -> some View {
+        VStack(spacing: 16) {
+            HStack {
+                Text("Payment Token")
+                    .font(.system(size: 20))
+                Spacer()
+            }
+            Text("ID")
+                .bold()
+                .frame(maxWidth: .infinity, alignment: .leading)
```
jaxdesmarais: Can we make a custom `Text` view that applies these modifiers across the board?

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,76 @@
...
+                cvv: cvvText
+            )
+
+            ZStack {
+                if case .loading = cardVaultViewModel.state.updateSetupTokenResponse {
+                    ProgressView()
+                        .progressViewStyle(CircularProgressViewStyle(tint: Color.white))
+                        .background(Color.black.opacity(0.4))
+                        .cornerRadius(10)
+                        .frame(maxWidth: .infinity)
```
jaxdesmarais: Since we use this across several views, can we extract this into a custom view that way we can ensure our ProgressViews are always consistent?

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,76 @@
...
+                }
+                Button("Vault Card") {
+                    Task {
+                        let config = await baseViewModel.getCoreConfig()
+                        if let config = config, let card = card {
```
jaxdesmarais:
```suggestion
                        if let config, let card {
```

Demo/Demo/SwiftUIComponents/ErrorView.swift
```diff
@@ -0,0 +1,17 @@
+import SwiftUI
+
+struct ErrorView: View {
+
+    let errorText: String
+
+    var body: some View {
+        VStack {
+			Text("\(errorText)")
```
jaxdesmarais: It looks like this `Text` may be indented too far

Demo/Demo/ViewModels/CardVaultState.swift
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+}
+
+enum LoadingState<T: Decodable & Equatable>: Equatable {
+
+    case idle
+    case loading
+	case error(message: String)
+	case loaded(_ value: T)
```
jaxdesmarais: These cases have an extra indent

Demo/Demo/ViewModels/CardVaultViewModel.swift
```diff
@@ -0,0 +1,109 @@
...
+            }
+            throw error
+        }
+    }
+
+    func vault(
+        config: CoreConfig,
+        card: Card,
+        setupToken: String
+    ) async {
```
jaxdesmarais: We can probably single line this and be under our linter line length

jaxdesmarais (commented):
👏  this is a really great addition and start of the refactor. Great job cleaning things up overall.

KunJeongPark:
Thank you Jax! I am going through your comments now. I have some ideas on how to separate out the BaseViewModel for the next PR to remove all UIKit components. I'd like to get everyone's input on it after this PR.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
KunJeongPark: I was copying Steven's UI on Android. I think his thinking was to guide whoever is using this for testing purposes what each step is for if they're not familiar with concept of setup token and permanent payment token.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
KunJeongPark: That's true. I should create a custom button style.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
KunJeongPark: I'm still uncertain where these paddings take effect but from putting color red on the VStack, I saw that the first modifier puts the default padding of 10 between the content in VStack and outline of the Stack itself, whereas the second padding adds a little space between the rounded outline and the main view outline.

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
KunJeongPark: I agree.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
jaxdesmarais: This may be dumb question but are there non permanent payment tokens? I think the word permanent is throwing me off a bit.

KunJeongPark: Not at all. We have SetupToken which is temporary safe token for the client. This token is first created in merchant server with empty card info and then updated in client with card info, then sent back to merchant for them to create a permanent payment token.

Demo/Demo/Networking/DemoMerchantAPI.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardVaultViews/PaymentTokenResultView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/ErrorView.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/ViewModels/CardVaultState.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/ViewModels/CardVaultViewModel.swift
KunJeongPark: addressed in commit 121c4eadb22995e34dbf68e2aee497baa92e2249

Demo/Demo/SwiftUIComponents/CardFormView.swift
```diff
@@ -0,0 +1,42 @@
...
+                }
+        }
+    }
+}
+
+struct CardFormView_Previews: PreviewProvider {
+    
+    @State static var mockCardNumberText: String = ""
+    @State static var mockExpirationDateText: String = ""
+    @State static var mockCvvText: String = ""
```
jaxdesmarais: For our previews should we add values to these string so we can quickly see in the preview what it looks like?

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
jaxdesmarais: Cool - I think in that case we should go with `Create a Payment Method Token` here and remove the word "Permanent" since there isn't an alternate type of payment token.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
```diff
@@ -0,0 +1,48 @@
...
+            }
+            .frame(maxWidth: .infinity)
+            .font(.headline)
+            FloatingLabelTextField(placeholder: "Vault Customer ID (Optional)", text: $vaultCustomerID)
+            ZStack {
+                Button("Create Setup Token") {
+                    Task {
+                        do {
+                            try await cardVaultViewModel.getSetupToken(
+                                customerID: vaultCustomerID,
```
jaxdesmarais: Right now if we don't pass in a customer ID this step fails since we don't check the value and default to an empty string. We can make this change to make the customer ID optional in demo app:
```suggestion
                                customerID: vaultCustomerID.isEmpty ? nil : vaultCustomerID,
```

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
jaxdesmarais: Gotcha - it looks like the second modifier for padding may actually for the RoundedRectangle background. It think adding it like this would clear up what it's for quite a bit (same with the other views that do this as well):
```swift
        .background(
            RoundedRectangle(cornerRadius: 10)
                .stroke(Color.gray, lineWidth: 2)
                .padding(5)
        )
```
Test this to verify but that's my guess to what is happening.

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -0,0 +1,68 @@
...
+                cardNumber: cardNumberText,
+                expirationDate: expirationDateText,
+                cvv: cvvText
+            )
+
+            ZStack {
+                Button("Vault Card") {
+                    Task {
+                        let config = await baseViewModel.getCoreConfig()
+                        if let config, let card = card {
```
jaxdesmarais:
```suggestion
                        if let config, let card {
```

Demo/Demo/SwiftUIComponents/CircularProgressView.swift
```diff
@@ -0,0 +1,19 @@
+import SwiftUI
+
+struct CircularProgressView: View {
```
jaxdesmarais, line 3: 😍

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
KunJeongPark: Oh good catch!

stash_diff_cr_e2d
KunJeongPark, line 1: Oops, deleted in next commit.

Demo/Demo/SwiftUIComponents/CardVaultViews/CreatePaymentTokenView.swift
KunJeongPark: Addressed in commit c81289d70593515ef867beadd4bd42a8c596572f

Demo/Demo/SwiftUIComponents/CardFormView.swift
KunJeongPark: Addressed in c81289d70593515ef867beadd4bd42a8c596572f

Demo/Demo/SwiftUIComponents/CardVaultViews/CreateSetupTokenView.swift
KunJeongPark: Addressed in c81289d70593515ef867beadd4bd42a8c596572f

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: addressed in c81289d70593515ef867beadd4bd42a8c596572f

Demo/Demo/Models/SetupTokenResponse.swift
```diff
@@ -1,6 +1,16 @@
 import Foundation
 
-struct SetUpTokenResponse: Decodable {
+struct SetUpTokenResponse: Decodable, Equatable {
+
+    static func == (lhs: SetUpTokenResponse, rhs: SetUpTokenResponse) -> Bool {
+        lhs.id == rhs.id
+    }
```
jaxdesmarais, line 7: Can we move this under the properties and add a `// MARK: - Equatable Conformance` above it?

jaxdesmarais (approved):
Really great work overall on this Victoria!! 👏  Love seeing our SwiftUI views. One very small file organization comment but otherwise 🚢 .

Demo/Demo/Models/SetupTokenResponse.swift
KunJeongPark: Oops, merged before seeing this. I will put it on the next PR.

PR #188, opened 2023-08-23
[PART 3] Networking Refactor - Add `VaultPaymentTokensAPI`

scannillo:
### Reason for changes

Restructure vault w/o purchase graphQL call to use new networking layer & add PPaaS API layer.

### Summary of changes

- Add GraphQL response parsing ability to `HTTPResponseParser`
- Move Post body encoding from `<PPaaSAPI>` layers to `APIClient` to unify process across GraphQL & Rest requests
   - Require `Encodable` type in `RESTRequest` & `GraphQLRequest`s instead of `Data`
- Add `VaultPaymentTokensAPI` PPaaS layer
   - Move vault w/o purchase API call here
- Tests
    - Remove `GraphQLClient_Tests` now that `GraphQLClient` is gone
    - Update `HTTPResponseParser_Tests`
    
### Next

- Unit tests for work in prior PRs
- Move JSONEncoding into custom class
- Reorganize file structure

### Note

- Tests still expected to fail

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


scannillo:
See slightly updated class diagram:

PPCP iOS Networking (1)

Sources/CardPayments/APIRequests/VaultPaymentTokensAPI.swift
```diff
@@ -0,0 +1,58 @@
...
+            ) {
+                updateVaultSetupToken(
+                    clientId: $clientID
+                    vaultSetupToken: $vaultSetupToken
+                    paymentSource: $paymentSource
+                ) {
+                    id,
+                    status,
+                    links {
+                        rel, href
```
jaxdesmarais: nit to match the style of the other portions of the mutation unless there is a reason it needs to be listed side by side:
```suggestion
                        rel, 
                        href
```

Sources/CorePayments/Networking/GraphQL/GraphQLHTTPResponse.swift
```diff
@@ -1,5 +1,5 @@
-/// Used to decode the HTTP reponse body of GraphQL requests
-struct GraphQLHTTPResponse<T: Codable>: Codable {
+/// :nodoc: Used to decode the HTTP reponse body of GraphQL requests
```
jaxdesmarais, line 1: This has 0 relation to this PR, but I wonder if we should consider DocC first for PPCP. 👀

Sources/CorePayments/Networking/HTTPResponseParser.swift
```diff
@@ -4,32 +4,68 @@ import Foundation
...
-            do {
-                let decodedData = try decoder.decode(T.self, from: data)
-                return (decodedData)
-            } catch {
-                throw APIClientError.jsonDecodingError(error.localizedDescription)
-            }
+            return try parseSuccessResult(data, as: T.self)
+        } else {
+            return try parseErrorResult(data, as: T.self)
+        }
```
jaxdesmarais, line 26: Is there ever a case where we couldn't parse the data into either an error or success result?

Sources/CardPayments/APIRequests/VaultPaymentTokensAPI.swift
scannillo: I don't think it needs to be side by side - this was just copied this from our existing vault w/o purchase implementation [here](https://github.com/paypal/paypal-ios/blob/c3dd92290e3da20a2cc857e46f390b5e05245484/Sources/CardPayments/APIRequests/UpdateSetupTokenQuery.swift#L61). Updated - c2d71e773a23fd98cefc25576813bb3cfce1ce3c!

Sources/CorePayments/Networking/HTTPResponseParser.swift
scannillo: The only case I can think of we wouldn't call this function at all, like for analytics. We aren't calling HTTPResponseParser at all from the TrackingEventsAPI, since we only care if it was a successful status or not.

(Example [here](https://github.com/paypal/paypal-ios/blob/networking-refactor/Sources/CorePayments/Networking/AnalyticsService.swift#L54) & [here](https://github.com/paypal/paypal-ios/blob/9aa4eb8031ffd646cbc4aef4343388d99a713494/Sources/CorePayments/Networking/TrackingEventsAPI.swift#L26))

scannillo: I went back and forth on the best way to write this file to be as DRY as possible and landed on this, but totally open to restructuring/ideas to make it more readable!

Sources/CorePayments/Networking/GraphQL/GraphQLHTTPResponse.swift
scannillo: We should give it a whirl! We'll run into that same issue of needing to host a landing page for each of our submodules, right? But it would definitely be easier to start with over BT, I agree!

But in terms of `nodoc` this doesn't actually do anything in PPCP since it isn't set up with Jazzy. We need to write a ticket to update these to `@_documentation(visibility: private)` so at least things look good in the Xcode generated docs.

scannillo: Ticket for this work - DTPPCPSDK-726

jaxdesmarais: Yeah - it would be a good test for structure/format since we don't have anything currently. We can start fresh with DocC vs switching from Jazzy down the road. 🎉

Sources/CorePayments/Networking/HTTPResponseParser.swift
jaxdesmarais: Gotcha, I think how you have it makes sense! I was moreso curious about cases where we hit the else block but cannot parse the data into an error result. Not sure if that would ever happen though.

jaxdesmarais (approved):
So close to a beautiful networking layer 🥲

sshropshire (approved):
🔥

PR #190, opened 2023-08-24
[Part 4] Networking Refactor - CorePayments unit tests

scannillo:
### Reason for changes

This PR re-writes our `CorePaymentsTests` based on the source code changes introduced in [this series ](https://github.com/paypal/paypal-ios/issues?q=label%3Anetworking-refactor+) of networking layer refactor PRs.

### Summary of changes

- Fix bug revealed by unit tests in URL queryParam encoding (though currently no API calls are adding queryParams)
- Update `TrackingEventsAPI` for mock `APIClient` injection
- Update mocks
    - `MockTrackingEventsAPI` (added), `MockAPIClient` (updated), `MockHTTP` (updated)
    - Use `captured` terminology to mirror Android for asserting on function passed params
 - Remove `APIRequest` protocol
 - Fix unit tests `HTTP_Tests`, `APIClient_Tests`, `AnalyticsService_Tests` & add `TrackingEventsAPI_Tests`
 
### Next

- Update unit tests for `CardPayments` module
    - This will include adding a `CheckoutOrdersAPI_Tests` & `VaultPaymentTokensAPI_Tests`
- Reorganize file structure & alphabetize 

### Note

- CI unit test build expected to fail
- Please provide any and all feedback!

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


UnitTests/PaymentsCoreTests/AnalyticsEventData_Tests.swift
```diff
@@ -1,31 +1,28 @@
...
             orderID: "fake-order"
         )
-        
-        sut = try! AnalyticsEventRequest(eventData: fakeAnalyticsEventData)
     }
 
-    func test_httpParameters() throws {
-        let bodyData = sut.body
-        let jsonBody = try? JSONSerialization.jsonObject(with: bodyData!) as? [String: [String: [String: Any]]]
+    func test_EncodedJSONFormat() throws {
```
scannillo:
```suggestion
    func testEncode_properlyFormatsJSON() throws {
```

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -78,12 +79,15 @@ public class APIClient {
...
     
-    private func constructRESTURL(path: String, queryParameters: [String: String]) throws -> URL {
+    private func constructRESTURL(path: String, queryParameters: [String: String]?) throws -> URL {
         let urlString = coreConfig.environment.baseURL.appendingPathComponent(path)
         var urlComponents = URLComponents(url: urlString, resolvingAgainstBaseURL: false)
         
-        queryParameters.forEach {
-            urlComponents?.queryItems?.append(URLQueryItem(name: $0.key, value: $0.value))
+        if let queryParameters {
+            urlComponents?.queryItems = []
```
jaxdesmarais: Do we need this line? I thought it'd just create it for us in the append below if needed.

Sources/CorePayments/Networking/TrackingEventsAPI.swift
```diff
@@ -4,13 +4,29 @@ import Foundation
...
 class TrackingEventsAPI {
+        
+    // MARK: - Internal Properties
+
+    private var apiClient: APIClient
+    var coreConfig: CoreConfig // exposed for testing
+
+    // MARK: - Initializer
+    
+    init(coreConfig merchantConfig: CoreConfig) {
```
jaxdesmarais, line 15: It looks like the other APIs don't have an internal parameter name for `coreConfig`. Do we see an advantage to doing it here vs the other API layers?

Sources/CorePayments/Networking/APIClient.swift
scannillo: Writing the unit tests for this function revealed that the previous implementation wasn't appending the query parameters at all before. If we didn't set to an empty array, `urlComponents.queryItems` was `nil` and the line in the `forEach` did nothing due to nil coalescing.

scannillo: None of our APIs currently set any query params, which is why our tests prior never caught this. But now that we're breaking this up and writing more specific unit tests we were able to see it not working.

Sources/CorePayments/Networking/TrackingEventsAPI.swift
scannillo: Good eye - so for the `TrackingEventsAPI` we're basically overriding the merchant-config to always be `.live` since FPTI doesn't support the traditional PP sandbox URL. 

How can we make that more clear? Maybe we can move the comment on line 16 around somewhere else?

jaxdesmarais: Gotcha, I was more curious around why line 17 couldn't be:
```swift
        self.coreConfig = CoreConfig(clientID: coreConfig.clientID, environment: .live)
```
Then the init can match the other API class inits `init(coreConfig: CoreConfig)`

scannillo: It could be! I just thought using the internal var name of `merchantConfig` might make it more clear that we're overriding it to set our own `coreConfig`. The public API of the init matches the others.

But I'm happy to change it back if this isn't clear

scannillo: The whole init setup for this class is ugly to me. We technically don't need custom inits at all (like it was before). But we _do_ need to set the APIClient & access to the coreConfig if we want to properly unit test (for injecting an APIClient mock & validating that the config env is always live).

That said - totally down for your opinion on what is easiest to read here!

Sources/CorePayments/Networking/APIClient.swift
jaxdesmarais: 🤘🏻 nice - glad our tests caught it!

Sources/CorePayments/Networking/TrackingEventsAPI.swift
jaxdesmarais: Gotcha - imo with your inline comment it makes it clear that we aren't using the environment from the config. But I don't feel super strongly if you feel like the internal parameter name makes it more clear in addition to the comment you have.

Sources/CorePayments/Networking/TrackingEventsAPI.swift
```diff
@@ -4,13 +4,29 @@ import Foundation
 ///
 /// Details on this PayPal API can be found in PPaaS under Infrastructure > Experimentation > Tracking Events > v1.
 class TrackingEventsAPI {
+        
+    // MARK: - Internal Properties
+
+    private var apiClient: APIClient
+    var coreConfig: CoreConfig // exposed for testing
```
jaxdesmarais: I think we can swap the order of these and have the private second

jaxdesmarais (approved):

Sources/CorePayments/Networking/TrackingEventsAPI.swift
scannillo: Done! b0b6d321001d5b8eb36d50f6e6e4b53cba6921a2

KunJeongPark (approved):

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -59,6 +59,7 @@ public class APIClient {
                 
         // TODO: - Move JSON encoding into custom class
         let postBody = GraphQLHTTPPostBody(query: request.query, variables: request.variables)
+        // TODO: - encoding `Data` results in mumbo jumbo string. Why
         let postData = try JSONEncoder().encode(postBody)
```
sshropshire: interess: what ends up happening to the String?

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -78,12 +79,15 @@ public class APIClient {
...
         let urlString = coreConfig.environment.baseURL.appendingPathComponent(path)
         var urlComponents = URLComponents(url: urlString, resolvingAgainstBaseURL: false)
         
-        queryParameters.forEach {
-            urlComponents?.queryItems?.append(URLQueryItem(name: $0.key, value: $0.value))
+        if let queryParameters {
+            urlComponents?.queryItems = []
+            queryParameters.forEach {
+                urlComponents?.queryItems?.append(URLQueryItem(name: $0.key, value: $0.value))
+            }
```
sshropshire:
```suggestion
        urlComponents?.queryItems = queryParameters?.map { URLQueryItem(name: $0.key, value: $0.value }
```
take or leave: we could use map here to streamline.

Sources/CorePayments/Networking/TrackingEventsAPI.swift
sshropshire: 2 cents: I do like the internal name. It prevents variable shadowing, and technically the signature of the constructor is still `init(coreConfig: CoreConfig)` externally. The internal parameter name won't leak outside of this file.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -40,6 +40,7 @@ public struct AnalyticsService {
     
     // MARK: - Internal Methods
     
+    /// Exposed to be able to execute this function synchronously in unit tests
```
sshropshire, line 43: What's the support like for async await in XCTest?

sshropshire (approved):
🌊

Sources/CorePayments/Networking/APIClient.swift
scannillo: Woops, this is an outdated comment! Removed in a467c992ab1637a07ba0ec7eae5f630d35beea78

Sources/CorePayments/Networking/APIClient.swift
scannillo: Great idea - done in f43ce8218fa5f1e0682fe0a53c31f778526f2919

Sources/CorePayments/Networking/AnalyticsService.swift
scannillo: Support is solid! But in this case `sendEvent()` it isn't an `async` function since the feature clients don't care if it completed or not. Having it be non-async lets us clean up the callsite code, as well. It doesn't return anything and just executes via a background Task. So in order for our unit tests to make assertions after `sendEvent` completes, I opted to expose the internal `async` function, which lets you do `await` on.

Though it's worth looking into to see if there's a way to manipulate `Tasks` (like [here](https://github.com/paypal/paypal-ios/blob/4e18740043971268993b600dea03a3e2e3dd3d01/Sources/CorePayments/Networking/AnalyticsService.swift#L36)) from within our test suite

PR #196, opened 2023-09-11
Use Documentation Annotations over nodoc

KunJeongPark:
### Summary of changes

- added  @_documentation(visibility: private) under nodoc comments to make
  sure public documentation is not generated for classes, functions or properties for internal purposes only.

### Checklist

### Authors
- @KunJeongPark


jaxdesmarais:
We will need to bump the Xcode and Swift version to use this attribute. See Sammy's PR here for those versions: https://github.com/braintree/braintree_ios/pull/1034

Sources/CardPayments/Models/Card.swift
```diff
@@ -23,6 +23,7 @@ public struct Card: Encodable {
 
     //// :nodoc: This is exposed for convenience in copying card contents to UpdateSetupTokenQuery
     /// The expiration year and month, in ISO-8601 `YYYY-MM` date format.
+    @_documentation(visibility: private)
     public var expiry: String {
```
sshropshire, line 27: q: Can we remove this property instead? It shouldn't be difficult to move the `"\(expirationYear)-\(expirationMonth)"` string to the demo app.

sshropshire (approved):

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -11,5 +11,6 @@ public struct CardResult {
 
     //// :nodoc: This is the deep link url returned from 3DS authentication
```
scannillo:
```suggestion
    //// This is the deep link url returned from 3DS authentication
```

We can remove the `nodoc` word from the docstrings. These never actually did anything in this SDK since it isn't set up for Jazzy.

Sources/CardPayments/Models/Card.swift
KunJeongPark: I prefer this since it is a good compromise of less error for us in future in demo app and hiding implementation from public interface.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -28,7 +29,7 @@ public struct AnalyticsService {
     
     // MARK: - Public Methods
         
-    /// :nodoc: This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+    /// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
```
jaxdesmarais, line 32: I think we want a `_documentation` annotation below

KunJeongPark: Oh, I thought you only need it at struct level.

jaxdesmarais: Ah, you may be right. You can check in Xcode by going to `Product` > `Build Documentation` and verifying this is hidden as expected.

KunJeongPark: Nice. No, I don't see it but is Demo app stuff supposed to be in there?

jaxdesmarais (approved):

Sources/CardPayments/Models/Card.swift
scannillo: Looks like this is [not public on Android](https://github.com/paypal/paypal-android/blob/main/CardPayments/src/main/java/com/paypal/android/cardpayments/Card.kt), and we should align our public APIs as much as possible. I agree that merchants should not need access to this.

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -9,7 +9,8 @@ public struct CardResult {
     /// The order ID associated with the transaction
     public let orderID: String
 
-    //// :nodoc: This is the deep link url returned from 3DS authentication
     // TODO: parse contents of this URL once we are clear on values returned from 3ds
+    /// This is the deep link url returned from 3DS authentication
```
scannillo, line 13: The above TODO shouldn't have been removed

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -27,8 +28,10 @@ public class APIClient {
     }
     
     // MARK: - Public Methods
-    
-    /// :nodoc:
+
+    /// This function makes a urlRequest from RESTRequest and returns HTTPResponse
```
scannillo: Working nit

```suggestion
    /// This function executes a network request from a RESTRequest and returns HTTPResponse
```

Sources/CorePayments/Networking/APIClient.swift
```diff
@@ -52,8 +55,10 @@ public class APIClient {
         
         return try await http.performRequest(httpRequest)
     }
-    
-    /// :nodoc:
+
+    /// This function makes a urlRequest from GraphQLRequest and returns HTTPResponse
```
scannillo:
```suggestion
    /// This function executes a network request from a GraphQLRequest and returns HTTPResponse
```

Sources/CardPayments/Models/Card.swift
KunJeongPark: We don't actually use this field in Demo app, we have a different `BasicCard` struct. I just have to modify `UpdateSetupToken` encoding.

scannillo (approved):

Sources/CardPayments/Models/Card.swift
KunJeongPark: I'll put this in separate PR

PR #203, opened 2023-10-03
[Docs] PayPal Buttons: Fix Typos

sshropshire:
### Reason for changes

Docs code for `PaymentButtons` does not compile when copy pasted into an Xcode project.

### Summary of changes

- Add missing `import` statements for `PaymentButtons` onboarding
- Fix typo in PayPal button target selector name 

### Checklist

- @sshropshire

### DIFF
--- a/docs/PaymentButtons/README.md
+++ b/docs/PaymentButtons/README.md
@@ -36,6 +36,8 @@ Each button as a `UKit` and `SwiftUI` implementation as follows:
 #### UKit
 
 ```swift
+import PaymentButtons
+
 class MyViewController: ViewController {
 
     lazy var payPalButton: PayPalButton = {
@@ -44,7 +46,7 @@ class MyViewController: ViewController {
         return payPalButton
     }()
     
-    @objc func paymentButtonTapped() {
+    @objc func payPalButtonTapped() {
         // Insert your code here
     }
     
@@ -58,6 +60,8 @@ class MyViewController: ViewController {
 #### SwiftUI
 
 ```swift
+import PaymentButtons
+
 struct MyApp: View {
     @ViewBuilder
     var body: some View {
--- a/docs/PaymentButtons/README.md
+++ b/docs/PaymentButtons/README.md
@@ -59,7 +59,6 @@ class MyViewController: ViewController {
         NSLayoutConstraint.activate([
             view.leadingAnchor.constraint(equalTo: safeArea.leadingAnchor, constant: 20)
             view.trailingAnchor.constraint(equalTo: safeArea.trailingAnchor, constant: -20)
-            view.centerXAnchor.constraint(equalTo: safeArea.centerXAnchor)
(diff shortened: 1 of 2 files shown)


docs/PaymentButtons/README.md
```diff
@@ -44,7 +46,7 @@ class MyViewController: ViewController {
         return payPalButton
```
sshropshire, line 46: ❓ I did struggle with adding view constraints here because it requires knowledge of a merchant app's view layout. Should we assume the merchant has an empty screen at this point and center the button vertically?

KunJeongPark: I think the developers know to take care of the constraints based on their layout. I don't think they expect us to provide constraints in our example. I'll take a closer look and see how this is used in demo app maybe we can give example of how.

sshropshire: I'm cool with leaving them out for now. It did make me think though, maybe a [GitHub gist](https://docs.github.com/en/get-started/writing-on-github/editing-and-sharing-content-with-gists/creating-gists) would be a good tool to use in the future for onboarding merchants.

KunJeongPark: Interesting.

KunJeongPark: Probably, they'd want to know how the color, edge, text customizations work but that should be another item.

KunJeongPark (approved):

docs/PaymentButtons/README.md
jaxdesmarais: We could add constraints to just center the button on a blank page, don't have to do anything wild just give a basic example so merchants know they have to add them.

sshropshire: I'll add some constraints in.

scannillo (approved):

docs/PaymentButtons/README.md
```diff
@@ -44,20 +46,31 @@ class MyViewController: ViewController {
...
         super.viewDidLoad()
         view.addSubview(payPalButton)
+
+        // example constraints to absolutely center the button with a 20 pt. horizontal margin
+        let safeArea = view.safeAreaLayoutGuide
+        NSLayoutConstraint.activate([
+            view.leadingAnchor.constraint(equalTo: safeArea.leadingAnchor, constant: 20)
+            view.trailingAnchor.constraint(equalTo: safeArea.trailingAnchor, constant: -20)
+            view.centerXAnchor.constraint(equalTo: safeArea.centerXAnchor)
+            view.centerYAnchor.constraint(equalTo: safeArea.centerYAnchor)
```
jaxdesmarais: nit: we shouldn't need a center X constraint if we have a leading and trailing constraint. Also, we may need a height constraint unless our SDK handles height of the button?

sshropshire: Oh right yeah good call. I can remove the center X constraint. The base class `UIButton` should have an intrinsic height based on the font size, so we should be good there.

PR #204, opened 2023-10-04
Remove PayPalCheckout dependency from Demo

scannillo:
### Reason for changes

The PPCP SDK pulls in NXO directly so there is no need for a merchant app to be directly fetching the NXO package.

### Summary of changes

- Remove `PayPalCheckout` from the Demo app's Package Dependency list

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/PayPal.xcodeproj/project.pbxproj
+++ b/PayPal.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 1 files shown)


scannillo:
Dang, our Demo app still needs refactoring to remove usage of `PayPalCheckout`. We have a separate ticket for this work - DTPPCPSDK-501.

sshropshire:
> Dang, our Demo app still needs refactoring to remove usage of `PayPalCheckout`. We have a separate ticket for this work - DTPPCPSDK-501.

Yoo that might fix the Archive bug. But on second thought I guess we'd have to remove the SPM build flags as well. Not sure if Xcode does that automatically.

scannillo:
> @sshropshire Yoo that might fix the Archive bug. But on second thought I guess we'd have to remove the SPM build flags as well. Not sure if Xcode does that automatically.

Exactly yeh, that's why I thought we should do this first. But then it ended up revealing that our Demo app is using PayPalCheckout components for the model layer, which we need to refactor and then rip out the NXO dependency in the Demo. 

New Xcode projects that don't include any SPM packages don't include the macro flag. I was hoping that removing the package would remove the macro, but it didn't. We should be safe to remove it manually though!

PR #205, opened 2023-10-10
Rename APIClient to NetworkingClient

scannillo:
### Reason for changes

This change came out of our series of [networking layer refactor PRs](https://github.com/paypal/paypal-ios/pulls?q=is%3Apr+networking+label%3Anetworking-refactor) which introduces the `<PPaaS>API.swift` files. 

The `APIClient` has a different responsibility now, to transform GraphQL & REST requests into the HTTP protocol + execute the request.

### Feedback

Do folks think this rename is helpful? Do folks have a better name idea for this class? I think we tend to overuse the term `Client`.

### Summary of changes

- Rename `APIClient` to `NetworkingClient`

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo

### DIFF
--- a/Sources/CorePayments/Networking/NetworkingClient.swift
+++ b/Sources/CorePayments/Networking/NetworkingClient.swift
@@ -1,11 +1,10 @@
 import Foundation
 
-// TODO: - Rename to `NetworkingClient`. Now that we have `<PPaaS>API.swift` classes, ths responsibility of this class really is to coordinate networking. It transforms REST & GraphQL into HTTP requests.
 /// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 ///
-/// `APIClient` is the entry point for each payment method feature to perform API requests. It also offers convenience methods for API requests used across multiple payment methods / modules.
+/// `NetworkingClient` transforms REST & GraphQL requests into the proper HTTP format and executes network requests.
 @_documentation(visibility: private)
-public class APIClient {
+public class NetworkingClient {
         
     // MARK: - Internal Properties
     
--- a/PayPal.xcodeproj/project.pbxproj
+++ b/PayPal.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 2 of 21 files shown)


sshropshire (approved):

KunJeongPark (approved):

Sources/CorePayments/Networking/NetworkingClient.swift
```diff
@@ -1,11 +1,10 @@
 import Foundation
 
-// TODO: - Rename to `NetworkingClient`. Now that we have `<PPaaS>API.swift` classes, ths responsibility of this class really is to coordinate networking. It transforms REST & GraphQL into HTTP requests.
 /// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 ///
-/// `APIClient` is the entry point for each payment method feature to perform API requests. It also offers convenience methods for API requests used across multiple payment methods / modules.
+/// `NetworkingClient` transforms REST & GraphQL requests into the proper HTTP format and executes network requests.
 @_documentation(visibility: private)
-public class APIClient {
+public class NetworkingClient {
```
jaxdesmarais, line 7: 🤔 on other name options. Not sure if `NetworkingOperation/Operator`, `NetworkConfiguration`, `NetworkConstructor`, `NetworkBuilder` makes more sense?

scannillo: 🤔 Along those lines, maybe just `Networking`? (... `Netwerker` 😆).

Also if we don't feel strongly, we could leave `NetworkingClient` too

scannillo: The methods aren't static, but here's how the definitions would look roughly:
```
Network.fetch(restRequest:_)
NetworkingClient.fetch(restRequest:_)
NetworkBuilder.fetch(restRequest:_)
NetworkOperator.fetch(restRequest:_)
```

jaxdesmarais: I think `NetworkOperator ` or `Network/Networking` look nice. But I also don't feel strongly enough about it to not go with `NetworkingClient` if we want. What do you think?

scannillo: Yeh, I agree about not feeling that strongly at the moment. Let's go with `NetworkingClient` and we could always change it again later on if the feeling strikes!

PR #206, opened 2023-10-10
Remove fastlane

scannillo:
### Reason for changes

We [added fastlane ](https://github.com/paypal/paypal-ios/pull/4)when this repo was first created. The main reasons were:
* To be able to run test commands easily on the command line
* For xcodecov reports in PRs

I personally haven't used Fastlane for either of those use cases. I always just run tests in Xcode as well as use Xcode to generate code coverage reports. For that reason, I think having Fastlane adds a layer of complexity to our setup.

**_Feedback_**: I would love to know what other folks experience is with Fastlane in this repo.

### Summary of changes

This PR removes Fastlane and instead relies on SwiftLint, where needed, directly
- Removes `Gemfile`
- Removes `Fastfile` & `xcovignore` files
- Updates CI to pull in SwiftLint directly & run `xcodebuild` command directly for tests

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


jaxdesmarais (approved):
I've never used Fastlane either and do the same as you (run tests in Xcode). I don't see it as necessary to our flow so agree with killing it.

sshropshire (approved):
🌊

PR #207, opened 2023-10-10
Prepare PPCP Demo App for TestFlight

sshropshire:
### Reason for changes

The `SWIFT_PACKAGE` compiler macro prevents our Xcode project from being archived. We did some investigation, and it seems like we can remove this compiler directive. SPM defines this compiler macro on it's own when building SPM framework binaries via Xcode.

We've also removed references to PayPalCheckout from the Demo app in an effort to gradually phase it out.

### Summary of changes

- Remove `PayPalCheckout` module from the Demo app's declared SPM dependencies.
- Remove `SWIFT_PACKAGE` compiler macro
- Update Bundle Identifier to Team SDK managed `com.paypal.ios.Demo`
- Add PPCP Demo app icon to project

### Checklist

- @sshropshire
- Thanks to @scannillo and @KunJeongPark for pairing on this!


scannillo:
Can we also remove the uses if MXO in the Demo app (for example [here](https://github.com/paypal/paypal-ios/blob/035f434e347652584dc2cbfefb73a7c3294efef0/Demo/Demo/Helpers/OrderRequestHelpers.swift#L2))? Or is that work planned for a separate PR?

PayPal.xcodeproj/project.pbxproj
```diff
@@ -1365,7 +1364,6 @@
 				80E2FDBE2A83528B0045593D /* CheckoutOrdersAPI.swift in Sources */,
 				3B79E4F72A8503CA00C01D06 /* UpdateVaultVariables.swift in Sources */,
 				8048D28C270B9DE00072214A /* ConfirmPaymentSourceResponse.swift in Sources */,
-				3B109B3D2A85D1CA00D8135F /* MockGraphQLClient.swift in Sources */,
 				3D1763A22720722A00652E1C /* CardResult.swift in Sources */,
```
KunJeongPark, line 1367: Hey Steven, why are these diffs showing here?

scannillo: That showed up in [this PR](https://github.com/paypal/paypal-ios/pull/205/files) too. It must have been forgotten in the original PR to remove that file, and Xcode is catching that unnecessary line now (could be also due to Xcode 15 bump).

I think it's fine to let slip in one of these PRs

KunJeongPark (approved):

Demo/Demo/Info.plist
```diff
@@ -2,8 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>CFBundleShortVersionString</key>
-	<string>1.0.0</string>
 	<key>UIApplicationSceneManifest</key>
```
KunJeongPark, line 5: Why is this deleted? I don't think it matters for TestFlight but curious

sshropshire: Yeah Xcode on my machine has been in a legit "hold my beer" mood. I'm not entirely sure why it's making some of these decisions.

scannillo: I know Xcode can handle automatically incrementing the version # based on what exists in TestFlight, which is nice since with every upload we make we don't have to make any source code changes.

I think if Xcode deleted it, we can assume its fine

scannillo:
I'm able to run this on a simulator, but running on a device results in a crash:

```
dyld[10171]: Library not loaded: @rpath/PayPalCheckout.framework/PayPalCheckout
  Referenced from: <A809CAFE-5DAD-3AE2-858D-28A50905DC35> /private/var/containers/Bundle/Application/F8555334-4348-44FB-852A-74E600F1952E/Demo.app/Demo
  Reason: tried: '/private/var/containers/Bundle/Application/F8555334-4348-44FB-852A-74E600F1952E/Demo.app/Frameworks/PayPalCheckout.framework/PayPalCheckout' (errno=2)
```

We should try removing references to PayPalCheckout in the demo to see if that resolves this.

jaxdesmarais:
@scannillo I just saw a similar crash earlier today on BT when running on device. I think we are pulling it in as a package since we are using a direct integration (vs a package manager). We should decide if we want to:
1. Continue to not use a package manager in our demo app
2. If so - how do we want to pull in MXO since it needs a reference to it to build on device

scannillo:
> @scannillo I just saw a similar crash earlier today on BT when running on device. I think we are pulling it in as a package since we are using a direct integration (vs a package manager). We should decide if we want to:

> Continue to not use a package manager in our demo app
If so - how do we want to pull in MXO since it needs a reference to it to build on device

We were talking about this in a huddle & slack thread the other day (I'll CC you in it) about moving to a structure like iOS Drop-In to pull PPCP into the Demo. The pro of it is that it mirrors a true SPM integration and uses the Package.swift to define the SDK. The con of it is that it's confusing to track down where the reference to the _local_ version of the SDK is set ([here](https://github.com/braintree/braintree-ios-drop-in/blob/2402049491d4654fa2337894c46ee082f5157918/Demo.xcodeproj/project.pbxproj#L61)). I do want to try it out, though!

jaxdesmarais:
> We were talking about this in a huddle & slack thread the other day (I'll CC you in it) about moving to a structure like iOS Drop-In to pull PPCP into the Demo. The pro of it is that it mirrors a true SPM integration and uses the Package.swift to define the SDK. The con of it is that it's confusing to track down where the reference to the _local_ version of the SDK is set ([here](https://github.com/braintree/braintree-ios-drop-in/blob/2402049491d4654fa2337894c46ee082f5157918/Demo.xcodeproj/project.pbxproj#L61)). I do want to try it out, though!

To me this makes the most sense. Especially since we try to discourage direct integrations then use it ourselves. 🙈 I'd be down to try it out on BT and/or PPCP!

scannillo (approved):

PR #208, opened 2023-10-10
FPTI Updates

jaxdesmarais:
### Reason for changes

Update component name to match Braintree formatting ([BT PR with this change](https://github.com/braintree/braintree_ios/pull/1104)) as will as send correlation ID where possible ([BT PR with this change](https://github.com/braintree/braintree_ios/pull/1108))

**Feedback:** currently the correlation ID was only added to the MXO flow. On BT PayPal Web we pull this value from the request (if present) or the Data Collector ([source code](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreePayPal/BTPayPalClient.swift#L267)). PPCP PayPal Web does not have the Data Collector as a dependency - we should consider if we want to add this dependency in a future PR or leave the correlation ID as is - only existing in NXO.

**Note:** We will make this same change on Android as well.

### Summary of changes

- Update `component` from `ppcpmobilesdk` to `ppcpclientsdk`
- Update `AnalyticsService` to accept an optional `correlationID` parameter
- Collect `correlationID` & send in analytics for `PayPalNativeCheckoutClient` when present

### Checklist

- [x] Added a changelog entry

### Authors

- @jwarmkessel @jaxdesmarais


Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -33,24 +33,25 @@ public struct AnalyticsService {
...
-            await performEventRequest(name)
+            await performEventRequest(name, correlationID: correlationID)
         }
     }
     
     // MARK: - Internal Methods
     
     /// Exposed to be able to execute this function synchronously in unit tests
-    func performEventRequest(_ name: String) async {
+    func performEventRequest(_ name: String, correlationID: String? = nil) async {
```
scannillo, line 48: Nit - update docstring for this new param

scannillo:
The API requests in the CardModule also return a riskCorrelationID as a response header ([see Android](https://github.com/paypal/paypal-android/blob/c4fd610e1e54881602ff5a98ea3e2c8d92807891/CardPayments/src/main/java/com/paypal/android/cardpayments/CardResponseParser.kt#L35)).

Should we also log the correlationID in the card module where possible? 

It also looks like we're mis-aligned with Android since they return riskCorrelationID in merchant-facing errors and iOS is not (slightly separate concern though).

Sources/CorePayments/Networking/AnalyticsService.swift
jaxdesmarais: Good call - done: [PR feedback - add docstrings for correlationID](https://github.com/paypal/paypal-ios/pull/208/commits/bcc835e12572d05e9987f4386d64ab229c1f64a8)

jaxdesmarais:
> The API requests in the CardModule also return a riskCorrelationID as a response header ([see Android](https://github.com/paypal/paypal-android/blob/c4fd610e1e54881602ff5a98ea3e2c8d92807891/CardPayments/src/main/java/com/paypal/android/cardpayments/CardResponseParser.kt#L35)).

Adding breakpoints in the request/response I am not seeing those headers. Additionally looking at `confirm-payment-source` in PPaaS [here](https://ppaas/api/3719176155273543#apiReference). It looks like the correlation ID is returned as the debug ID from this endpoint in the response body. I assume that header came from somewhere on the Android side, but not seeing it when triggering an error in the raw response.

> Should we also log the correlationID in the card module where possible?

The networking layers and confirm payment source request/response parsing on Android and iOS are pretty divergent right now. I put up [this commit](https://github.com/paypal/paypal-ios/pull/208/commits/192e5fbb3f4ce0149d490c990e732e9b00d8b3a3) which extracts the correlation/debug ID from the `httpResponse` to allow us to pass it in with errors. Curious to hear your thoughts - it's not the cleanest solution but works well without fully redoing how we make these requests/responses on iOS for cards.

> It also looks like we're mis-aligned with Android since they return riskCorrelationID in merchant-facing errors and iOS is not (slightly separate concern though).

I think we can probably punt on this for now. We should align card as a whole more closely across the platforms as they are pretty different right now.

Sources/CorePayments/AnalyticsEventData.swift
```diff
@@ -41,7 +42,9 @@ struct AnalyticsEventData: Encodable {
 
     let clientOS: String = UIDevice.current.systemName + " " + UIDevice.current.systemVersion
 
-    let component = "ppcpmobilesdk"
+    let component = "ppcpclientsdk"
```
scannillo, line 45: 😈 advocate - will changing the component name cause more trouble than it's worth?

The pro is to mirror BT analytics component name, but the con is definitely a pain of needing to manage data across components.

jaxdesmarais: That's a great question! I think if we plan to document this and use it for future SDKs, then yes. It's also probably easier to make the change now vs once we have a ton more data coming through the SDK. It also allows us to be consistent across products vs doing `mobile` in one and `client` in another, which I think has been the goal with moving to FPTI - consistency.

sshropshire (approved):

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
```diff
@@ -39,6 +43,9 @@ class CheckoutOrdersAPI {
         )
         
         let httpResponse = try await apiClient.fetch(request: restRequest)
+        let httpResponseBodyJSON = try? JSONSerialization.jsonObject(with: httpResponse.body ?? Data()) as? [String: Any]
+        correlationID = httpResponseBodyJSON?["debug_id"] as? String
+
         return try HTTPResponseParser().parseREST(httpResponse, as: ConfirmPaymentSourceResponse.self)
```
scannillo: I think we should 🏈 do this work in a separate PR/ workstream (for parsing out the riskCorrelationID from the body or response headers). If it's in the body, this should be included in the call on line 49, and added to `ConfirmPaymentSourceResponse` model.

Sources/CorePayments/AnalyticsEventData.swift
scannillo: > It also allows us to be consistent across products vs doing mobile in one and client in another, which I think has been the goal with moving to FPTI - consistency.

I agree with this, but also we don't know the analytics story for PPCP server or PPCP web. Since PPCP web is baked into smart payment buttons, there is low chance they'll do a rename to align with us. I'll try to find out more about their analytics setup.

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
jaxdesmarais: > If it's in the body, this should be included in the call on like 49, and added to ConfirmPaymentSourceResponse model.

Yeah - that's why I mentioned the entire layer would likely need to be rewritten. It's _only_ returned when there is an error, so we don't actually return a `ConfirmPaymentSourceResponse` when a `debug_id` is present since the SDK throws an error as part of the try at that point. Adding it to the response would mean it's always `nil` since we only get a `ConfirmPaymentSourceResponse` on success (and there is no `debug_id` on success according to PPaaS)

Sources/CorePayments/AnalyticsEventData.swift
jaxdesmarais: Agree it's unlikely that PPCP web/SPB uses anything close to this. We likely do still have an opportunity to align with the server SDKs since those are still being written.

scannillo: I guess what I'm saying is I kind of doubt the PPCP web product will want to change their component name from whatever it is to `ppcpclientsdk` just to match ours, so is there really a point to changing ours?

scannillo: Yeh, the server SDKs won't use `<product>clientsdk`, so mobile might just always be it's own special name

jaxdesmarais: Ah - good point about server. We can revert it. I guess my 👿 advocate argument would be why did we change it on BT iOS. Lol. I don't feel strongly enough to argue in either direction besides it being divergent from what we decided on BT, so if you think we should use `mobile` I can make that change.

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
scannillo: Gotcha, so we only get that value back on errors. 

We do have error specific parsing that uses a different underlying `Decodable` model (triggered [here](https://github.com/paypal/paypal-ios/blob/0fb35166d971f3001eb2aa22f55d4918fef7d8bb/Sources/CorePayments/Networking/HTTPResponseParser.swift#L25) & defined [here](https://github.com/paypal/paypal-ios/blob/main/Sources/CorePayments/Networking/Models/ErrorResponse.swift)).

So we would update that model to parse out `debug_id`.

scannillo: Maybe in another ticket we can dig into when we expect that value to come in the error body, or in the response headers and which to prefer when 🏈

scannillo (approved):
Actual changes look good! Thanks for entertaining all of my ❓s!

Sources/CorePayments/AnalyticsEventData.swift
scannillo: Yeh, I get that. I think changing on BT made more sense since we have old events coming in that use a different name (from Arachne), so now they're consistent.

I could go either way too!

Sources/CardPayments/APIRequests/CheckoutOrdersAPI.swift
jaxdesmarais: Yeah - If we add it to those thrown errors ([from here](https://github.com/paypal/paypal-ios/blob/0fb35166d971f3001eb2aa22f55d4918fef7d8bb/Sources/CorePayments/Networking/HTTPResponseParser.swift#L58-L70)), we'd end up needing to parse the error again on the client to extract out the debug ID from the thrown error. It could make more sense to return a result object or closure as part of `parseREST` as well as update `CheckoutOrdersAPI` to do the same. Or we can take the same approach as web and just return it with all errors.

It's certainly a bit of an odd series of events and would be a good opportunity to align with Android on! Looking between the two codebases for this certainly illustrated how different this layer is in particular. I'll revert this for now in any case!

jaxdesmarais: Reverted: https://github.com/paypal/paypal-ios/pull/208/commits/b031319cbdf8d2133e82a5c9652d7ccd152b68f7

PR #209, opened 2023-10-12
Remove uses of `PayPalCheckout` in Demo code

scannillo:
### Reason for changes

Our Demo app should accurately reflect a merchant integration. Our merchants will not be using the `PayPalCheckout` SDK directly.

### Summary of changes

- Replace use of MXO model types for create order ID `POST` request
- Replace use of MXO model types for update order ID `PATCH` request
- Remove SPM `PayPalCheckout` package dependency in Demo project

I tested the NXO demo & patching works as expecting

### Checklist

- ~Added a changelog entry~

### Authors
@scannillo


Demo/Demo/ViewModels/PayPalViewModel.swift
```diff
@@ -39,18 +38,18 @@ class PayPalViewModel: ObservableObject {
...
 
     func checkoutWithProvidedAddress() {
-        checkout(.setProvidedAddress)
+        checkout(ShippingPreference.providedAddress)
     }
 
     func checkoutWithGetFromFile() {
-        checkout(.getFromFile)
+        checkout(ShippingPreference.getFromFile)
     }
```
scannillo, line 50: There's likely a simpler more "SwiftUI-y" way to do this selection communication b/w View & ViewModel layers, but that refactor should be a separate PR

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -1,104 +1,97 @@
...
+    
+    static func getShippingMethods(selectedID: String = "ShipTest1") -> [ShippingOption]? {
+        let shipOption1 = ShippingOption(
             selected: false,
-            type: .shipping,
-            amount: UnitAmount(currencyCode: currency, value: String(0.99 + Double(baseValue)))
+            id: "ShipTest1",
+            amount: ItemTotal(value: "3.99", currencyValue: "3.99", currencyCode: currencyCode),
+            label: "Standard Shipping",
+            type: "SHIPPING"
```
sarahkoop, line 51: Should there be an enum in our SDK for this? Not necessary for this PR but looks cleaner with `.shipping` in the checkout SDK

sarahkoop (approved):

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -1,104 +1,97 @@
...
-import PayPalCheckout
 
 enum OrderRequestHelpers {
 
-    static var orderAmount = 100.0
-
-    static func getOrderRequest(_ shippingPreference: OrderApplicationContext.ShippingPreference) -> OrderRequest {
-        return OrderRequest(
-            intent: .authorize,
+    static let orderAmount = 100.0
```
jaxdesmarais, line 5: nit: I don't think the PPCP sample merchant server is set up to auto void like BT yet. Do we want to set the order amount to like $1 incase we get a prod account stood up so we don't accidentally change ourselves $100 the first time we test? 🙈

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: Great question - so this helper function is formatting the POST body for a `PATCH` API call to orders v2 (see public docs https://developer.paypal.com/docs/api/orders/v2/#orders_patch). The merchant is responsible for doing this on their server after the shipping information changes in the MXO flow. Our sample server is just propagating whatever POST body we're sending on the client, through to orders/v2, but I imagine most of this logic would live on a merchant server instead.

That said, we could offer this enum, but to me feels out of the mobile SDK's domain since it is part of a server-side call they have to make.

What do you think?

Demo/Demo/Helpers/OrderRequestHelpers.swift
```diff
@@ -1,104 +1,97 @@
...
+                        address: Shipping.Address(
                             addressLine1: "345 Sesame Street",
                             addressLine2: "Apt 9",
-                            adminArea1: "NY",
                             adminArea2: "New York City",
+                            adminArea1: "NY",
+                            countryCode: "US",
                             postalCode: "32422"
                         ),
+                        name: Shipping.Name(fullName: "Cookie Monster"),
```
jaxdesmarais, line 24: 😍

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: Another option too is an enum internal to the Demo app

jaxdesmarais (approved):

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: The PayPalNative & PayPalWeb demos aren't setup for Auth/Capture, I left a TODO for it.

But definitely agree once we get a prod account we should do that!

Demo/Demo/Helpers/OrderRequestHelpers.swift
scannillo: Just carrying Jose's torch!

PR #210, opened 2023-10-12
Vault with Purchase Direct

KunJeongPark:
### Summary of changes

- Decode payment source for authorize/capture response in demo app
- Remove disable modifier on vault toggle option in Card approval view
- Pass in vault parameters if vault with purchase is selected
- Add 4000 0000 0000 0002 as Verified Vault (no 3DS) on card selection menu
- Adjust views to display payment source and vaultID and customerID

https://github.com/paypal/paypal-ios/assets/9273272/e636bb36-a80b-4811-9b99-a63ce34c980b

### Checklist

### Authors
- @KunJeongPark @sarahkoop


Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderApproveView.swift
```diff
@@ -6,6 +6,7 @@ struct CardOrderApproveView: View {
 
     let cardSections: [CardSection] = [
         CardSection(title: "Successful Authentication Visa", numbers: ["4868 7194 6070 7704"]),
+        CardSection(title: "Vault with Purchase(no 3DS)", numbers: ["4000 0000 0000 0002"]),
```
sshropshire:
```suggestion
        CardSection(title: "Vault with Purchase (no 3DS)", numbers: ["4000 0000 0000 0002"]),
```
🧶 : take or leave.

Demo/Demo/ViewModels/CardPaymentViewModel.swift
```diff
@@ -10,14 +10,22 @@ class CardPaymentViewModel: ObservableObject, CardDelegate {
...
+    ) async throws {
         // might need to pass in payee as payee object or as auth header
 
         let amountRequest = Amount(currencyCode: "USD", value: amount)
         // TODO: might need to pass in payee as payee object or as auth header
         let orderRequestParams = CreateOrderParams(
             intent: intent,
-            purchaseUnits: [PurchaseUnit(amount: amountRequest)]
+            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
+            isVaulting: isVaulting,
```
sshropshire:
```suggestion
            shouldVault: shouldVault,
```
🧶 : personal style, i've always liked "should" when paired with vault. I feel like it bucks strict conventions, but it reads well and serves the same purpose as "is" in regards to the "Vault" feature.

sshropshire (approved):
🙌

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -2,6 +2,21 @@ struct CreateOrderParams: Codable {
...
+        if isVaulting {
+            var customer: CardVaultCustomer?
+            if let customerID {
+                customer = CardVaultCustomer(id: customerID)
+            }
+            let attributes = Attributes(vault: Vault(storeInVault: "ON_SUCCESS"), customer: customer)
+            let card = VaultCard(attributes: attributes)
+            self.paymentSource = VaultCardPaymentSource(card: card)
+        }
+    }
```
jaxdesmarais: Should we move this logic out of the `init`? It seems like we could do this check at a later point or in the view directly vs the init of the model.

Demo/Demo/Models/Order.swift
```diff
@@ -1,6 +1,6 @@
 struct Order: Codable, Equatable {
 
-    enum CodingKeys: String, CodingKey {
+    enum OrderCodingKeys: String, CodingKey {
```
jaxdesmarais: It looks like everywhere else in the SDK we go with the default `CodingKeys`: https://github.com/search?q=repo%3Apaypal%2Fpaypal-ios%20CodingKeys&type=code. Is there a reason we want to diverge from that here?

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderApproveView.swift
jaxdesmarais: ➕ on adding the space!

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderCompletionResultView.swift
```diff
@@ -35,6 +35,24 @@ struct CardOrderCompletionResultView: View {
             LeadingText("\(orderResponse.id)")
             LeadingText("Status", weight: .bold)
             LeadingText("\(orderResponse.status)")
+            if let lastDigits = orderResponse.paymentSource?.card.lastDigits {
```
jaxdesmarais, line 38: Since this seems to be the same logic below can we can maybe extract this out into a method where we pass in the `orderResponse.paymentSource`?

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardPaymentOrderCompletionView.swift
```diff
@@ -6,21 +6,32 @@ struct CardPaymentOrderCompletionView: View {
...
+                            intent: cardPaymentViewModel.state.intent,
+                            orderID: orderID,
+                            selectedMerchantIntegration: DemoSettings.merchantIntegration,
+                            cardPaymentViewModel: cardPaymentViewModel
+                        )
+                    }
 
-            if cardPaymentViewModel.state.authorizedOrder != nil || cardPaymentViewModel.state.capturedOrder != nil {
-                CardOrderCompletionResultView(cardPaymentViewModel: cardPaymentViewModel)
+                    if cardPaymentViewModel.state.authorizedOrder != nil || cardPaymentViewModel.state.capturedOrder != nil {
```
jaxdesmarais: It looks like we use `cardPaymentViewModel.state` across this view quite a bit. Should we consider extracting it out into a constant?

Demo/Demo/SwiftUIComponents/CardPaymentViews/CreateOrderCardPaymentView.swift
```diff
@@ -47,7 +45,9 @@ struct CreateOrderCardPaymentView: View {
                             try await cardPaymentViewModel.createOrder(
                                 amount: "10.00",
                                 selectedMerchantIntegration: DemoSettings.merchantIntegration,
-                                intent: selectedIntent.rawValue
+                                intent: selectedIntent.rawValue,
+                                isVaulting: shouldVaultSelected,
```
jaxdesmarais:
```suggestion
                                isVaulting: isVaultSelected,
```

Demo/Demo/Models/CreateOrderParams.swift
vpark777: Not sure what you mean Jax. The vaulting options are passed in the request body, so I just passed the params into here.

KunJeongPark: Not sure what you mean Jax. We pass in vault params in the request body so I am passing in the vault options into param that we use for order creation in demo app.

Demo/Demo/Models/Order.swift
KunJeongPark: I just realized that we don't need any of these codingKeys. Just use 
 ```
let decoder = JSONDecoder()
decoder.keyDecodingStrategy = .convertFromSnakeCase
```
in parse function in` DemoMerchantAPI.` 
We already have keyEncodingStrategy in buildURLRequest function but not in parse function.

File: Demo/Demo/Models/CreateOrderParams.swift (hunk 3 above)
Comment (jaxdesmarais) [reply Demo/Demo/Models/CreateOrderParams.swift]: Typically we want our models to be "dumb" - if you look at the rest of the models in this directory, the `init` is typically super simple and logic/mutations are handled elsewhere. We should be able to do something like this...

EDITED - updated with cleaner code for this here: https://github.com/paypal/paypal-ios/pull/210#discussion_r1358669507

In `CreateOrderParams`:
```swift

    let intent: String
    let isVaulting: Bool
    var purchaseUnits: [PurchaseUnit]?
    var paymentSource: VaultCardPaymentSource?

    init(intent: String, purchaseUnits: [PurchaseUnit]?, isVaulting: Bool = false, customerID: String? = nil) {
        self.intent = intent
        self.purchaseUnits = purchaseUnits
        self.isVaulting = isVaulting
    }
}
```

In `CardPaymentViewModel`:
```swift
    func createOrder(
        amount: String,
        selectedMerchantIntegration: MerchantIntegration,
        intent: String,
        isVaulting: Bool,
        customerID: String? = nil
    ) async throws {
...
        var orderRequestParams = CreateOrderParams(
            intent: intent,
            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
            isVaulting: isVaulting,
            customerID: customerID
        )

        if isVaulting {
            var customer: CardVaultCustomer?
            if let customerID {
                customer = CardVaultCustomer(id: customerID)
            }
            let attributes = Attributes(vault: Vault(storeInVault: "ON_SUCCESS"), customer: customer)
            let card = VaultCard(attributes: attributes)
            orderRequestParams.paymentSource = VaultCardPaymentSource(card: card)
        }

        do {
...
```
^ this is just pseudo code so worth adding some breakpoints to test it's working as expected.

Demo/Demo/Models/Order.swift
KunJeongPark: Because I get warning about nested logic if I have CodingKeys for Order and for Card inside the respective structs. 
I am just putting decodingStrategy in parse method in DemoMerchantAPI so that there is no need for
CodingKeys for snakeCase. I realized that the CodingKey that I had defined for paymentMethodToken 
causes decoding error if I use this strategy on top of that, so I am removing the CodingKey there as well.

Demo/Demo/SwiftUIComponents/CardPaymentViews/CreateOrderCardPaymentView.swift
KunJeongPark: I wanted to distinguish the isVaulting passed to the viewModel vs the selection in the view.

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderApproveView.swift
KunJeongPark: Addressed in e2f1c5967aa90e94f758493b70ef36e2732e78b3

Demo/Demo/ViewModels/CardPaymentViewModel.swift
KunJeongPark: Addressed in e2f1c5967aa90e94f758493b70ef36e2732e78b3

Demo/Demo/Models/Order.swift
KunJeongPark: CodingKeys taken out, replaced with decoding strategy in `parse` method of `DemoMerchantAPI` in commit e2f1c5967aa90e94f758493b70ef36e2732e78b3

Demo/Demo/Models/CreateOrderParams.swift
KunJeongPark: Thank you. I will take a look.

KunJeongPark: Oh I see, you mean,  attach the paymentSource outside of the CreateOrderParams struct, in the viewModel's createOrder function. I just thought it was neater to handle it in creating the params, since it just gets passed to createOrder in DemoMerchantAPI to create the body.

KunJeongPark: We could potentially have other options if requirements come in for these parameters. So should the logic be 
in the CreateOrderParams? or out in createOrder function? It could be header or it could be included in the purchase Units.

jaxdesmarais: I think it's less about creating something outside of the struct and more about separation of concerns. The model `CreateOrderParams ` shouldn't need to know about unrelated types, it should only know about itself. I think we could even take it a step further (same as before - you'll want to confirm everything looks as expected) by doing something like this:
```swift

    let intent: String
    let purchaseUnits: [PurchaseUnit]?
    let paymentSource: VaultCardPaymentSource?
    let shouldVault: Bool

    init(
        intent: String,
        purchaseUnits: [PurchaseUnit]?,
        paymentSource: VaultCardPaymentSource? = nil,
        shouldVault: Bool = false
    ) {
        self.intent = intent
        self.purchaseUnits = purchaseUnits
        self.paymentSource = paymentSource
        self.shouldVault = shouldVault
    }
}
```

```swift
class CardPaymentViewModel: ObservableObject, CardDelegate {

    ...

    func createOrder(
        amount: String,
        selectedMerchantIntegration: MerchantIntegration,
        intent: String,
        shouldVault: Bool,
        customerID: String? = nil
    ) async throws {
        // might need to pass in payee as payee object or as auth header

        var customer: CardVaultCustomer?
        if shouldVault, let customerID {
            customer = CardVaultCustomer(id: customerID)
        }

        let amountRequest = Amount(currencyCode: "USD", value: amount)
        let attributes = Attributes(vault: Vault(storeInVault: "ON_SUCCESS"), customer: customer)
        let card = VaultCard(attributes: attributes)

        // TODO: might need to pass in payee as payee object or as auth header
        let orderRequestParams = CreateOrderParams(
            intent: intent,
            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
            paymentSource: VaultCardPaymentSource(card: card),
            shouldVault: shouldVault
        )
    ...
}
```
Curious to hear your thoughts! 😄

KunJeongPark: Addressed in da203cbe1bbbd9c4cf91b7153fcee4b3e82b2547

KunJeongPark: I like it! Will use it except I don't think we need shouldVault field for CreateOrderRequestParam

KunJeongPark: Addressed in a3a138e88d058c26a3b39aa711fd244b15569114

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardPaymentOrderCompletionView.swift
KunJeongPark: Addressed here: 8a9996c0037e4b0c6086aff0de882bfb42881fe1

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderCompletionResultView.swift
KunJeongPark: Not sure what you mean? Do you mean checking for values in paymentSource?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -167,7 +167,9 @@ final class DemoMerchantAPI {
 
     private func parse<T: Decodable>(from data: Data) throws -> T {
         do {
-            return try JSONDecoder().decode(T.self, from: data)
+            let decoder = JSONDecoder()
+            decoder.keyDecodingStrategy = .convertFromSnakeCase
```
scannillo, line 152: This could break other API calls that our demo makes, can we check to make sure this isn't the case?

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderCompletionResultView.swift
scannillo: Lines 38-55 and 77-92 are doing the same thing. This is usually a good sign that we can refactor the code to be more DRY and re-usable

Demo/Demo/ViewModels/CardPaymentViewModel.swift
```diff
@@ -10,14 +10,33 @@ class CardPaymentViewModel: ObservableObject, CardDelegate {
...
+            let card = VaultCard(attributes: attributes)
+            vaultPaymentSource = VaultCardPaymentSource(card: card)
+        }
+
         let orderRequestParams = CreateOrderParams(
             intent: intent,
-            purchaseUnits: [PurchaseUnit(amount: amountRequest)]
+            purchaseUnits: [PurchaseUnit(amount: amountRequest)],
+            paymentSource: vaultPaymentSource
         )
```
scannillo: This goes against the design of our SDK since in our other payment method flows we never expect the merchant to use the card details in an API call that they make on their own. Only the PPCP SDK is handling any API call that involves card details.

scannillo (changes requested):

Demo/Demo/Networking/DemoMerchantAPI.swift
KunJeongPark: I checked, it broke the PaymentToken response, so I made changes there. I tested all other payment methods and it is fine now. I searched for everything that uses this function in demo app

scannillo: Awesome, thanks for checking!

Demo/Demo/ViewModels/CardPaymentViewModel.swift
KunJeongPark: But there are no card details in here Sammy. Just vault option passed in. Card details are passed in in confirmPaymentSource. PaymentSource only contains vault store_in_vault option and optional customerID if merchant decides to enter it.

scannillo: Hm gotcha. Another question - why does it matter if the merchant send specific params in their CreateOrder call for the vaulting flow, if our `vault` method doesn't accept an orderID at all? (See [`CardVaultRequest` init](https://github.com/paypal/paypal-ios/blob/8b04e5fda77152afa860ea897448479e192ff256/Sources/CardPayments/Models/CardVaultRequest.swift#L17))

scannillo: In this PR it looks like we're using `CardClient.approveOrder()` for vaulting, where I thought our original design/plan was to isolate the vaulting functionality to the separate [`CardClient.vault()` method](https://github.com/paypal/paypal-ios/blob/8b04e5fda77152afa860ea897448479e192ff256/Sources/CardPayments/CardClient.swift#L42)

KunJeongPark: This is for vault with pay during checkout. vault method will be for vault without payment flow. There are no changes in the client SDK for vault with purchase. We are starting to add merchant actions that mimic the flow of different merchant integrations in the demo app.

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderCompletionResultView.swift
KunJeongPark: Addressed 8b7702c71cdef997347e2ea250ff88f6bd107d57

Demo/Demo/ViewModels/CardPaymentViewModel.swift
scannillo: Ah, I see! Thank you for clarifying that. How do we make that clear to the merchants via both our docstrings & our dev docs?

Maybe we update the docstrings for [`CardRequest.orderID`](https://github.com/paypal/paypal-ios/blob/8b04e5fda77152afa860ea897448479e192ff256/Sources/CardPayments/Models/CardRequest.swift#L5-L6) to specify that if they set the `storeInVault` property in their orderID request, it will be respected & vaulted.

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -15,3 +16,30 @@ struct Amount: Codable {
...
+
+struct VaultCard: Codable {
+
+    let attributes: Attributes
+}
+
+struct VaultCardPaymentSource: Codable {
+
+    let card: VaultCard
+}
```
scannillo: Styling suggestion - IMO it's more readable to define these in the order that they appear in the struct above

```suggestion

struct VaultCardPaymentSource: Codable {

    let card: VaultCard
}

struct VaultCard: Codable {

    let attributes: Attributes
}

struct Attributes: Codable {

    let vault: Vault
    let customer: CardVaultCustomer?
}

struct Vault: Codable {

    let storeInVault: String
}

struct CardVaultCustomer: Codable {

    let id: String
}
```

scannillo: Another nitpick - use `Encodable` instead of `Codable` since we never will decode this model type

Demo/Demo/ViewModels/CardPaymentViewModel.swift
KunJeongPark: I don't think we have any documentation for vaulting yet. Should we include this as part of documentation ticket for vaulting with purchase when we are ready?

Demo/Demo/Models/CreateOrderParams.swift
KunJeongPark: Addressed in e1e0fa3a4110e34eef0e850107e304ec56e8377c (changing above structs to Encodable) and  8e95c5d1aff1212b258d840f60f056c4359c5c18 (reordering in CreateOrderParam along with merge conflict resolution, changing ApplicationContext structs from [this PR](https://github.com/paypal/paypal-ios/pull/209) to Encodable as well)

Demo/Demo/ViewModels/CardPaymentViewModel.swift
scannillo: Either way works for me, as long as we don't forget to document that it's the `orderID` that is the magic piece required for vault with purchase

Demo/Demo/Models/Order.swift
```diff
@@ -1,10 +1,4 @@
 struct Order: Codable, Equatable {
```
scannillo, line 1: Nitpick / separate PR to make this Decodable only (mostly for readability)

Demo/Demo/ViewModels/CardPaymentViewModel.swift
```diff
@@ -10,15 +10,34 @@ class CardPaymentViewModel: ObservableObject, CardDelegate {
...
+        amount: String,
+        selectedMerchantIntegration: MerchantIntegration,
+        intent: String,
+        shouldVault: Bool,
+        customerID: String? = nil
+    ) async throws {
         // might need to pass in payee as payee object or as auth header
 
         let amountRequest = Amount(currencyCode: "USD", value: amount)
         // TODO: might need to pass in payee as payee object or as auth header
```
scannillo: Are these TODOs still relevant? Also they seem to be duplicate, we can remove 1

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardOrderCompletionResultView.swift
scannillo: Nice! 👏

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardPaymentOrderCompletionView.swift
```diff
@@ -6,21 +6,33 @@ struct CardPaymentOrderCompletionView: View {
...
-            CardApprovalResultView(cardPaymentViewModel: cardPaymentViewModel)
-            if cardPaymentViewModel.state.approveResult != nil {
-                CardOrderActionButton(
-                    intent: cardPaymentViewModel.state.intent,
-                    orderID: orderID,
-                    selectedMerchantIntegration: DemoSettings.merchantIntegration,
-                    cardPaymentViewModel: cardPaymentViewModel
-                )
-            }
+        let state = cardPaymentViewModel.state
```
scannillo, line 9: Do we need to set this as a variable? Might be more clear to just reference `cardPaymentViewModel.state` when needed

scannillo:
Just want to confirm, that this `Card Vaulting` feature is not yet working and was built for Vault w/o Purchase? Can we make that more clear in the UI?

Screenshot 2023-10-18 at 11 00 45 AM

scannillo:
Just a question for some cleanup/organization - what is the difference between `CardPaymentOrderCompletionView` & `CardOrderCompletionResultView`? Maybe in a future PR we can standardize the naming a bit of the views within the `CardPaymentViews` directory

scannillo (approved):
Thanks for addressing my questions! Left a few other styling/cleanup type comments, but otherwise LGTM!

Demo/Demo/SwiftUIComponents/CardPaymentViews/CardPaymentOrderCompletionView.swift
KunJeongPark: It was referencing Jax's feedback, that there were multiple references to this and that it might be more readable to have a simpler variable.

KunJeongPark:
> Just want to confirm, that this `Card Vaulting` feature is not yet working and was built for Vault w/o Purchase? Can we make that more clear in the UI?
> 
> Screenshot 2023-10-18 at 11 00 45 AM

Card vaulting without purchase is working from test cards we've used but we are blocked on 3DS for this feature for now.

Demo/Demo/ViewModels/CardPaymentViewModel.swift
KunJeongPark: Great catch! I left a comment I made for myself not realizing I already added one a while back.
I think it's good to note for now as reminder for different scenarios we may address in the demo app.

PR #212, opened 2023-10-23
Remove Async keyword from PayPalNativeCheckoutClient.start function and fix tests

KunJeongPark:
### Summary of changes

- Remove async from PayPalNativeCheckoutClient.start function since
the results are returned via delegate methods.
- Checked with Project build setting Strict Concurrency Checking option set to complete for this change,
    no warnings, errors found. There are warnings and compile errors in other modules separate from this PR, 
    so I made another ticket to address them.

-Reverted above changes with comment to remove Async for next major version.

### Checklist

### Authors

- @KunJeongPark @sshropshire

### DIFF
--- a/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
+++ b/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
@@ -35,13 +35,17 @@ public class PayPalNativeCheckoutClient {
         self.networkingClient = networkingClient
     }
 
-    private func startPayPalNativeCheckout(
+    /// Present PayPal Paysheet and start a PayPal transaction
+    /// - Parameters:
+    ///   - request: The PayPalNativeCheckoutRequest for the transaction
+    ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
+    public func start(
         request: PayPalNativeCheckoutRequest,
         presentingViewController: UIViewController? = nil
-    ) {
+    ) async {
         correlationID = State.correlationIDs.riskCorrelationID
         analyticsService = AnalyticsService(coreConfig: config, orderID: request.orderID)
-
+        
         let nxoConfig = CheckoutConfig(
             clientID: config.clientID,
             createOrder: nil,
@@ -52,7 +56,7 @@ public class PayPalNativeCheckoutClient {
             environment: config.environment.toNativeCheckoutSDKEnvironment()
         )
         delegate?.paypalWillStart(self)
-
+        
         analyticsService?.sendEvent("paypal-native-payments:started")
         self.nativeCheckoutProvider.start(
             presentingViewController: presentingViewController,
@@ -90,30 +94,7 @@ public class PayPalNativeCheckoutClient {
             nxoConfig: nxoConfig
         )
     }
-
-    /// Present PayPal Paysheet and start a PayPal transaction
-    /// - Parameters:
-    ///   - request: The PayPalNativeCheckoutRequest for the transaction
-    ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
-    public func startNativeCheckout(
-        request: PayPalNativeCheckoutRequest,
-        presentingViewController: UIViewController? = nil
-    ) {
-        startPayPalNativeCheckout(request: request, presentingViewController: presentingViewController)
-    }
-
-    /// Present PayPal Paysheet and start a PayPal transaction
-    /// - Parameters:
-    ///   - request: The PayPalNativeCheckoutRequest for the transaction
-    ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
-    @available(*, deprecated, message: "Async function is deprecated. Use the non-async start function")
-    public func start(
-        request: PayPalNativeCheckoutRequest,
-        presentingViewController: UIViewController? = nil
-    ) async {
-        startPayPalNativeCheckout(request: request, presentingViewController: presentingViewController)
-    }
-
+    
     private func notifySuccess(for result: PayPalNativeCheckoutResult) {
         analyticsService?.sendEvent("paypal-native-payments:succeeded", correlationID: nativeCheckoutProvider.correlationID)
         delegate?.paypal(self, didFinishWithResult: result)
--- a/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
+++ b/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
@@ -39,6 +39,7 @@ public class PayPalNativeCheckoutClient {
     /// - Parameters:
     ///   - request: The PayPalNativeCheckoutRequest for the transaction
     ///   - presentingViewController: the ViewController to present PayPalPaysheet on, if not provided, the Paysheet will be presented on your top-most ViewController
+    // NEXT_MAJOR_VERSION: - Change to non-async
     public func start(
         request: PayPalNativeCheckoutRequest,
         presentingViewController: UIViewController? = nil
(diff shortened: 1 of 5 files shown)


sshropshire (approved):

Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
```diff
@@ -42,7 +42,7 @@ public class PayPalNativeCheckoutClient {
     public func start(
         request: PayPalNativeCheckoutRequest,
         presentingViewController: UIViewController? = nil
-    ) async {
+    ) {
```
jaxdesmarais: We will want to add a CHANGELOG entry for this since it's a public method

sshropshire: That's a good call. Is this considered a breaking change in iOS?

scannillo: Yes this is a breaking change, if we made this change a merchant app would not compile as-is

scannillo:
Our [repo docs](https://github.com/paypal/paypal-ios/tree/main/docs/PayPalNativePayments) also need to be updated for this change, and our dev docs as well

Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
scannillo: I would deprecate the old method until we're ready for V2

sshropshire: Definitely yeah that makes sense 👍 .

KunJeongPark: Steven and I tried creating another function with same name with async and annotate it as deprecated.
There is a problem with the calling function with same name as deprecated async function in the Task block. Even without the "await" keyword, it still expects the async function of same name because of the async Task context, causing a compile error. I thought this was strange behavior.
There is another async function call in the Task block, so we need the Task initializer there.
I just changed the new function name.

KunJeongPark:
> Our [repo docs](https://github.com/paypal/paypal-ios/tree/main/docs/PayPalNativePayments) also need to be updated for this change, and our dev docs as well

Do we keep the documentation for deprecated functions here and in the dev docs? The dev docs needs to be another ticket, different repo. I'll hold on dev docs change until we agree on the new function name

Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
jaxdesmarais: I think the goal is to _not_ call this in the task since it shouldn't be async. We can either restructure how we are calling things in the demo to separate out the order ID call or add a `DispatchSemaphore` to wait for the async to `getOrderID` to finish:
```swift
    private func checkout(_ shippingPreference: ShippingPreference) {
        state = .loading(content: "Initializing checkout")
        let semaphore = DispatchSemaphore(value: 0)

        Task {
            do {
                orderID = try await self.getOrderID(shippingPreference)
                self.shippingPreference = shippingPreference
                semaphore.signal()
            } catch let error {
                publishStateToMainThread(.mainContent(title: "Error", content: "\(error.localizedDescription)", flowComplete: true))
            }
        }

        semaphore.wait()
        let request = PayPalNativeCheckoutRequest(orderID: orderID)
        payPalClient?.start(request: request)
    }
```

KunJeongPark: Hey Jax, do we really want to use semaphore? This function is reliant on result from getOrderID, that's why it's in the same Task block. Mixing old low level methods and new concurrency stuff seems weird to me. My take is to use the newer higher level stuff to make sure things work well together.

scannillo: I am able to replicate the weirdness you're seeing Victoria. It looks like the compiler _is_ able to distinguish between `func start() async` and `func start()` only if you move the non-async one out of a Task like in Jax's example. 

I also wasn't expecting that behavior, but it seems to be by design [(see Swift proposal where this got added)](https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#overloading-and-overload-resolution).

I agree it's weird though, since for non-async methods without naming conflicts, there is no enforcement that it must be called outside of a Task.

KunJeongPark: Yes, it's a weird assumption on Swift compiler since Task block is supposed to handle both async and sync functions.

jaxdesmarais: We don't have to as I mentioned in my comment you can also restructure the method calls. Though a semaphore is calling down to the kernel only when the calling thread needs to be blocked, which as you mentioned is the goal since we need the result from the `getOrderID` call to call `start`. 

Is there a specific reason why the `getOrderID` call is happening at the start of the checkout flow instead of earlier? Looking though PayPal Web for example, we have this as a separate step/view? If anything I think `getOrderID` happening in this same task is an architecture issue since this is our only view/viewmodel that appears to do it at this point.

KunJeongPark: I don't think there was a good reason for this in the demo app except maybe there is extra step of verifying clientID sent into for native checkout and maybe it was decided not to have that extra step to check orderID?

KunJeongPark: I guess that is a good solution for not having to rename the function. Should we rename the new function? What would be easier for the integrators? I find having same name for both confusing.

jaxdesmarais: I don't think it's confusing to use the same method name for the async/non async method. This also means we won't have to go back and change the name to the one we actually want once we remove the deprecated method.

scannillo:
> Do we keep the documentation for deprecated functions here and in the dev docs? 

Great question. If a function is deprecated in our SDK, our docs should not be using it. The docs should be updated to showcase the newest, suggested way of integrating.

> The dev docs needs to be another ticket, different repo. I'll hold on dev docs change until we agree on the new function name

Agreed! Dev docs is another repo, so a separate ticket is a great call

Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
scannillo: It is pretty annoying that Xcode won't let you use the non-async `start()` function in a Task just because one exists with a duplicate name. That isn't the behavior of a method without a duplicate name.

Also this is the first time we're dealing with the fact that Swift doesn't support silencing deprecation warnings via Macros ([example](https://stackoverflow.com/questions/31540446/how-to-silence-a-warning-in-swift)).

We _could_ leave a `NEXT_MAJOR_VERSION` note to change this function signature later on when we're ready for v2? Or lean into how Swift handles this and refactor our demo.

KunJeongPark: Steven's vote was for change in v2 since it is a breaking change. 

And the issue that I ran into with the naming, merchant might run into also with duplicate names with deprecated method. We break out getting orderID in the demo app for testing purposes in other payment methods (I will make a ticket to do so for native payment module as well), but we have no guarantee of how merchant integration works.

scannillo: Cool, I'm down if we want to leave a `NEXT_MAJOR_VERSION`. We use that tag so that when time comes for doing a breaking major version release we have all of our TODOs related to it easily searchable

KunJeongPark: Sorry, but is it just comment // NEXT_MAJOR_VERSION? I don't think I've seen it.

KunJeongPark: Just noting that we can have the two separate async and non-async functions if we rename the new function as I did in earlier commits or just assume merchants will not run into issue of having non-async function in a Task block with another async function ( which I don't think we should )

Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift
```diff
@@ -43,6 +43,7 @@ public class PayPalNativeCheckoutClient {
         request: PayPalNativeCheckoutRequest,
         presentingViewController: UIViewController? = nil
     ) async {
+        // NEXT_MAJOR_VERSION: - Change to non-async
```
jaxdesmarais: I think we want to move this above `/// Present PayPal Paysheet and start a PayPal transaction`

KunJeongPark: Addressed here: dd91cffd9ccb5cd849acaf586d5379a2c82d3dff

jaxdesmarais (approved):

PR #214, opened 2023-10-31
Resolve Compile Errors with Strict Swift Concurrency Checking

KunJeongPark:
### Reason for changes
Apple has suggested you work towards complete checking, not just to avoid potential problems, but also in preparation for the expected checking that a future release of Swift 6 will enforce. 

In WWDC 2022 video "[Eliminate data races using Swift Concurrency](https://developer.apple.com/videos/play/wwdc2022/110351/)", they state "Complete checking approximates the intended Swift 6 semantics to completely eliminate data races".

### Summary of changes
- Resolve compile errors with Swift Concurrency Check set to "Complete" in preparation for Swift 6 
    - References to properties in UIDevice, UIApplication in structs or classes result in an error unless
      the struct or classes or specific functions are referencing these properties are also explicitly guaranteed to run
      on main thread.
    - Unit tests using waitForExpectations function needed to be annotated with @MainActor to resolve compile errors
    - In PaymentButton's subclasses, their Representable structs that wrap the UIKit elements for SwiftUI views, needed to have initializers that are annotated @MainActor to resolve compile errors. 
   
      PaymentButton is a subclass of UIButton which is annotated as @MainActor. I think it is not enforced for the UIButton subclass initializers because of backward compatibility.
    
I will address some Sendable warnings in [another PR](https://github.com/paypal/paypal-ios/pull/215). 
We can decide when we want to implement these changes. 
It's good to be proactive in at least spiking potential scope of work before Swift 6.

### Checklist

### Authors
- @KunJeongPark

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -161,6 +161,7 @@ public class CardClient: NSObject {
 
 extension CardClient: ASWebAuthenticationPresentationContextProviding {
 
+    @MainActor
     public func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
         if #available(iOS 15, *) {
             let firstScene = UIApplication.shared.connectedScenes.first as? UIWindowScene
--- a/Sources/CorePayments/AnalyticsEventData.swift
+++ b/Sources/CorePayments/AnalyticsEventData.swift
@@ -1,5 +1,6 @@
 import UIKit
 
+@MainActor
 struct AnalyticsEventData: Encodable {
     
     enum TopLevelKeys: String, CodingKey {
--- a/Sources/CorePayments/AnalyticsEventData.swift
+++ b/Sources/CorePayments/AnalyticsEventData.swift
@@ -102,7 +102,7 @@ struct AnalyticsEventData: Encodable {
         self.correlationID = correlationID
     }
     
-    func encode(to encoder: Encoder) throws {
+    nonisolated func encode(to encoder: Encoder) throws {
         var topLevel = encoder.container(keyedBy: TopLevelKeys.self)
         var events = topLevel.nestedContainer(keyedBy: EventKeys.self, forKey: .events)
         var eventParameters = events.nestedContainer(keyedBy: EventParameterKeys.self, forKey: .eventParameters)
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -49,7 +49,7 @@ public struct AnalyticsService {
         do {
             let clientID = coreConfig.clientID
             
-            let eventData = AnalyticsEventData(
+            let eventData = await AnalyticsEventData(
                 environment: coreConfig.environment.toString,
                 eventName: name,
                 clientID: clientID,
--- a/UnitTests/CardPaymentsTests/CardClient_Tests.swift
+++ b/UnitTests/CardPaymentsTests/CardClient_Tests.swift
@@ -5,6 +5,7 @@ import AuthenticationServices
 @testable import CardPayments
 @testable import TestShared
 
+@MainActor
 class CardClient_Tests: XCTestCase {
 
     // MARK: - Helper Properties
--- a/UnitTests/CardPaymentsTests/CardClient_Tests.swift
+++ b/UnitTests/CardPaymentsTests/CardClient_Tests.swift
@@ -5,7 +5,6 @@ import AuthenticationServices
 @testable import CardPayments
 @testable import TestShared
 
-@MainActor
 class CardClient_Tests: XCTestCase {
 
     // MARK: - Helper Properties
@@ -47,7 +46,7 @@ class CardClient_Tests: XCTestCase {
     
     // MARK: - vault() tests
 
-    func testVault_withValidResponse_returnsSuccess() {
+    func testVault_withValidResponse_returnsSuccess() async throws {
         let setupTokenID = "testToken1"
         let vaultStatus = "APPROVED"
         let vaultRequest = CardVaultRequest(card: card, setupTokenID: setupTokenID)
@@ -67,10 +66,10 @@ class CardClient_Tests: XCTestCase {
         sut.vaultDelegate = cardVaultDelegate
         sut.vault(vaultRequest)
         
-        waitForExpectations(timeout: 10)
+        await self.fulfillment(of: [expectation], timeout: 10)
     }
-    
-    func testVault_whenVaultAPIError_bubblesError() {
+
+    func testVault_whenVaultAPIError_bubblesError() async throws {
         let setupTokenID = "testToken1"
         let vaultRequest = CardVaultRequest(card: card, setupTokenID: setupTokenID)
                
@@ -88,10 +87,10 @@ class CardClient_Tests: XCTestCase {
         sut.vaultDelegate = cardVaultDelegate
         sut.vault(vaultRequest)
 
-        waitForExpectations(timeout: 10)
+        await self.fulfillment(of: [expectation], timeout: 10)
     }
 
-    func testVault_whenUnknownError_returnsVaultError() {
+    func testVault_whenUnknownError_returnsVaultError() async throws {
         let setupTokenID = "testToken1"
         let vaultRequest = CardVaultRequest(card: card, setupTokenID: setupTokenID)
 
@@ -109,12 +108,12 @@ class CardClient_Tests: XCTestCase {
         sut.vaultDelegate = cardVaultDelegate
         sut.vault(vaultRequest)
 
-        waitForExpectations(timeout: 10)
+        await self.fulfillment(of: [expectation], timeout: 10)
     }
 
     // MARK: - approveOrder() tests
 
-    func testApproveOrder_withInvalid3DSURL_returnsError() {
+    func testApproveOrder_withInvalid3DSURL_returnsError() async throws {
         mockCheckoutOrdersAPI.stubConfirmResponse = FakeConfirmPaymentResponse.withInvalid3DSURL
         
         let expectation = expectation(description: "approveOrder() completed")
@@ -133,10 +132,10 @@ class CardClient_Tests: XCTestCase {
         sut.delegate = mockCardDelegate
         sut.approveOrder(request: cardRequest)
 
-        waitForExpectations(timeout: 10)
+        await self.fulfillment(of: [expectation], timeout: 10)
     }
 
-    func testApproveOrder_withNoThreeDSecure_returnsOrderData() {
+    func testApproveOrder_withNoThreeDSecure_returnsOrderData() async throws {
         mockCheckoutOrdersAPI.stubConfirmResponse = FakeConfirmPaymentResponse.without3DS
         
         let expectation = expectation(description: "approveOrder() completed")
@@ -153,10 +152,10 @@ class CardClient_Tests: XCTestCase {
         sut.delegate = mockCardDelegate
         sut.approveOrder(request: cardRequest)
 
-        waitForExpectations(timeout: 10)
+        await self.fulfillment(of: [expectation], timeout: 10)
     }
 
-    func testApproveOrder_whenConfirmPaymentSDKError_bubblesError() {
+    func testApproveOrder_whenConfirmPaymentSDKError_bubblesError() async throws {
         mockCheckoutOrdersAPI.stubError = CoreSDKError(code: 123, domain: "sdk-domain", errorDescription: "sdk-error-desc")
 
         let expectation = expectation(description: "approveOrder() completed")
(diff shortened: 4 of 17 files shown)


Sources/CorePayments/AnalyticsEventData.swift
```diff
@@ -101,7 +102,7 @@ struct AnalyticsEventData: Encodable {
         self.correlationID = correlationID
     }
     
-    func encode(to encoder: Encoder) throws {
+    nonisolated func encode(to encoder: Encoder) throws {
         var topLevel = encoder.container(keyedBy: TopLevelKeys.self)
```
KunJeongPark, line 106: This just gave warning during compile time but caused failure in Cocoapods test in the pipeline.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -49,7 +49,7 @@ public struct AnalyticsService {
         do {
             let clientID = coreConfig.clientID
             
-            let eventData = AnalyticsEventData(
+            let eventData = await AnalyticsEventData(
```
sshropshire, line 52: q: This is an asynchronous constructor?

jaxdesmarais:
It seems like there is also an intermediary `Targeted` level of checking before `Complete` that we can consider for making this move progressively. I found [this medium article](https://medium.com/@mega-blog/adopting-targeted-swift-strict-concurrency-setting-ba9e92ed06bb) helpful for talking about these checks. Since we are an SDK we likely won't be able to release with Swift 6 until a bit after the release since that is a breaking change. With that in mind we should consider if we want to jump right to `Complete` or start with `Targeted`.

Sources/CorePayments/Networking/AnalyticsService.swift
KunJeongPark: Because AnalyticsEventData was marked with @MainActor, to instantiate it in async function that isn't isolated to the @MainActor, I needed "await".

KunJeongPark:
> It seems like there is also an intermediary `Targeted` level of checking before `Complete` that we can consider for making this move progressively. I found [this medium article](https://medium.com/@mega-blog/adopting-targeted-swift-strict-concurrency-setting-ba9e92ed06bb) helpful for talking about these checks. Since we are an SDK we likely won't be able to release with Swift 6 until a bit after the release since that is a breaking change. With that in mind we should consider if we want to jump right to `Complete` or start with `Targeted`.

I checked with Targeted option and we have no errors or warnings other than same old warnings without the concurrency checking. I think for the compile errors with complete option, the only breaking change would be for the buttons if merchant is not instantiating them on main thread. 

I think marking AnalyticsEventData @MainActor is ok since the work in there is lightweight. Alternative would be to fetch device information somewhere else and store it for access here, I don't see an obvious place to store the info. I can implement something like DeviceInfoCache. I thought @MainActor was the simplest way to deal with compile errors. 

I think it would be surprising if any version of Swift in the future would enforce this one, any reference to UIDevice's read only properties, for example, giving compile errors without explicit main thread context.

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -5,6 +5,7 @@ import AuthenticationServices
 @testable import CardPayments
 @testable import TestShared
 
+@MainActor
```
jaxdesmarais: For the tests, it looks like we may have the opportunity to make the test async and await on `waitForExpectations` instead ([from this swift forum](https://forums.swift.org/t/tasks-closure-not-called-in-async-xctest-function/59093/3)). Not sure if you've had a chance to try this yet.

Sources/CardPayments/CardClient.swift
```diff
@@ -161,6 +161,7 @@ public class CardClient: NSObject {
 
 extension CardClient: ASWebAuthenticationPresentationContextProviding {
 
+    @MainActor
```
jaxdesmarais, line 164: I'm curious what portion of this method needs to be called on the main thread. Is it our default to `ASPresentationAnchor` if we cannot find a window or something else?

jaxdesmarais (commented):
Are the test updates that we needed to make an indication at all that these would be a breaking change for merchants or are these changes mainly behind the scenes? What changes would this cause for merchant demo apps that also have swift strict concurrency checking enabled?

I also don't see anywhere in this PR where we are enabling this setting. Will that be in a future PR along with a changelog entry if needed?

KunJeongPark:
> all that these would be a breaking change for merchants or are these changes mainly behind the scenes? What changes would this cause for merchant demo apps that also have swift strict concurrency checking enabled?

The AnalyticEventData change, I think that's just us, so no breaking change for merchant. PresentationAnchor change, that function is used behind the scenes as part of ASWebAuthenticationSession, so I don't think it will affect merchants' implementations. It would only be problematic if merchant has the strict checking and we don't. 
The SwiftUI buttons, the change is only for the Representables, for merchants who are using it in SwiftUI and buttons on SwiftUIViews, I don't think there should be any difference in the way it is instantiated by the merchant. 

I didn't submit the complete concurrency checking option but I fixed compile errors that occurred as a result of that option check.

Sources/CardPayments/CardClient.swift
KunJeongPark: We start the authentication session on the main thread in the start function. 
Since UIApplication is marked @MainActor, there is compile error with accessing its properties in this function with complete checking option since it checks for explicit main thread context. This seems extreme to me, as it would break many implementations for others as well.

Sources/CorePayments/Networking/AnalyticsService.swift
KunJeongPark: I used this async function and it passed. What do you think?
 ```
 func testVault_withValidResponse_returnsSuccess() async throws {
        let setupTokenID = "testToken1"
        let vaultStatus = "APPROVED"
        let vaultRequest = CardVaultRequest(card: card, setupTokenID: setupTokenID)
        let updateSetupTokenResponse = UpdateSetupTokenResponse(
            updateVaultSetupToken: TokenDetails(id: setupTokenID, status: vaultStatus, links: [TokenDetails.Link(rel: "df", href: "h")])
        )
        mockVaultAPI.stubSetupTokenResponse = updateSetupTokenResponse
        
        let expectation = expectation(description: "vault completed")
        let cardVaultDelegate = MockCardVaultDelegate(success: {_, result in
            XCTAssertEqual(result.setupTokenID, setupTokenID)
            XCTAssertEqual(result.status, vaultStatus)
            expectation.fulfill()
        }, error: {_, _ in
            XCTFail("Invoked error() callback. Should invoke success().")
        })
        sut.vaultDelegate = cardVaultDelegate
        sut.vault(vaultRequest)
        
        await self.fulfillment(of: [expectation], timeout: 10)
    }
```

UnitTests/CardPaymentsTests/CardClient_Tests.swift
KunJeongPark: addressed in bd8cd78c0818a1627c64e549010f4af5af7910e4

jaxdesmarais:
> I didn't submit the complete concurrency checking option but I fixed compile errors that occurred as a result of that option check.

Personally, I think if we are adding the code to support complete strict concurrency checking, we should enable it in the SDK at the same time. This way it is clear to merchants/us when we added support and what the necessary code changes were. But curious to hear how other folks feel.

jaxdesmarais:
@KunJeongPark I just came across [this Swift Forum thread on Swift 6](https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315). There are a few things of note with regards to this PR:

> Readiness of data-race safety by default
There are still a number of bugs and holes in the Sendable checking model that admit data races under strict concurrency checking. Equally important, strict concurrency checking currently issues a significant number of false positive reports of data races, making the complete checking difficult to program against.
> 
> The Swift 6 language mode will only be declared ready once the Language Steering Group determines that the programming model is effective and usable. The remaining language work to complete data-race safety for the Swift 6 language mode will fall into two categories:
> 
> Fill all holes in the strict concurrency model so that data races are diagnosed statically, or dynamically where static safety cannot be proven.
Mitigate false reports of data races on patterns that are proven to be safe.
Swift 5.10 contains a number of significant bug fixes in the actor isolation and Sendable checking. In addition, the following language changes are currently being designed and implemented to undergo the Swift evolution review process:
> 
> [SE-0411: Isolated default value expressions 17](https://forums.swift.org/t/se-0411-isolated-default-value-expressions/68065)
[Inferring @Sendable for methods and key-path literals 15](https://forums.swift.org/t/pitch-inferring-sendable-for-methods-and-key-path-literals/68011)
[Strict concurrency checking for global and static variables 9](https://forums.swift.org/t/pitch-strict-concurrency-for-global-variables/66908)
[Improved control over closure actor isolation 13](https://github.com/apple/swift-evolution/blob/3644f682913c8d0b2eafb1945ff81105fbdc7783/proposals/NNNN-improved-control-over-closure-actor-isolation.md)
[Lifting restrictions on non-Sendable values through isolated value regions 12](https://forums.swift.org/t/pitch-region-based-isolation/67888)
> 
> Together, these changes fill the remaining major holes in strict concurrency checking and improve the usability of strict concurrency checking by introducing more Sendable inference and enabling safe ways to transfer non-Sendable values across isolation boundaries. The Language Steering Group acknowledges that other language changes in the area of concurrency are important, but the changes above are necessary for defining the Swift 6 language mode. Other concurrency features are additive and can be explored independently.

Based on these comments and since we do not have a major version planned to update to either Swift 5.10 (that includes some of these fixes) or Swift 6 (that does not yet have a targeted release date) it seems like we should hold off on these changes until the lauguage has ironed out some of these bugs. Curious to know what other folks think!

KunJeongPark:
> @KunJeongPark I just came across [this Swift Forum thread on Swift 6](https://forums.swift.org/t/progress-toward-the-swift-6-language-mode/68315). There are a few things of note with regards to this PR:
> 
> > Readiness of data-race safety by default
> > There are still a number of bugs and holes in the Sendable checking model that admit data races under strict concurrency checking. Equally important, strict concurrency checking currently issues a significant number of false positive reports of data races, making the complete checking difficult to program against.
> > The Swift 6 language mode will only be declared ready once the Language Steering Group determines that the programming model is effective and usable. The remaining language work to complete data-race safety for the Swift 6 language mode will fall into two categories:
> > Fill all holes in the strict concurrency model so that data races are diagnosed statically, or dynamically where static safety cannot be proven.
> > Mitigate false reports of data races on patterns that are proven to be safe.
> > Swift 5.10 contains a number of significant bug fixes in the actor isolation and Sendable checking. In addition, the following language changes are currently being designed and implemented to undergo the Swift evolution review process:
> > [SE-0411: Isolated default value expressions 17](https://forums.swift.org/t/se-0411-isolated-default-value-expressions/68065)
> > [Inferring @Sendable for methods and key-path literals 15](https://forums.swift.org/t/pitch-inferring-sendable-for-methods-and-key-path-literals/68011)
> > [Strict concurrency checking for global and static variables 9](https://forums.swift.org/t/pitch-strict-concurrency-for-global-variables/66908)
> > [Improved control over closure actor isolation 13](https://github.com/apple/swift-evolution/blob/3644f682913c8d0b2eafb1945ff81105fbdc7783/proposals/NNNN-improved-control-over-closure-actor-isolation.md)
> > [Lifting restrictions on non-Sendable values through isolated value regions 12](https://forums.swift.org/t/pitch-region-based-isolation/67888)
> > Together, these changes fill the remaining major holes in strict concurrency checking and improve the usability of strict concurrency checking by introducing more Sendable inference and enabling safe ways to transfer non-Sendable values across isolation boundaries. The Language Steering Group acknowledges that other language changes in the area of concurrency are important, but the changes above are necessary for defining the Swift 6 language mode. Other concurrency features are additive and can be explored independently.
> 
> Based on these comments and since we do not have a major version planned to update to either Swift 5.10 (that includes some of these fixes) or Swift 6 (that does not yet have a targeted release date) it seems like we should hold off on these changes until the lauguage has ironed out some of these bugs. Curious to know what other folks think!

Those are great finds from the forums. I agree, I did have a conversation with some people and consensus is to mark possible issues and keep the branches updated periodically and revisit as warnings and errors in Xcode get refined for next versions.

 I will keep making changes, checking forums, keeping the branches updated and I look forward to input from all of you whenever you have a chance. I think it's good for us to be ahead of it and start thinking about possible issues.

Sources/CorePayments/AnalyticsEventData.swift
warmkesselj: Curious if you could add the Sendable protocol to this struct to resolve this issue? 

```struct AnalyticsEventData: Encodable, Sendable {```

KunJeongPark: No, it does not.

jaxdesmarais:
Closing due to inactivity, when we are ready to prioritize this work we can reopen this PR or start a new PR using this as a reference

PR #216, opened 2023-11-07
Add Authorize/Capture for PayPalWebPayments in Demo App

KunJeongPark:
### Summary of changes

- Create authorize/capture on order in demo app after successful checkout 

https://github.com/paypal/paypal-ios/assets/9273272/321a510a-eea6-4ac0-ada2-2d31cf23bdac

### Checklist

### Authors
- @KunJeongPark

### DIFF
--- a/Demo/Demo/Models/Order.swift
+++ b/Demo/Demo/Models/Order.swift
@@ -6,7 +6,8 @@ struct Order: Codable, Equatable {
     
     struct PaymentSource: Codable, Equatable {
         
-        let card: Card
+        let card: Card?
+        let paypal: PayPal?
     }
 
     init(id: String, status: String, paymentSource: PaymentSource? = nil) {
@@ -20,6 +21,13 @@ struct Order: Codable, Equatable {
         let lastDigits: String?
         let brand: String?
         let attributes: Attributes?
+    }
+
+    struct PayPal: Codable, Equatable {
+
+        let emailAddress: String
+        let accountId: String
+        let accountStatus: String
     }
 
     struct Attributes: Codable, Equatable {
--- a/Demo/Demo/Models/Order.swift
+++ b/Demo/Demo/Models/Order.swift
@@ -26,8 +26,6 @@ struct Order: Codable, Equatable {
     struct PayPal: Codable, Equatable {
 
         let emailAddress: String
-        let accountId: String
-        let accountStatus: String
     }
 
     struct Attributes: Codable, Equatable {
--- a/Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
+++ b/Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
@@ -1,6 +1,6 @@
 import SwiftUI
 
-struct OrderCreateCardPaymentResultView: View {
+struct OrderCreateCardResultView: View {
 
     @ObservedObject var cardPaymentViewModel: CardPaymentViewModel
 
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
@@ -1,10 +1,3 @@
-//
-//  PayPalOrderActionButton.swift
-//  Demo
-//
-//  Created by Victoria Park on 11/7/23.
-//
-
 import SwiftUI
 
 struct PayPalOrderActionButton: View {
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
@@ -24,29 +24,15 @@ struct PayPalOrderActionButton: View {
     }
 
     private func completeOrder() {
-        if intent == .capture {
-            Task {
-                do {
-                    try await paypalWebViewModel.captureOrder(
-                        orderID: orderID,
-                        selectedMerchantIntegration: selectedMerchantIntegration
-                    )
-                    print("Order Captured. ID: \(paypalWebViewModel.state.capturedOrder?.id ?? "")")
-                } catch {
-                    print("Error capturing order: \(error.localizedDescription)")
-                }
-            }
-        } else {
-            Task {
-                do {
-                    try await paypalWebViewModel.authorizeOrder(
-                        orderID: orderID,
-                        selectedMerchantIntegration: selectedMerchantIntegration
-                    )
-                    print("Order Authorized. ID: \(paypalWebViewModel.state.authorizedOrder?.id ?? "")")
-                } catch {
-                    print("Error authorizing order: \(error.localizedDescription)")
-                }
+        Task {
+            do {
+                try await paypalWebViewModel.completeOrder(
+                    with: intent,
+                    orderID: orderID,
+                    selectedMerchantIntegration: selectedMerchantIntegration
+                )
+            } catch {
+                print("Error capturing order: \(error.localizedDescription)")
             }
         }
     }
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
@@ -1,0 +1,60 @@
+//
+//  PayPalOrderActionButton.swift
+//  Demo
+//
+//  Created by Victoria Park on 11/7/23.
+//
+
+import SwiftUI
+
+struct PayPalOrderActionButton: View {
+
+    let intent: Intent
+    let orderID: String
+    let selectedMerchantIntegration: MerchantIntegration
+
+    @ObservedObject var paypalWebViewModel: PayPalWebViewModel
+
+    var body: some View {
+        ZStack {
+            Button("\(intent.rawValue)") {
+                completeOrder()
+            }
+            .buttonStyle(RoundedBlueButtonStyle())
+            .padding()
+
+            if .loading == paypalWebViewModel.state.authorizedOrderResponse ||
+                .loading == paypalWebViewModel.state.capturedOrderResponse {
+                CircularProgressView()
+            }
+        }
+    }
+
+    private func completeOrder() {
+        if intent == .capture {
+            Task {
+                do {
+                    try await paypalWebViewModel.captureOrder(
+                        orderID: orderID,
+                        selectedMerchantIntegration: selectedMerchantIntegration
+                    )
+                    print("Order Captured. ID: \(paypalWebViewModel.state.capturedOrder?.id ?? "")")
+                } catch {
+                    print("Error capturing order: \(error.localizedDescription)")
+                }
+            }
+        } else {
+            Task {
+                do {
+                    try await paypalWebViewModel.authorizeOrder(
+                        orderID: orderID,
+                        selectedMerchantIntegration: selectedMerchantIntegration
+                    )
+                    print("Order Authorized. ID: \(paypalWebViewModel.state.authorizedOrder?.id ?? "")")
+                } catch {
+                    print("Error authorizing order: \(error.localizedDescription)")
+                }
+            }
+        }
+    }
+}
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
@@ -37,7 +37,7 @@ struct PayPalTransactionButtonsView: View {
                     Text("Complete Order Transaction")
                 }
                 .buttonStyle(RoundedBlueButtonStyle())
-                    .padding()
+                .padding()
             }
             Spacer()
         }
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
@@ -30,6 +30,15 @@ struct PayPalTransactionButtonsView: View {
                     .padding(5)
             )
             PayPalWebApprovalView(paypalWebViewModel: paypalWebViewModel)
+            if paypalWebViewModel.state.checkoutResult != nil {
+                NavigationLink {
+                    PayPalWebOrderCompletionView(orderID: orderID, payPalWebViewModel: paypalWebViewModel)
+                } label: {
+                    Text("Complete Order Transaction")
+                }
+                .buttonStyle(RoundedBlueButtonStyle())
+                    .padding()
+            }
             Spacer()
         }
     }
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionsView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionsView.swift
@@ -1,7 +1,7 @@
 import SwiftUI
 import PaymentButtons
 
-struct PayPalTransactionButtonsView: View {
+struct PayPalTransactionsView: View {
 
     @ObservedObject var paypalWebViewModel: PayPalWebViewModel
     let orderID: String
@@ -29,7 +29,7 @@ struct PayPalTransactionButtonsView: View {
                     .stroke(.gray, lineWidth: 2)
                     .padding(5)
             )
-            PayPalWebApprovalView(paypalWebViewModel: paypalWebViewModel)
+            PayPalWebApprovalResultView(paypalWebViewModel: paypalWebViewModel)
             if paypalWebViewModel.state.checkoutResult != nil {
                 NavigationLink {
                     PayPalWebOrderCompletionView(orderID: orderID, payPalWebViewModel: paypalWebViewModel)
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
@@ -35,7 +35,10 @@ struct PayPalWebOrderCompletionResultView: View {
             LeadingText("\(orderResponse.id)")
             LeadingText("Status", weight: .bold)
             LeadingText("\(orderResponse.status)")
-            
+            if let email = orderResponse.paymentSource?.paypal?.emailAddress {
+                LeadingText("Email", weight: .bold)
+                LeadingText("\(email)")
+            }
             Text("")
                 .id("bottomView")
         }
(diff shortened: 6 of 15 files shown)


Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
```diff
@@ -0,0 +1,60 @@
+//
+//  PayPalOrderActionButton.swift
+//  Demo
+//
+//  Created by Victoria Park on 11/7/23.
+//
```
sshropshire:
```suggestion
```
Good to remove the headers.

KunJeongPark: Oops I thought I got them all. Thanks Steven.

sshropshire (approved):
🌊

Demo/Demo/Models/Order.swift
```diff
@@ -22,6 +23,13 @@ struct Order: Codable, Equatable {
         let attributes: Attributes?
     }
 
+    struct PayPal: Codable, Equatable {
+
+        let emailAddress: String
+        let accountId: String
```
jaxdesmarais:
```suggestion
        let accountID: String
```

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
```diff
@@ -30,6 +30,15 @@ struct PayPalTransactionButtonsView: View {
...
             )
             PayPalWebApprovalView(paypalWebViewModel: paypalWebViewModel)
+            if paypalWebViewModel.state.checkoutResult != nil {
+                NavigationLink {
+                    PayPalWebOrderCompletionView(orderID: orderID, payPalWebViewModel: paypalWebViewModel)
+                } label: {
+                    Text("Complete Order Transaction")
+                }
+                .buttonStyle(RoundedBlueButtonStyle())
+                    .padding()
```
jaxdesmarais: Should this indent be back a bit or should it be in the `.buttonStyle`? Not sure which element it applies to based on the spacing.

Demo/Demo/Models/Order.swift
```diff
@@ -22,6 +23,13 @@ struct Order: Codable, Equatable {
         let attributes: Attributes?
     }
 
+    struct PayPal: Codable, Equatable {
```
jaxdesmarais, line 26: It doesn't look like we are using this struct anywhere currently, are there plans to use this in a future PR?

KunJeongPark: I was thinking about our setupToken for vault without purchase for future. We have paymentSource struct and in iOS we have hard coded Card in there so I was thinking about this option.

KunJeongPark: Without this struct,  decoding compile error on response with PaymentSource because it is expecting format with PayPal field for PayPal payments

KunJeongPark: I see what you meant. I wasn't sure which fields I might use for displaying results for this PR or for the vault without purchase PR. For now, I wanted to make it explicit that the PaymentSource we are using here is both for card payments and for PayPalPayment

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
KunJeongPark: it is a modifier for the NavigationLink which is pretty much a button. My indentation is off. I will fix it.

Demo/Demo/Models/Order.swift
KunJeongPark: To use codingKey to change this, I have to include every field from the response in the CodingKeys, I think?

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionButtonsView.swift
KunJeongPark: Addressed here: 293097be913c95f2564c5f0fe0805570d66a9b4e

Demo/Demo/Models/Order.swift
jaxdesmarais: Maybe we can just keep `email_address` for now and display it somewhere similar to what we do for `lastDigits` with Card? We can always add to this when we get to the vault with purchase feature, but can wait to add to it until we decide what we actually need across iOS and Android for those views.

KunJeongPark: Addressed here: e8d1cf5f0218bab1e18e3717dcf07275907e50a9

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
```diff
@@ -0,0 +1,53 @@
+import SwiftUI
+
+struct PayPalWebOrderCompletionResultView: View {
```
scannillo, line 3: The names of these classes confuse me a bit. What is the difference b/w `PayPalWebOrderCompletionResultView` & `PayPalWebOrderCompletionView`?

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionView.swift
```diff
@@ -0,0 +1,38 @@
+import SwiftUI
+
+struct PayPalWebOrderCompletionView: View {
```
scannillo, line 3: This class looks pretty much the exact same as [CardPaymentOrderCompletionView.swift](https://github.com/paypal/paypal-ios/blob/main/Demo/Demo/SwiftUIComponents/CardPaymentViews/CardPaymentOrderCompletionView.swift). The views should be re-usable across payment method/ view model types.

This way if we want to update this view, we do it in 1 place versus multiple.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -82,6 +82,45 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+            DispatchQueue.main.async {
+                self.state.authorizedOrderResponse = .loaded(order)
+            }
+        } catch {
+            DispatchQueue.main.async {
+                self.state.authorizedOrderResponse = .error(message: error.localizedDescription)
+            }
+            print("Error authorizing order: \(error.localizedDescription)")
+        }
+    }
```
scannillo, line 134: Same here, this looks like a copy-paste from `CardPaymentViewModel` ([see snippet](https://github.com/paypal/paypal-ios/blob/a490ffdc891f5b5c3137749655f08bd9c7513339/Demo/Demo/ViewModels/CardPaymentViewModel.swift#L61-L99)). That duplication is usually a good sign that we should refactor.

Happy to talk through some refactor options. We could go the route of a parent class or a protocol to represent a generic `PaymentMethodViewModel` that has a set of default functionality that is re-usable across all payment methods (ex: auth, capture).

PayPal.xcodeproj/project.pbxproj
```diff
@@ -3184,6 +3184,7 @@
 				SUPPORTS_MACCATALYST = YES;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "$(inherited) DEBUG";
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_STRICT_CONCURRENCY = minimal;
```
scannillo: Should this be included in this PR?

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
KunJeongPark: I have this view (`PayPalOrderCompletionView`) where I carry over order approval results to second screen to display authorize/capture results (`PayPalWebOrderCompletionResultView`)
<table>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

PayPal.xcodeproj/project.pbxproj
KunJeongPark: No. It snuck in there. I will get rid of it.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: That is a great suggestion. I will start with PaymentMethodViewModel protocol and incorporate it into common views.

PayPal.xcodeproj/project.pbxproj
KunJeongPark: Addressed here: 80ac172b81e04c24f3d9062ee5ee22e934b5546f

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: @scannillo , I looked into making a `PaymentMethodViewModel` protocol and extension to provide common default `authorizeOrder` and `captureOrder` function implementations and I ran into issue of having to change stored properties(states) in the conforming types. Protocol extensions, as I understand, cannot mutate stored properties of conforming types. For instance:  

`self.state.authorizedOrderResponse = .loaded(order)`

I could provide in `PaymentMethodViewModel` protocol delegate methods like `didAuthorizeOrder` and `didCaptureOrder` but it would reduce duplication only slightly and introduce complexity. 
I am looking into protocol with associatedType but it doesn't look pretty imo.

I ran into similar issue with factoring out views that looks the same - `state` in each ViewModel is of different type. I would need to use associated types and also run into issue of observing different properties of states of different types for each viewModels. 

I think benefits of reducing a bit of code duplication doesn't justify the complexity introduced.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionView.swift
KunJeongPark: See comment below.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
jaxdesmarais: If the order completion has already taken place maybe `PayPalWebResultView` is more appropriate?

Demo/Demo/ViewModels/PayPalWebViewModel.swift
jaxdesmarais: It seems like there could be a few options in this StackOverflow that may be worth exploring: https://stackoverflow.com/questions/38885622/swift-protocol-extensions-property-default-values. Not sure if Sammy had a different approach in mind as well!

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
KunJeongPark: There are several results in these flows. For PayPalWeb, I have order creation result, approval result and the order completion result (result of authorize/capture). So I am distinguishing them with the names.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: I will continue working w PaymentMethodViewModel protocol. I guess all both flows have views for order creation result, order approval result and order capture/authorize result.

KunJeongPark: I created another ticket for refactoring common views and functionalities in the viewModels. https://paypal.atlassian.net/browse/DTPPCPSDK-1437

Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
```diff
@@ -1,6 +1,6 @@
 import SwiftUI
 
-struct OrderCreateCardPaymentResultView: View {
+struct OrderCreateCardResultView: View {
```
jaxdesmarais, line 3: In the refactor ticket can we please include updating some of these names? This name doesn't seem to match up with what the code in this file is doing and it's not clear why it's being changed since there are no other code changes in this file. I don't think the original name was super clear but it's especially confusing we made this change with no other changes here.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
```diff
@@ -0,0 +1,53 @@
...
+
+            if .loading == paypalWebViewModel.state.authorizedOrderResponse ||
+                .loading == paypalWebViewModel.state.capturedOrderResponse {
+                CircularProgressView()
+            }
+        }
+    }
+
+    private func completeOrder() {
+        if intent == .capture {
```
jaxdesmarais: Since intent is an enum can we use a switch here instead?

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
```diff
@@ -0,0 +1,53 @@
...
+            Task {
+                do {
+                    try await paypalWebViewModel.captureOrder(
```
jaxdesmarais: It seems like both this and the else have basically the same code repeated. Can we pass the intent to the view model to dry up this code? Something like:
```swift
try await paypalWebViewModel(with: intent, orderID: orderID, selectedMerchantIntegration: selectedMerchantIntegration)
```
Then in the view model we can do a switch on the intent passed in.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionsView.swift
```diff
@@ -1,7 +1,7 @@
 import SwiftUI
 import PaymentButtons
 
-struct PayPalTransactionButtonsView: View {
+struct PayPalTransactionsView: View {
```
jaxdesmarais:
```suggestion
struct PayPalTransactionView: View {
```

KunJeongPark: There are three options of transactions on that view PayPal, PayPal Credit, Pay Later. That's why I put Transactions. What do you think?

Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
KunJeongPark: Basically we have result views for each Orders API calls to check if they are working as expected.
so there are OrderCreateCardResultView, OrderCreatePayPalWebResultView, etc.
And there are also container views since I didn't want them to be all on one scrollable page.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
KunJeongPark: I just used if since there are only two cases. I don't anticipate that there will be other forms of order operations after order approval but I am not tied to if-then.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
KunJeongPark: I have the intent toggle on the main screen for PayPalWebPayment.

 It would not make sense imo to instantiate the PayPalWebViewModel with intent option set. The pattern we have set is instantiating the viewModels as @StateObject with the main view for the payment method. We publish modifications in viewModel through actions from each API calls after.

jaxdesmarais: We aren't instantiating the view model here we are calling a method called `paypalWebViewModel.captureOrder/authroizeOrder`. The suggestion is to combine the logic here into 1 method on  `paypalWebViewModel`. It can be called whatever we want, so if order is more clear we can do `try await paypalWebViewModel.order(with: intent, orderID: orderID, selectedMerchantIntegration: selectedMerchantIntegration)`. I just removed the method name `captureOrder/authroizeOrder` to illustrate the idea.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionsView.swift
jaxdesmarais: I think the singular is fine even though it's used in multiple places there will ever only be 1 transaction

Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
jaxdesmarais: It sounds like this will be a great thing to refactor or can be shared in some way across payment methods since most of the logic is the same. I think that will help with naming as then we can have a singular `OrderResultView`.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
KunJeongPark: That makes sense. I will have one function that redirects from the viewModel. Thank you!

Demo/Demo/SwiftUIComponents/CardPaymentViews/OrderCreateCardResultView.swift
KunJeongPark: That is what I want to investigate in the refactor ticket. How to organize the different states to minimize views where it makes sense. Most of the result views have different content. There are two that are identical, order create result views and the "action buttons" for authorize/capture. To me, it's easier to follow if there are separate views for each API call result so we can find them more easily but I will explore different options.

That's why I wanted to create a different PR for the refactor as it will be quite a bit of code change in the viewModels and some backtracking and modifications based on feedback. I wanted to be able to track these changes.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -82,6 +82,57 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
         }
     }
 
+    func completeOrder(with intent: Intent, orderID: String, selectedMerchantIntegration: MerchantIntegration) async throws {
+
```
jaxdesmarais:
```suggestion
```

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalOrderActionButton.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    private func completeOrder() {
+        Task {
+            do {
+                try await paypalWebViewModel.completeOrder(
+                    with: intent,
+                    orderID: orderID,
+                    selectedMerchantIntegration: selectedMerchantIntegration
+                )
```
jaxdesmarais, line 33: 🎉

jaxdesmarais (approved):
Thanks for addressing all of my questions/suggestions!

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark, line 1: Thank you!

PR #218, opened 2023-11-13
Fixed Production GraphQLURL

quintonpryce:
### Reason for Changes
This pull request addresses the need to update the `graphQLURL` to the proper live URL. The current URL in the codebase is either outdated or points to a development/staging environment, which is not suitable for the production build.

### Summary of Changes
- Updated the `graphQLURL` in the relevant configuration file(s) to point to the correct live URL. This ensures that the application will interact with the live backend environment.

### Checklist
- [ ] Added a changelog entry to document this change. The entry is concise and clearly explains the nature of the update and its impact on the system.

### Authors
- Quinton Pryce

### DIFF
--- a/Sources/CorePayments/Networking/Enums/Environment.swift
+++ b/Sources/CorePayments/Networking/Enums/Environment.swift
@@ -19,7 +19,7 @@ public enum Environment {
         case .sandbox:
             return URL(string: "https://www.sandbox.paypal.com/graphql")!
         case .live:
-            return URL(string: "https://paypal.com/graphql")!
+            return URL(string: "https://www.paypal.com/graphql")!
         }
     }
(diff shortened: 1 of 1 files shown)


KunJeongPark:
Oh my! It looks like a typo on our end. 
I looked elsewhere in our codebase and I think url without "www." is resolved automatically. 

 But I think this change is good for clarity. 

Thank you for your contribution! 

Update 11/15/23 - I am checking with the team to make sure that these are interchangeable. According to a team member, adding www subdomain prevents a redirect. I am double checking to test both of these in live environment. Thank you!

scannillo:
Closing this PR due to inactivity, and the specified URLs are working as expected.

PR #220, opened 2023-11-14
Update GitHub Action CI for Forks

jaxdesmarais:
### Reason for changes

CI should run as expected on forks!

### Summary of changes

- We made this change a while ago on BT and never got around to adding it here but now that we have some outside contributors we should ensure that all PRs (even forks) work as expected
    - Pulled from the BT [test.yml](https://github.com/braintree/braintree_ios/blob/main/.github/workflows/tests.yml) and [build.yml](https://github.com/braintree/braintree_ios/blob/main/.github/workflows/build.yml)

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,5 @@
 name: Build
-on: [pull_request, workflow_dispatch]
+on: [pull_request]
 concurrency:
   group: build-${{ github.event.number }}
   cancel-in-progress: true
@@ -10,6 +10,9 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
       - name: Use Xcode 14
         run: sudo xcode-select -switch /Applications/Xcode_14.3.app
       - name: Run pod lib lint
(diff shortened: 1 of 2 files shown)


scannillo (approved):
👏 Thank you!

.github/workflows/build.yml
```diff
@@ -10,6 +10,9 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
KunJeongPark, line 15: Hey Jax, thanks for doing this! Can you tell me how this allows CI to run as expected on forks? Just high level explanation would be fine!

KunJeongPark (approved):

.github/workflows/build.yml
jaxdesmarais: ### Ref Line
`github.event` references where we want to run CI ([context docs](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context)) in our case we want to run CI on the `pull_request`. For PRs we want to run CI on the pull request's head branch name `head.ref` vs the base branch name ([PR workflow events docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)). So this line `github.event.pull_request.head.ref` is essentially saying: run CI on the branch the PR was opened from.

### Repository Line
`github.event.pull_request.head` is the same as above so it's saying for the repository use the PRs head vs main. Then `repo.full_name` pulls the full path of the repo the fork is on vs attempting to pull the branch name from our path where it wouldn't exist ([reference docs here](https://docs.github.com/en/webhooks/webhook-events-and-payloads#pullrequestevent)).

KunJeongPark: Thank you Jax!

PR #221, opened 2023-11-16
Vault with Purchase PayPalWeb

KunJeongPark:
### Summary of changes

- Created UI for vault option for PayPalWeb checkout in demo app
- Added experience_context and vault parameters with vault option selected in create order 
- Display vaultID and customerID in capture/authorize if PayPal payment is vaulted with checkout

https://github.com/paypal/paypal-ios/assets/9273272/2ec267d0-4b2d-487f-998d-a322cf9ff5ad

### Checklist

### Authors
- @KunJeongPark

### DIFF
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -63,6 +63,11 @@ struct Attributes: Encodable {
 
     let vault: Vault
     let customer: VaultCustomer?
+
+    init(vault: Vault, customer: VaultCustomer? = nil) {
+        self.vault = vault
+        self.customer = customer
+    }
 }
 
 struct Vault: Encodable {
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -62,9 +62,9 @@ struct VaultCard: Encodable {
 struct Attributes: Encodable {
 
     let vault: Vault
-    let customer: VaultCustomer?
+    let customer: Customer?
 
-    init(vault: Vault, customer: VaultCustomer? = nil) {
+    init(vault: Vault, customer: Customer? = nil) {
         self.vault = vault
         self.customer = customer
     }
@@ -76,12 +76,6 @@ struct Vault: Encodable {
     let usageType: String?
     let customerType: String?
 }
-
-struct VaultCustomer: Encodable {
-
-    let id: String
-}
-
 
 struct PurchaseUnit: Encodable {
 
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -3,7 +3,7 @@ struct CreateOrderParams: Encodable {
     let applicationContext: ApplicationContext?
     let intent: String
     var purchaseUnits: [PurchaseUnit]?
-    var paymentSource: VaultCardPaymentSource?
+    var paymentSource: VaultPaymentSource?
 }
 
 struct ApplicationContext: Codable {
@@ -15,6 +15,38 @@ struct ApplicationContext: Codable {
         case userAction
         case shippingPreference
     }
+}
+
+enum VaultPaymentSource: Encodable {
+    case card(VaultCardPaymentSource)
+    case paypal(VaultPayPalPaymentSource)
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.singleValueContainer()
+        switch self {
+        case .card(let cardSource):
+            try container.encode(cardSource)
+        case .paypal(let paypalSource):
+            try container.encode(paypalSource)
+        }
+    }
+}
+
+struct VaultPayPalPaymentSource: Encodable {
+
+    let paypal: VaultPayPal
+}
+
+struct VaultPayPal: Encodable {
+
+    let attributes: Attributes
+    let experienceContext: ExperienceContext
+}
+
+struct ExperienceContext: Encodable {
+
+    let returnURL: String
+    let cancelURL: String
 }
 
 struct VaultCardPaymentSource: Encodable {
@@ -30,15 +62,17 @@ struct VaultCard: Encodable {
 struct Attributes: Encodable {
 
     let vault: Vault
-    let customer: CardVaultCustomer?
+    let customer: VaultCustomer?
 }
 
 struct Vault: Encodable {
 
     let storeInVault: String
+    let usageType: String?
+    let customerType: String?
 }
 
-struct CardVaultCustomer: Encodable {
+struct VaultCustomer: Encodable {
 
     let id: String
 }
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -45,6 +45,7 @@ struct VaultPayPal: Encodable {
 
 struct ExperienceContext: Encodable {
 
+    // these fields are not encoded for our SDK but are required for create order with PayPal vault option
     let returnURL: String
     let cancelURL: String
 }
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -76,6 +76,12 @@ struct Vault: Encodable {
     let storeInVault: String
     let usageType: String?
     let customerType: String?
+
+    init(storeInVault: String, usageType: String? = nil, customerType: String? = nil) {
+        self.storeInVault = storeInVault
+        self.usageType = usageType
+        self.customerType = customerType
+    }
 }
 
 struct PurchaseUnit: Encodable {
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
@@ -5,7 +5,6 @@ struct CreateOrderPayPalWebView: View {
     @ObservedObject var paypalWebViewModel: PayPalWebViewModel
 
     @State private var selectedIntent: Intent = .authorize
-    @State private var vaultCustomerID: String = ""
     @State var shouldVaultSelected = false
 
     let selectedMerchantIntegration: MerchantIntegration
@@ -28,7 +27,6 @@ struct CreateOrderPayPalWebView: View {
                 Toggle("Should Vault with Purchase", isOn: $shouldVaultSelected)
                 Spacer()
             }
-            FloatingLabelTextField(placeholder: "Vault Customer ID (Optional)", text: $vaultCustomerID)
             ZStack {
                 Button("Create an Order") {
                     Task {
@@ -38,8 +36,7 @@ struct CreateOrderPayPalWebView: View {
                                 amount: "10.00",
                                 selectedMerchantIntegration: DemoSettings.merchantIntegration,
                                 intent: selectedIntent.rawValue,
-                                shouldVault: shouldVaultSelected,
-                                customerID: vaultCustomerID.isEmpty ? nil : vaultCustomerID
+                                shouldVault: shouldVaultSelected
                             )
                         } catch {
                             print("Error in getting setup token. \(error.localizedDescription)")
--- a/Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
@@ -5,6 +5,8 @@ struct CreateOrderPayPalWebView: View {
     @ObservedObject var paypalWebViewModel: PayPalWebViewModel
 
     @State private var selectedIntent: Intent = .authorize
+    @State private var vaultCustomerID: String = ""
+    @State var shouldVaultSelected = false
 
     let selectedMerchantIntegration: MerchantIntegration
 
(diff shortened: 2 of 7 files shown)


Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -10,15 +10,46 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+            var customer: VaultCustomer?
+            if let customerID {
+                customer = VaultCustomer(id: customerID)
+            }
+            let attributes = Attributes(
+                vault: Vault(
+                    storeInVault: "ON_SUCCESS",
+                    usageType: "MERCHANT",
+                    customerType: "CONSUMER"
+                ),
```
KunJeongPark: Just hard coded this for direct merchant but I am investigating what different parameters are for different selected merchant integrations.

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -30,15 +62,17 @@ struct VaultCard: Encodable {
 struct Attributes: Encodable {
 
     let vault: Vault
-    let customer: CardVaultCustomer?
+    let customer: VaultCustomer?
```
jaxdesmarais: I wonder if we can just call this `Customer` instead of `VaultCustomer`.

Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
```diff
@@ -5,6 +5,8 @@ struct CreateOrderPayPalWebView: View {
     @ObservedObject var paypalWebViewModel: PayPalWebViewModel
 
     @State private var selectedIntent: Intent = .authorize
+    @State private var vaultCustomerID: String = ""
```
jaxdesmarais: I thought you had said during stand that regardless of what's customer ID is passed in the API always generates one? If that's the case, I think we should remove this from our demo.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
```diff
@@ -39,6 +39,14 @@ struct PayPalWebOrderCompletionResultView: View {
                 LeadingText("Email", weight: .bold)
                 LeadingText("\(email)")
             }
+            if let vaultID = orderResponse.paymentSource?.paypal?.attributes?.vault.id {
+                LeadingText("Vault ID / Payment Token", weight: .bold)
+                LeadingText("\(vaultID)")
+            }
```
jaxdesmarais, line 45: General question: is there an agreed upon name that JS calls this in our public docs? I see we use both vault ID and payment token here. We should use whichever name is used in our public docs.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
jaxdesmarais: We should get input from product on what we need to support here for v1 in the demo app.


-Victoria : yes, Walter referenced docs during stand and I plan on working with Nirvan and Walter on the specs. 

So right now, in demo app we only have direct, connected and managed path configured, not specific paths like CP2 or CP4. The usageType would depend on specific integration patterns. 

I think specs for server side inputs like auth headers and payee info should be made for each integration types in another PR.

Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
KunJeongPark: Yes, my plan is to do that today.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: yes, Walter referenced docs during stand and I plan on working with Nirvan and Walter on the specs.

Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionResultView.swift
KunJeongPark: I think for the customer point of view, we call it payment token , but we reference it as vault ID in the APIs. 
I just referenced vault Id on the label to clarify that they are the same.

Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
KunJeongPark: Addressed in  commit dc0c32e4f431174aa5db43e4a7155b266a97f774 and f07d0bccd20bb2cde3f7f560affff71d5554ba8d

Demo/Demo/Models/CreateOrderParams.swift
KunJeongPark: Addressed here: 62480f60df4f9705c4092f02d9a983298e08833e

KunJeongPark:
62480f6 commit message should have been "reuse for Vault customer encoding"

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: So right now, in demo app we only have direct, connected and managed path configured, not specific paths like CP2 or CP4. The usageType would depend on specific integration patterns. 

I think specs for server side inputs like auth headers and payee info should be made for each integration types in another PR.

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -3,7 +3,7 @@ struct CreateOrderParams: Encodable {
     let applicationContext: ApplicationContext?
     let intent: String
     var purchaseUnits: [PurchaseUnit]?
-    var paymentSource: VaultCardPaymentSource?
+    var paymentSource: VaultPaymentSource?
```
scannillo, line 6: Just to clarify my understanding - a merchant only needs to set the `paymentSource` in their create order call if they want to vault, correct?

How do we document this to a merchant?

Demo/Demo/SwiftUIComponents/PayPalWebViews/CreateOrderPayPalWebView.swift
```diff
@@ -22,6 +23,10 @@ struct CreateOrderPayPalWebView: View {
                 Text("CAPTURE").tag(Intent.capture)
             }
             .pickerStyle(SegmentedPickerStyle())
+            HStack {
+                Toggle("Should Vault with Purchase", isOn: $shouldVaultSelected)
```
scannillo, line 27: This looks great!

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -10,15 +10,40 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+        var vaultPayPalPaymentSource: VaultPayPalPaymentSource?
+        if shouldVault {
+            let attributes = Attributes(
+                vault: Vault(
+                    storeInVault: "ON_SUCCESS",
+                    usageType: "MERCHANT",
+                    customerType: "CONSUMER"
+                )
+            )
+            let paypal = VaultPayPal(attributes: attributes, experienceContext: ExperienceContext(returnURL: "https://example.com/returnUrl", cancelURL: "https://example.com/cancelUrl"))
```
scannillo, line 28: In your video, the redirect back to the Demo app looked fine, even though you would think the experienceContext should be a URL scheme to the app. What happens if the merchant doesn't set these URLs?

KunJeongPark: The endpoint throws an error without the experienceContext field for PayPal vaulting.

Demo/Demo/Models/CreateOrderParams.swift
KunJeongPark: Yes, JS SDK has documentation on this on their developer site. We should do something similar.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: In our SDK, when we launch the ASWebSession for PayPalWeb, we construct the redirectURLString and append it as a query parameter to the checkout url : https://github.com/paypal/paypal-ios/blob/29307dec8c267d85aac89b19c433ee357c50b038/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift#L86
The redirect url is included as a parameter in the url that we construct to launch the webSession. 
I don't know if this is a typo or intentional but query parameter name for redirect url is strange.

So this experienceContext in create Order API is only relevant to JS SDK but the API fails nonetheless
without it.

scannillo: :sigh: That's right, this is an issue with the create-order API. It doesn't accept non-https URLs.

We should add a comment to make this issue clear in the code. That basically createOrder requires an https URL, which doesn't get used by the SDK flow at all. The merchant could put `https://bogus.com` and things will work fine.

KunJeongPark: Yes. That might be an actual site.

KunJeongPark: Added comments above to clarify that required experienceContext fields for PayPal vault with purchase are not used in the SDK.

Demo/Demo/ViewModels/CardPaymentViewModel.swift
```diff
@@ -21,15 +21,20 @@ class CardPaymentViewModel: ObservableObject, CardDelegate {
...
+        var vaultCardPaymentSource: VaultCardPaymentSource?
         if shouldVault {
-            var customer: CardVaultCustomer?
+            var customer: Customer?
             if let customerID {
-                customer = CardVaultCustomer(id: customerID)
+                customer = Customer(id: customerID)
             }
-            let attributes = Attributes(vault: Vault(storeInVault: "ON_SUCCESS"), customer: customer)
+            let attributes = Attributes(vault: Vault(storeInVault: "ON_SUCCESS", usageType: nil, customerType: nil), customer: customer)
```
jaxdesmarais: Can we default `usageType` and `customerType` to `nil` in the struct?

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -10,15 +10,41 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+
+        var vaultPayPalPaymentSource: VaultPayPalPaymentSource?
+        if shouldVault {
+            let attributes = Attributes(
+                vault: Vault(
+                    storeInVault: "ON_SUCCESS",
+                    usageType: "MERCHANT",
+                    customerType: "CONSUMER"
+                )
+            )
```
jaxdesmarais: nit: if we single line this are we still under our linter warning?

Demo/Demo/ViewModels/PayPalWebViewModel.swift
jaxdesmarais: I don't suppose there is a way for use to pass a URL on our merchants behalf so they don't need to do this step?

scannillo: The createOrder will just error if that URL value isn't set or is an empty string. It also errors if the string doesn't start with `https`. Victoria - can we make a JIRA for tracking taking up this issue with the API team?

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -10,15 +10,41 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+            // The returnURL is not used in our mobile SDK, but a required field for create order with PayPal payment source.
```
scannillo: Nitpick - I would also add a reference to the JIRA ticket # that summarizes this issue we need to take up with the API team!

scannillo (approved):

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: Roger

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: I'll try it. No warning and it's much more readable.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: Created - DTPPCPSDK-1492

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: addressed here: c77464c24fd783711665df8cbc5e07f6f0d09fd2

Demo/Demo/ViewModels/PayPalWebViewModel.swift
KunJeongPark: addressed here: c77464c24fd783711665df8cbc5e07f6f0d09fd2

Demo/Demo/ViewModels/CardPaymentViewModel.swift
KunJeongPark: addressed here: c77464c24fd783711665df8cbc5e07f6f0d09fd2

jaxdesmarais (approved):

PR #223, opened 2023-11-21
PayPal Web Demo App Refactor

jaxdesmarais:
### Reason for changes

There was a lot of logic that's currently duplicated across views or using the same type. This PR overall reduces the number of states used in the PayPal Demo app. It also includes refactors that can be used or duplicated across other demo apps. As necessary we can extract out portions of this code to be reused as we refactor the other portions of the demo app.

### Summary of changes

- Create `CurrentState` enum - this should be able to be shared in other demo features
- Rename `PayPalWebView` to `PayPalWebDemoView` to reduce confusion with using the term "web view" - this is the entry point to the feature
- Rename `PayPalTransactionView ` to `PayPalWebButtonsView` - this view contains the PayPal buttons so the name did not accurately represent the contents of the view
- Rename `CreateOrderPayPalWebView` to `PayPalWebCreateOrderView` to match other file naming conventions
- Removed unneeded `PayPalWebOrderCompletionView` and `PayPalWebState`
- Singular `PayPalWebResultView` can now be used vs the previous `PayPalWebApprovalResultView`, `OrderCreatePayPalWebResultView`, `PayPalWebOrderCompletionResultView`
- Created new `PayPalWebStatusView` that contains logic for what to display based on the status vs having this in multiple views
- Created new `PayPalWebTransactionView` which contains the view for authorizing/capturing an order (replaces `PayPalWebOrderActionButton`)
- `PayPalWebViewModel` changes
    - Moved logic into the ViewModel directly where views didn't need to control logic for certain pieces of code - dumb views down where possible
    - Update logic as needed to update state as expected
    - Cleanup method signatures where possible

### Future Considerations / Improvements
- For the most part aside from updating labels this PR keeps the UI 1:1 with what exists today - we should consider if we want to improve the UI further in a future update or if this is our desired flow
- For the PayPal Web flow do we need to display 3 buttons with 3 different funding sources or are we okay just demo-ing 1 button (on BT we have all 3 so may make sense to leave as is)
- There are likely pieces of code that can be extracted out and shared among other features such as buttons - most of them add the same attributes so we can consider this in a future refactor
- Once all features are complete we should clean up the file structure of each feature and ensure the demo app is well organized and easy to navigate
- There are some classes that can be removed entirely like `CustomButton` and `CustomTestField` these are UIKit Views and no longer used at all in the Demo app - there are likely other things we can also consider cleaning up

### Next Steps
- Pick the next view to tackle!

### Checklist

### Authors

- @jaxdesmarais


Demo/Demo/Models/Order.swift
```diff
@@ -2,8 +2,8 @@ struct Order: Codable, Equatable {
     
     let id: String
     let status: String
-    var paymentSource: PaymentSource?
-    
+    let paymentSource: PaymentSource?
```
jaxdesmarais, line 5: This isn't related to this PR but at one point I was poking around at this file and noticed we had this one as a variable - should be a constant to match all other structs in this file

Demo/Demo/SwiftUIComponents/CurrentState.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+
+enum CurrentState: Equatable {
```
jaxdesmarais, line 3: Ideally we can use this across the other demo features as well vs the individual states per feature that exist today

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
```diff
@@ -38,10 +38,10 @@ struct FeatureSelectionView: View {
                         Text("Card Vaulting")
                     }
                     NavigationLink {
-                        PayPalWebView()
-                            .navigationTitle("PayPalWeb Payment")
+                        PayPalWebDemoView()
```
jaxdesmarais: The previous name containing webview could be confusing - updated the entry point to the feature to be called "demo view" but open to other names if folks prefer something else

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebButtonsView.swift
```diff
@@ -1,24 +1,23 @@
 import SwiftUI
 import PaymentButtons
 
-struct PayPalTransactionView: View {
+struct PayPalWebButtonsView: View {
```
jaxdesmarais: Renamed from "TransactionView" to "ButtonView" since this contains all of the payment buttons and does not do anything with transactions

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebCreateOrderView.swift
```diff
@@ -1,13 +1,11 @@
 import SwiftUI
 
-struct CreateOrderPayPalWebView: View {
+struct PayPalWebCreateOrderView: View {
```
jaxdesmarais: We may be able to reuse some sort of create order view across all of the features - left separate for now but could be something to consider as we continue this refactor

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebResultView.swift
```diff
@@ -0,0 +1,25 @@
+import SwiftUI
+
+enum Status {
```
jaxdesmarais: We may be able to extract/reuse this across the other features as well - left here for now but I expect many of the other features may have a similar status. If not we can rename this to `PayPalWebStatus`

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
```diff
@@ -0,0 +1,64 @@
+import SwiftUI
+
+struct PayPalWebStatusView: View {
```
jaxdesmarais: We can also consider if we want different data displayed for each status - this PR matched what was previously displayed 1:1 but we may be able to reduce this down to a singular status view with the same data displayed across all portions of the flow.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -4,60 +4,62 @@ import PayPalWebPayments
...
     var payPalWebCheckoutClient: PayPalWebCheckoutClient?
 
     let configManager = CoreConfigManager(domain: "PayPalWeb Payments")
 
-    func createOrder(amount: String, selectedMerchantIntegration: MerchantIntegration, intent: String) async throws {
-        // might need to pass in payee as payee object or as auth header
-
-        let amountRequest = Amount(currencyCode: "USD", value: amount)
+    func createOrder() async throws {
+        let amountRequest = Amount(currencyCode: "USD", value: "10.00")
```
jaxdesmarais: We were hardcoding this amount in a view so just hardcoding it here instead - if we add an amount text box in the future we can update this method signature to take in an amount

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -70,64 +72,54 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
-        DispatchQueue.main.async {
-            self.state.checkoutResultResponse = .loaded(checkoutResult)
-        }
-    }
-
-    func paypalWebCheckoutFailureResult(checkoutError: CorePayments.CoreSDKError) {
-        DispatchQueue.main.async {
-            self.state.checkoutResultResponse = .error(message: checkoutError.localizedDescription)
-        }
-    }
```
jaxdesmarais, line 103: These methods are only doing 1 thing so moved this logic into the delegates directly vs calling a method

Sources/PayPalWebPayments/PayPalWebCheckoutFundingSource.swift
```diff
@@ -6,6 +6,7 @@ public enum PayPalWebCheckoutFundingSource: String {
     /// Eligible costumers receive a revolving line of credit that they can use to pay over time.
     case paypalCredit = "credit"
 
+    // NEXT_MAJOR_VERSION: rename to `payLater`
```
jaxdesmarais, line 9: Noticed we have the casing weird here so added this next major version note

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -53,7 +53,21 @@ final class DemoMerchantAPI {
             throw error
         }
     }
-    
+
+    func completeOrder(intent: Intent, orderID: String) async throws -> Order {
```
jaxdesmarais, line 57: When we refactor card we should use this singular method - the only difference between authorize and capture is the intent so we don't need 2 separate methods for that

sshropshire (approved):
🌊

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -70,83 +72,44 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
-            )
-            DispatchQueue.main.async {
-                self.state.authorizedOrderResponse = .loaded(order)
+            if let orderID = order?.id {
+                let order = try await DemoMerchantAPI.sharedService.completeOrder(intent: intent, orderID: orderID)
+                
+                DispatchQueue.main.async {
+                    self.order = order
+                    self.state = .success
+                }
```
sshropshire: random: I wonder if one day we'll be able to remove calls to `DispatchQueue.main.async`. If we can find a cleaner way to update the UI without having to explicitly dispatch to the main thread, we can DRY up the code and make it more readable.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
KunJeongPark, line 1: 🤯

KunJeongPark: Now trying to understand the distinction between .loaded and .success. I just understood .loaded as representing success.
 - I see now.
    loaded => order has been created, .success => order authorized/captured
    and check for approval in SDK by presence of checkoutResult value.

Demo/Demo/ViewModels/PayPalWebViewModel.swift
jaxdesmarais: I believe the work with `@MainActor` would handle this as it would call these updates from the main thread always. Specifically we can make the entire view model a `MainActor` and remove all calls to `DispatchQueue.main` in this class. Makes sense to leave for now but that workstream should greatly clean up all the main queue calls here.

sshropshire: Nice yeah that's sort of what I was thinking. It's out of scope for this PR but it'd be nice to establish that pattern one day to clean everything up.

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -33,12 +31,9 @@
 		3BA56FF82A9FDB5A0081D14F /* CardPaymentOrderCompletionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BA56FF72A9FDB5A0081D14F /* CardPaymentOrderCompletionView.swift */; };
 		3BA56FFA2A9FE4180081D14F /* CardOrderCompletionResultView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BA56FF92A9FE4180081D14F /* CardOrderCompletionResultView.swift */; };
 		3BA56FFC2A9FEFE90081D14F /* PayPalWebViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3BA56FFB2A9FEFE90081D14F /* PayPalWebViewModel.swift */; };
```
scannillo, line 33: Could be for another PR, but could we rename `PayPalViewModel.swift` --> `PayPalNativeViewModel.swift` to follow the same convention as the Web one? Or is there a plan to further consolidate the 2?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebCreateOrderView.swift
scannillo: Agreed!

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebResultView.swift
scannillo: Does this represent the status of the orderID? Maybe we can make the name more descriptive

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebResultView.swift
```diff
@@ -0,0 +1,25 @@
...
+struct PayPalWebResultView: View {
+
+    @ObservedObject var payPalWebViewModel: PayPalWebViewModel
+
+    var status: Status
+
+    var body: some View {
+        switch payPalWebViewModel.state {
+        case .idle, .loading:
+            EmptyView()
```
scannillo: Would it be better to show some text like "Loading ..." onscreen vs empty?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
scannillo: Similar Q - maybe more clear to name this something along the lines of "OrderID" status view. We could also add docstrings to the class-level to make things more clear. Also agree this should def be re-used across payment methods at some point

Demo/Demo/ViewModels/PayPalWebViewModel.swift
```diff
@@ -70,83 +72,44 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
 
     func payPal(
-        _ payPalClient: PayPalWebPayments.PayPalWebCheckoutClient,
-        didFinishWithResult result: PayPalWebPayments.PayPalWebCheckoutResult
+        _ payPalClient: PayPalWebCheckoutClient,
+        didFinishWithResult result: PayPalWebCheckoutResult
     ) {
-        paypalWebCheckoutSuccessResult(checkoutResult: PayPalWebState.CheckoutResult(id: result.orderID, payerID: result.payerID))
+        checkoutResult = result
     }
```
KunJeongPark: oh 🤣

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
```diff
@@ -0,0 +1,64 @@
...
+                HStack {
+                    Text("Order Approved")
+                        .font(.system(size: 20))
+                    Spacer()
+                }
+                if let order = payPalViewModel.order {
+                    LeadingText("Intent", weight: .bold)
+                    LeadingText("\(payPalViewModel.intent)")
+                    LeadingText("Order ID", weight: .bold)
+                    LeadingText("\(order.id)")
```
scannillo: These 4 lines should apply the same to each case, right? Maybe we can pull the intent & orderID part out of the switch statement

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
```diff
@@ -0,0 +1,64 @@
+import SwiftUI
+
+struct PayPalWebStatusView: View {
+
+    var status: Status
+    var payPalViewModel: PayPalWebViewModel
```
scannillo: Why do we need a `Status` in addition to a ViewModel?

KunJeongPark: I think the events are linear and looking at presence of value in event 2 (for instance, order) in viewModel could mean event 2, 3 or 4, so I think that value is needed from the asking container view to tell status view what to display.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebResultView.swift
jaxdesmarais: Agreed - renamed to `OrderStatus` here to be more descriptive: https://github.com/paypal/paypal-ios/pull/223/commits/6c304140b0fa866b668711c44eb6c7411b5a437a

Demo/Demo/ViewModels/PayPalWebViewModel.swift
jaxdesmarais: I did move these updates into their own methods here to keep the calls to the dispatch queue in 1 place, which cleans it up a bit more for now: https://github.com/paypal/paypal-ios/pull/223/commits/a7b9b64a46f2d4e2c29f8950f676fa806711ab6f

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: Nah, I agree it's a bit confusing. I updated it here to combine it into a singular `.success` state. It just involves resetting the state for each view which I think is a more accurate representation of the expected behavior. Update here: https://github.com/paypal/paypal-ios/pull/223/commits/a7b9b64a46f2d4e2c29f8950f676fa806711ab6f

Demo/Demo.xcodeproj/project.pbxproj
jaxdesmarais: Yeah, I was leaving that for a follow up PR when we do those files as a whole. I _think_ they will need to be separate but agree that one should be renamed to PayPalNative. 

Right now we also have a separate directory for the Views vs the ViewModels. Do we like that pattern or should we move each features ViewModel into it's top level directory? It could be nice for merchants to just look in 1 place for everything related to a feature vs traversing multiple directories.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebResultView.swift
jaxdesmarais: So we _do_ add a loading indicator to the button when the state is loading ([example here](https://github.com/paypal/paypal-ios/blob/6c304140b0fa866b668711c44eb6c7411b5a437a/Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift#L24-L26)). But I don't think we want the entire view to show the loading indicator which is why we set the View to EmptyView here. Once the state is updated from loading we display the entire result view.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: What do we think of `PayPalWebOrderStatusView`?

jaxdesmarais: What do we think of `PayPalWebOrderStatusView`?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: They don't apply to all of them, each view has the following:
- Order Started: order created header with order ID and status
- Order Approved: order approved header with intent, order ID, and payer ID
- Order Completed: order intent header with order ID, status, and optional email, customer ID, vault ID

I posed the [question here](https://github.com/paypal/paypal-ios/pull/223#discussion_r1400899581) of do we need different data for different states. We could consider displaying the same data for all views and adding additional data in later states.

I am not totally sure why we need to display the status for created and the intent + payer ID for approved. Do we find this data useful? If not we can still update the headers but display a more consistent set of data. What do you think?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: It's used for the switch below

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebButtonsView.swift
```diff
@@ -29,17 +28,22 @@ struct PayPalTransactionView: View {
                     .stroke(.gray, lineWidth: 2)
                     .padding(5)
             )
-            PayPalWebApprovalResultView(paypalWebViewModel: paypalWebViewModel)
-            if paypalWebViewModel.state.checkoutResult != nil {
+
```
KunJeongPark: what would be a case where it's success and result is nil? Something went wrong on API side?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift
```diff
@@ -0,0 +1,39 @@
...
+                        CircularProgressView()
+                    }
+                }
+                
+                if payPalWebViewModel.state == .success {
+                    PayPalWebResultView(payPalWebViewModel: payPalWebViewModel, status: .completed)
+                }
+            }
+            .onAppear {
+                payPalWebViewModel.resetState()
```
KunJeongPark: One behavior change is that if I navigate back from this view to original screen, I have this weird look. I think you can just check a condition for that checkout button.
That's one thing about the view resetting state in the vm on appear. I am thinking about other modules where there are many steps and I have container views that contain button,  input fields and result view. And if I go back and forth, the views with results would be wiped out. 

I have done this in my testing so I can check the results in previous screen. There is also different setup in card vault without purchase flow in demo app where I just display results in a large scrollview and scroll to bottom as results come up. So I think I am reliant on viewModel keeping track of state. I guess then I can always NOT reset the state.

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
sshropshire: ✌️cents: On Android we do lean toward `<FEATURE_NAME>View` because the Demo app is basically a contrived sample merchant integration for us to debug and QA the SDK. The fundamental idea is to create a Demoable "view" of each feature. The `FeatureName` prefix also scales nicely as a naming convention in other parts of the project.

Also, I believe the web team has a separate sample app with fictitious products to serve as a more complete onboarding example for merchants. Not super related, but worth mentioning.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift
jaxdesmarais: Easy, we can rely on the status instead of resetting the state: https://github.com/paypal/paypal-ios/pull/223/commits/94f8f9b51a6e5e640a238996751ddadbcac0134d. If you just want to triple check me but the behavior looks to be the same as before.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
sshropshire: Same thing here, since it's technically a view of the `PayPalWebResult` we go with `PayPalWebResultView` on Android. A merchant typically would have their own `OrderCompleteView` style naming convention, but since our Demo apps are a glorified console log, I feel like it makes sense to couple the view to the underlying type it displays.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift
KunJeongPark: Thank you for fast fixes! I just want to make sure that I have complete understanding for following this pattern for other modules.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebButtonsView.swift
jaxdesmarais: Yeah, we update the `checkoutResult` based on what we get back in the `didFinishWithResult` delegate method, so if we hit another delegate we would not have a `checkoutResult` and thus not be allowed to move to the `PayPalWebTransactionView`

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
jaxdesmarais: Typically I'd agree with that convention but it's a bit more convoluted imo here since `webview` is already a separate thing on iOS. We could use the full module name as our convention so `PayPalWebPaymentView` if we like that better. What do you think?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: We already have a `PayPalWebResultView` and since these views are conditionally displayed based on the status I think it'd be _more_ confusing to move the status view logic into here since it'd bloat this file quite a bit. Now if we simplify this view and display the same data across all status' ([as I mentioned here](https://github.com/paypal/paypal-ios/pull/223#discussion_r1412247314)) I think we could move this logic into our result view.

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
sshropshire: Oh yeah total collision I see that went over my head completely lol. We still call it `PayPalWebView` on the Android side, but yeah I could see how that makes it weird. I'm indifferent I like the brevity of it, but at the same time I see the utility of adding "Demo."

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebButtonsView.swift
KunJeongPark: oh I c,  success state is used for different results.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
sshropshire: Got it yeah, looking at ours we may need to clean up some of the naming. We go with `PayPalWebCheckoutResultView` to display the fields in a `PayPalWebCheckoutResult`. We do have some bloat on the Android side, probably a side effect of the UX. Another option would be to split everything up into screens to reduce the conditional logic, but that's a separate PR for sure.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: > I agree, the results should correlate with the steps that are performed.

Is there a reason we only need status for order started and completed? Same with intent and payer ID only for approved?

Imo I don't think we need status displayed at all. Instead we could have: order ID, and intent for all views. Then we could append payer ID with approved and append email, customer ID and vault ID with completed. What do you think?

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift
KunJeongPark: I am missing scroll to bottom behavior that I had with vaulting option selected where the result view is longer.
I used this scrolling to bottom with view model's state change a lot in vault without purchase for card payment.
just give me some think time on things being tracked outside of vm.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebStatusView.swift
jaxdesmarais: Actually I take that back - I think we do want different views here otherwise we end up updating all views based on the latest order data. For now I think it makes sense to have some data tied to status to avoid updating the earlier views on other screens.

Demo/Demo/SwiftUIComponents/PayPalWeb/PayPalWebTransactionView.swift
jaxdesmarais: We can rely on the change to order (updated with the call to complete transaction) to auto scroll to the bottom. Updated here: https://github.com/paypal/paypal-ios/pull/223/commits/c2785ae8e63b361b7052c57590639683cd54a3a8

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebTransactionView.swift
```diff
@@ -0,0 +1,44 @@
...
+                        .buttonStyle(RoundedBlueButtonStyle())
+                        .padding()
+
+                        if payPalWebViewModel.state == .loading {
+                            CircularProgressView()
+                        }
+                    }
+
+                    if payPalWebViewModel.state == .success && payPalWebViewModel.order?.status == "COMPLETED" {
+                        PayPalWebResultView(payPalWebViewModel: payPalWebViewModel, status: .completed)
```
KunJeongPark, line 31: I am thinking of cases where order status COMPLETED  (APPROVED is an example) is not the end state that is not an error in some of flows after capture/authorize. If we know if it's only APPROVED or COMPLETED after authorized and capture, it's fine. Looking in Ppass

jaxdesmarais: According to PPaaS "COMPLETED" is always the result of a successful authorize/capture attempt which is what is being displayed here - the result of a completed call to auth/capture

KunJeongPark: So as we were discussing, in some vaulting flows, it was helpful to surface unexpected "APPROVED" state after authorize/capture. It seems a departure from checking for presence of order after order creation, checking for checkoutResult for approval and here checking for value of order status being "COMPLETED".

jaxdesmarais: I think in those cases we can do `viewModel.order?.status == "COMPLETED" || viewModel.order?.status == "APPROVED"`. From the testing I've done thus far we always get `COMPLETED` back from PayPal Web. We won't want to display the result of a successfully authorized/captured transaction for the other status' returned from PPaaS: saved, pending approval, voided, payer action required, and partially completed.

This conditional is specifically for displaying the result of a successful authorize or capture API call, which we only want to display with specific status'.

KunJeongPark: but you know Jax, that situation made me sus. It was not a few in that situation. 
Success case should be some value but we got something else. 
And in that case we need to surface that. This was my original reason for having separate variables of states for each stage of API results, to be able to track them in the vm.

Demo/Demo.xcodeproj/project.pbxproj
scannillo: I like that idea - breaking it up by feature instead. Things are a little hard to find ATM

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebTransactionView.swift
jaxdesmarais: > but you know Jax, that situation made me sus. It was not a few in that situation.

I am not totally sure I follow, but were there cases where the call to authorize/capture returned one of: saved, pending approval, voided, payer action required, or partially completed but was actually successful? If so that doesn't match the description of those status on the authorize/capture API call according to PPaaS, those status all would indicate the transaction was not completed which we would not display the final completed status view for.

> Success case should be some value but we got something else.

The success state is set by us as a result of the `completeOrder` API call. We are checking for both a success state as well as the `order.status` raw value string from the API.

> And in that case we need to surface that.

If we want to surface the other status that should _not_ be in the conditional here. Instead we would want a separate conditional for the other status results. Just want to make sure I am understand what you are saying correctly though?

> This was my original reason for having separate variables of states for each stage of API results, to be able to track them in the vm.

We can still track these! The previous state variables were all being set to a generic order which has now been combined into a singular order. The logic is the exact same (result of each call in the view model updates the order) so we have the exact same information we did previously!

Demo/Demo.xcodeproj/project.pbxproj
jaxdesmarais: Agreed! Moved: https://github.com/paypal/paypal-ios/pull/223/commits/5a957651b375af45307874a19ff23a8b900bcc8b

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebTransactionView.swift
KunJeongPark: yes it was success and returned Approved after authorize/capture. 
So I think previously, we could track loaded(order) for authorized/capture variable.

I think now, after authorize/capture, we look for success state and order status value being  "COMPLETE".
I think you can just take out the order status value "COMPLETE" checking for success case for authorize/capture API call. 

I am a bit uncomfortable with there not being way to track this specific success state of different API calls.

I think work you did is great. Looking at it, it's so clear, organized and easier to follow.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebPaymentsView.swift
```diff
@@ -0,0 +1,25 @@
...
+
+    @StateObject var payPalWebViewModel = PayPalWebViewModel()
+
+    var body: some View {
+        ScrollView {
+            VStack(spacing: 16) {
+                PayPalWebCreateOrderView(payPalWebViewModel: payPalWebViewModel)
+                if payPalWebViewModel.order != nil {
+                    PayPalWebResultView(payPalWebViewModel: payPalWebViewModel, status: .created)
+                    NavigationLink {
```
KunJeongPark, line 13: If there is an error in creating order, the error view would not be displayed if we check for nil value for order here. I think the same is true for checks for `PayPalWebButtonsView` and `PayPalWebTransactionsView`. But removing this condition is problematic in the `PayPalWebButtonsView` because without this condition, it just creates an order approval view without the call. This is because it cannot distinguish success states of different API calls without checking existence published values returned. 

What do you think of keeping the file renaming, and consolidating result views but reverting to having three published state variables? I think that pattern might be less confusing and more flexible for different types of view setups also.

KunJeongPark: What do you think of this solution? It is a small change from what you have now. Instead of single state,
I have createOrderState, approveOrderState, authorizeCaptureOrderState of CurrentState type initialized to .idle.
I check for these states' success on the buttons to next step. In PayPalWebResultView, I toggle the state based on status. I kept your idea of using status to tell generic view which result to output.

struct PayPalWebResultView: View {

    @ObservedObject var payPalWebViewModel: PayPalWebViewModel

   ```
 var status: OrderStatus
    var state: CurrentState {
        switch status {
        case .approved:
            return payPalWebViewModel.approveOrderState
        case .created:
            return payPalWebViewModel.createOrderState
        case .completed:
            return payPalWebViewModel.authorizeCpatureOrderState
        }
    }

    var body: some View {
        switch state {
        case .idle, .loading:
            EmptyView()
```
Of course, I had to update the "update" functions in the viewModel.
I tested it for both non-vault and vault options for authorize, capture. I put up a branch ( https://github.com/paypal/paypal-ios/commit/85de4d9488928c8a746fa19f4685f88c1bb05e42 )

jaxdesmarais: > If there is an error in creating order, the error view would not be displayed if we check for nil value for order here.  I think the same is true for checks for PayPalWebButtonsView and PayPalWebTransactionsView. 

This is a bug on main which is why it was not addressed in this PR. We can test this on main by hardcoding a bad value for intent in `CreateOrderParams` and passing a bad order ID string to authorize/capture. Since the bug existed on main and the intent of this PR was to retain what existed today I had planned to address it in a follow up PR but we can also address it here.

KunJeongPark: Let me check but I do recall seeing the error view in main. Logic was to guard next steps with success state from API call but show result view for all cases (empty if loading or initial) and result view shows results for error or success. I tested it by returning a url error from the API call but I will check the other issue. If there were, thank you for addressing them. The error I was pointing out was the fact that you guarded the result view with success or value from API and that prevented errorViews from being surfaced. 

I do think separation of different states is good! Thank you for adding that.

jaxdesmarais: I've gone ahead and fixed the bug that existed on main with ErrorView, additionally added different results which I think is a _slightly_ better way then handing separate states in the latest 2 commits. Personally I feel that this amount of switching of handling views points out a flaw in our demo app. While this PRs intent was to simplify the views I think we can take a second pass at further cleaning up the demo app for this feature before moving on to the next one. 

Specifically we should have this all on 1 view (similar to card) with 1 result that is appended to as we have more data (vs the separate status views that exist on main/in this PR). I think another couple of refactors that we can consider is a picker view for the buttons (similar to auth/capture when creating an order) as well as 1 button that changes as we progress from create order > button view > transaction view. Additionally we don't have a way to reset PayPal web so running the demo multiple times become difficult as we never reset the views.

I think this PR can/should be the first of a few progressive refactors before we move onto other feature! 🚀

KunJeongPark: I did have results appended in Card Vault without purchase and that made it easier to have "reset" button on bottom of scrollview which is much more convenient than to have to navigate back through container views for each API result views to main feature selection view to reset. I think separate container views are easier to view but long scrollView is easier to "reset".

With different states for each API results, that makes things easier.

And I never thought about the bad value for intent being possible.

jaxdesmarais: > The error I was pointing out was the fact that you guarded the result view with success or value from API and that prevented errorViews from being surfaced.

Yep this was the same on main as I said ([code here](https://github.com/paypal/paypal-ios/blob/ae1adc3788995995adb677f39826d49bbdd3ffeb/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebView.swift#L15)) - we never handled the else so if an order wasn't created we just console logged the error but never displayed the error view.

> With different states for each API results, that makes things easier.

We should be able to remove the need for different states entirely with moving to a singular success view that's appended to as we get additional data. This would be more in line with the Braintree demo app as well (we actually display way less data there and if we need additional information for troubleshooting rely on breakpoints) - I think there is a good middle ground between the two though.

KunJeongPark: ooh, that was an oversight just for the order creation result. I meant to do what I did in TransactionView (https://github.com/paypal/paypal-ios/blob/ae1adc3788995995adb677f39826d49bbdd3ffeb/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalTransactionView.swift#L32)  which was to guard only the next step views. Thank you for correcting the create order error view issue.

KunJeongPark: I'm in favor of different states for different API's but I am ok with single state. I will adjust the current pattern as needed in other implementations.

jaxdesmarais: No problem! It was the same on the final view as well (we only previously handled if auth/capture was not nil - [here](https://github.com/paypal/paypal-ios/blob/ae1adc3788995995adb677f39826d49bbdd3ffeb/Demo/Demo/SwiftUIComponents/PayPalWebViews/PayPalWebOrderCompletionView.swift#L23) - can be tested by hardcoding the string "orderID" in the auth/capture call).

KunJeongPark: Yes, my intention was what I showed above in PayPalTransactionView. Well, I did catch my own errors in your PR. I think with different states, you don't need separate handling of error views.

KunJeongPark:
Adding the `TransactionOrder` in viewModel on auth/capture order works as well to distinguish the last  step. 
Great job renaming the files, and consolidating the views!

KunJeongPark (approved):

PR #225, opened 2023-11-30
Vault Without Purchase PayPal

KunJeongPark:
### Summary of changes

- Change `PaymentSource` in demo app to allow `SetUpToken` creation for both card and PayPal
- Implement parent class `VaultViewModel` inherited by `CardVaultViewModel` and `PayPalVaultViewModel`
  to reuse common properties and views.
- Rename `CardVaultState` to `VaultState` to define all common properties, move payment specific properties as 
   published properties in their respective view models
- Implement getting url from `SetUpToken` request 
-  Create UI for PayPal Vault without Purchase
- Implement `vault(url:)` function to launch `ASWebSession`
- Parse results from` ASWebSession`
- Define `PayPalVaultDelegate` protocol, methods. Add `PayPalVaultDelegate` protocol conformance to `PayPalVaultViewModel` and implement methods in `PayPalVaultViewModel`
- Update states for UI with vault method.
- Handle success, failure results from `vault(url:)` in SDK through `PayPalWebCheckoutClient`'s vault delegate functions
- Merchant server side `PaymentToken` call and Demo app output 
- Unit tests for `vault(url:)` function

https://github.com/paypal/paypal-ios/assets/9273272/496553c9-efd3-44fe-be0d-e699853c80a8

### Checklist

 - [x] Added a changelog entry

### Authors
@KunJeongPark


Demo/Demo/ViewModels/VaultState.swift
```diff
@@ -12,6 +12,7 @@ struct CardVaultState: Equatable {
     var setupToken: SetUpTokenResponse?
     var updateSetupToken: UpdateSetupTokenResult?
     var paymentToken: PaymentTokenResponse?
+    var paypalVaultToken: String?
```
KunJeongPark, line 19: I am wondering if I should add cancel state in LoadingState. 
I do provide delegate function for cancel so merchant can handle it the way they want
but I didn't include it as a separate state in the demo app because it didn't make sense for me to
display anything for canceling.

Demo/Demo/ViewModels/VaultState.swift
```diff
@@ -21,6 +22,14 @@ struct CardVaultState: Equatable {
         }
     }
 
+    var paypalVaultTokenResponse: LoadingState<String> = .idle {
+        didSet {
```
KunJeongPark, line 29: I think I prefer the separate variables of loading states for each API response.
This way, I can have a result view to toggle errorView and successView.
I prefer separated result views here instead of toggling generic one with different options from the 
container view, especially with just a few different result views here that are shared across different payment methods.

KunJeongPark:
I've been thinking about the approach to minimize the number of views in the demo app as suggested in previous PR's such as suggestion to use general view that toggles content for each API result and approach that I explored above, which was to use parent viewModel between two payment methods for vaulting without payment so that   `SetupTokenView` and `PaymentTokenView` and their result views can be reused across different payment methods when applicable.

I considered other payment methods being added as stand alone vault features in this approach to reducing the number of views by having a parent VaultViewModel class with shared state and payment specific properties as published properties in their respective viewModels. In JS, so far we have PayPal and Card vaulting and in Braintree, we have PayPal, Card and Venmo vaulting if I remember correctly.

Assuming that purpose of our demo app is for internal testing, demonstration of features and giving example usage of SDK to merchants, I think simplicity might be the best approach. 

We can adapt as we add more payment methods to vault. As far as I know, the flow for vault without pay is the same: create setupToken, do approval in SDK with payment method and merchants use this temporary setupToken to create paymentToken.

Demo/Demo/ViewModels/PayPalVaultViewModel.swift
```diff
@@ -0,0 +1,66 @@
+import UIKit
+import PayPalWebPayments
+import CorePayments
+
+class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
+
+    @Published var paypalVaultToken: PayPalVaultResult?
+
```
KunJeongPark: I considered having same `VaultResult` for card and PayPal vaulting but `UpdateSetupToken` endpoint
for card vaulting returns a lot more values than url string from PayPal approval web session, so I opted to keep them separate in case there is request in future to return more values to merchant.

If only minimum values that I display here are needed, they can be consolidated into a single `VaultResult` and `VaultView` and included in the `VaultState`.

Demo/Demo/SwiftUIComponents/VaultViews/SetupTokenResultView.swift
```diff
@@ -28,6 +28,10 @@ struct SetupTokenResultView: View {
             LeadingText("\(setupTokenResponse.customer?.id ?? "")")
             LeadingText("Status", weight: .bold)
             LeadingText("\(setupTokenResponse.status)")
+            if let url = setupTokenResponse.paypalURL {
+                LeadingText("PayPalURL", weight: .bold)
+                LeadingText("\(setupTokenResponse.paypalURL ?? "No url")")
```
richherrera:
```suggestion
                LeadingText(url)
```

Demo/Demo/ViewModels/VaultState.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 import CardPayments
+import PayPalWebPayments
 
-struct CardVaultState: Equatable {
+struct VaultState: Equatable {
 
     struct UpdateSetupTokenResult: Decodable, Equatable {
```
richherrera: This structure is duplicated on `CardVaultViewModel`. I see that we don't have any rules for handling nested structures in the States structures. For now, I think we should just define which one to work with

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
```diff
@@ -12,21 +12,22 @@ struct CardVaultView: View {
                 VStack(spacing: 16) {
                     CreateSetupTokenView(
                         selectedMerchantIntegration: DemoSettings.merchantIntegration,
-                        cardVaultViewModel: cardVaultViewModel
+                        vaultViewModel: cardVaultViewModel,
+                        paymentSourceType: PaymentSourceType.card
                     )
-                    SetupTokenResultView(cardVaultViewModel: cardVaultViewModel)
+                    SetupTokenResultView(vaultViewModel: cardVaultViewModel)
```
richherrera, line 18: nitpick: How about we just use `viewModel` instead of specifying which viewModel it is (cardVaultViewModel or vaulViewModel)? It's most convenient for a view to have only one ViewModel, so it's not really necessary to specify the type of ViewModel in the parameter

```suggestion
                    SetupTokenResultView(viewModel: cardVaultViewModel)
```

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
```diff
@@ -24,6 +24,104 @@ class PayPalClient_Tests: XCTestCase {
...
+
+        let url = URL(string: "https://sandbox.paypal.com/vault")
+        let expectedTokenIDResult = "fakeTokenID"
+        let expectedSessionIDResult = "fakeSessionID"
+        let mockVaultDelegate = MockPayPalVaultDelegate(success: {_, result in
+            XCTAssertEqual(expectedTokenIDResult, result.tokenID)
+            XCTAssertEqual(expectedSessionIDResult, result.approvalSessionID)
+            expectation.fulfill()
+        }, error: {_, _ in
+            XCTFail("Invoked error() callback. Should invoke success().")
```
richherrera, line 41: I just noticed that In some tests, we are omitting the fulfillment of the expectation :/

Demo/Demo/ViewModels/PayPalVaultViewModel.swift
```diff
@@ -0,0 +1,66 @@
+import UIKit
+import PayPalWebPayments
+import CorePayments
+
+class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
+
+    @Published var paypalVaultToken: PayPalVaultResult?
```
richherrera: nitpick: I think renaming it to `paypalVaultResult` would be the most convenient option. This name was from before the refactor. What do you think?

```suggestion
    @Published var paypalVaultResult: PayPalVaultResult?
```

Demo/Demo/ViewModels/VaultState.swift
richherrera: I think it depends on the Merchant's needs. For now, we can leave it as is; It LGTM. In fact, if you'd like, we can even consider the approach used in the `func cardDidCancel(_ cardClient: CardPayments.CardClient)` within the `CardPaymentViewModel`, where we use the .idle state.

Demo/Demo/ViewModels/PayPalVaultViewModel.swift
richherrera: From my perspective, communicating the viewModel with the view without using the `State` structure somewhat breaks the paradigm. For now, I think it's better to move PayPalVaultViewModel.`paypalVaultTokenResponse` and CardVaultViewModel.`updateSetupTokenResponse` back to `VaultState`, as it was in previous commits. Maybe someone else on the team could provide some input. Considering that the purpose of the Demo app is all about internal testing, showcase functionalities,  and providing merchants with examples of SDK usage

KunJeongPark: Yes, I thought I might get comment that it's confusing to be able to access updateSetupToken from PayPalVaultViewModel but I prefer them all in vaultState as well.

Demo/Demo/SwiftUIComponents/VaultViews/SetupTokenResultView.swift
KunJeongPark: Thank you for catching that.

Demo/Demo/ViewModels/VaultState.swift
KunJeongPark: that was a mistake leaving it there. Good catch.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: that's true, we are only dealing with vaultViewModels here, so it's redundant.

Demo/Demo/ViewModels/PayPalVaultViewModel.swift
KunJeongPark: I remember why I used Token as the name: to be consistent with updateSetupToken of UpdateSetupTokenResult type for card vaulting. In the same vault state, especially, I wanted the to have same naming convention

Demo/Demo/Models/PaymentTokenResponse.swift
```diff
@@ -14,7 +14,8 @@ struct Customer: Codable, Equatable {
 
 struct PaymentSource: Decodable, Equatable {
     
-    let card: BasicCard
+    var card: BasicCard?
+    var paypal: BasicPayPal?
 }
 
 struct BasicCard: Decodable, Equatable {
```
scannillo: What does BasicCard and BasicPayPal mean?

Demo/Demo/Models/SetupTokenRequest.swift
```diff
@@ -17,13 +23,24 @@ struct SetUpTokenRequest {
...
+                ]
             ]
+        }
+
+        let requestBody: [String: Any] = [
+            "customer": ["id": customerID],
+            "payment_source": paymentSourceDict
         ]
 
         return try? JSONSerialization.data(withJSONObject: requestBody)
```
scannillo: Instead of using a raw dict + [JSONSerizliation](https://developer.apple.com/documentation/foundation/jsonserialization), throughout the rest of the demo app we conform structs to `Encodable` & then use [JSONEncoder](https://developer.apple.com/documentation/foundation/jsonencoder)

We should keep things consistent

KunJeongPark: Hey Sammy, I did this because I didn't know a way of using struct to make blank payment JSON object
{
"payment_source":
    "card:" {}
}

Demo/Demo/Models/PaymentTokenResponse.swift
KunJeongPark: meaning they contain very minimal results like last four digits and email.

Demo/Demo/Models/SetupTokenRequest.swift
KunJeongPark: I know when we get the update setup token endpoint to work with custom url scheme, we will use experience context field but right now, it errors out with that field set with our custom url scheme so I am sending an empty card in the card vault request.

Demo/Demo/Models/PaymentTokenResponse.swift
sshropshire: 2cents: On Android we went with `CardPaymentSource` and `PayPalPaymentSource`. JSON encoding of this request is weird on Android too.

Demo/Demo/Models/SetupTokenRequest.swift
sshropshire: I remember looking at this a while back and not finding a quick answer on `Encodable` support for empty objects. Maybe we can use an empty `struct` to represent `[:]` like [here](https://stackoverflow.com/a/69348020)?

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultResultView.swift
```diff
@@ -0,0 +1,41 @@
+import SwiftUI
+import PayPalWebPayments
+
+struct PayPalVaultResultView: View {
+
+    @ObservedObject var paypalVaultViewModel: PayPalVaultViewModel
```
sshropshire:
```suggestion
    @ObservedObject var viewModel: PayPalVaultViewModel
```
🧶 : Similar to Rich's callout, since the `var` is strongly typed here we can [omit type](https://www.swift.org/documentation/api-design-guidelines/#name-according-to-roles) from the name.

sshropshire (approved):

Demo/Demo/Models/SetupTokenRequest.swift
KunJeongPark: I've never seen this. Thanks Steven. I will check it out.

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultResultView.swift
KunJeongPark: Forgot to fix this one on last PR feedback fix commit

Demo/Demo/Models/SetupTokenResponse.swift
```diff
@@ -8,9 +8,22 @@ struct SetUpTokenResponse: Decodable, Equatable {
...
     let id, status: String
     let customer: Customer?
+    let links: [Link]
+    var paypalURL: String? {
+        if let link = links.first(where: { $0.rel == "approve" }) {
+            let url = link.href
+            return url
+        }
+        return nil
+    }
```
sshropshire:
```suggestion
    var paypalURL: String? {
        return links.first { $0.rel == "approve" }?.href
    }
```
🧶 : ^ would be technically equivalent.

richherrera: 😅

```suggestion
   var paypalURL: String? {
        links.first { $0.rel == "approve" }?.href
    }
```

sshropshire: Oh nice haha. 🙌

Demo/Demo/Models/SetupTokenRequest.swift
KunJeongPark: We found EmptyParams()

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
KunJeongPark: Did you mean other tests here that we didn't modify? I can create a ticket for those, we generally like to
keep things separate.

I think we only call expectation.fulfill() only on the path that we expect, right?

richherrera: You're right, since we are expecting it to fail, it's not necessary to use fulfill

richherrera (approved):

PR #227, opened 2023-12-14
Refactor PayPal Web Demo App into Single View

jaxdesmarais:
### Reason for changes

Currently the PayPal Web feature exists across 3 total views with totally different result views. This simplifies the PayPal Web feature into a single view. It also gives us the ability to reset the PayPal flow for subsequent checkouts.

### Summary of changes

- Add `LabelViewText` to stack the text horizontally vs vertically to save space
- Remove `PayPalWebStatusView` - now that we have a single success view we can remove this fine
- Add segmented control for `PayPalWebButtonsView` - this saves space by not displaying all buttons at once
- Add `Reset` button to `PayPalWebCreateOrderView` - this allows for resetting the flow and clearing out stored properties on the ViewModel
- Add all views to `PayPalWebPaymentsView`
    - Conditionally hide/show views depending on where we are in the flow
- Remove no longer needed `OrderStatus`
- Replace `PayPalWebStatusView` with `successView` property on `PayPalWebResultView`
- Simplify `PayPalWebTransactionView` now that we have 1 shared result view
- Remove no longer needed `createOrderResult` and `transactionResult` - this can be shared across 1 updated `order` property as the order progresses

### Visuals
| Before | After |
|-----|-----|
|Simulator Screen Recording - iPhone 15 - 2023-12-14 at 17 01 18|Simulator Screen Recording - iPhone 15 - 2023-12-14 at 17 02 43|

### Checklist

### Authors

- @jaxdesmarais


Demo/Demo/SwiftUIComponents/CommonComponents/LabelViewText.swift
```diff
@@ -0,0 +1,18 @@
+import SwiftUI
+
+struct LabelViewText: View {
```
jaxdesmarais, line 3: Open to alternate names for this `struct` as well. Right now it's only used in our result views but we could potentially use it in other views.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebButtonsView.swift
```diff
@@ -1,49 +1,52 @@
...
+                    Text("Pay Later").tag(PayPalWebCheckoutFundingSource.paypalCredit)
                 }
-                .frame(maxWidth: .infinity, maxHeight: 40)
-                PayPalPayLaterButton.Representable(color: .silver, edges: .rounded, size: .full) {
-                    payPalWebViewModel.paymentButtonTapped(funding: .paylater)
+                .pickerStyle(SegmentedPickerStyle())
+
+                switch selectedFundingSource {
+                case .paypalCredit:
+                    PayPalCreditButton.Representable(color: .black, size: .full) {
```
jaxdesmarais, line 30: All the buttons are the same size now. We have a separate `Payment Button` demo that displays all of the options of size/color/etc.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebPaymentsView.swift
```diff
@@ -6,20 +6,31 @@ struct PayPalWebPaymentsView: View {
...
+
+                    PayPalWebResultView(payPalWebViewModel: payPalWebViewModel)
+
+                    if payPalWebViewModel.checkoutResult != nil {
+                        PayPalWebTransactionView(payPalWebViewModel: payPalWebViewModel)
+                            .padding(.bottom, 20)
+                            .id("bottomView")
+                            .onAppear {
+                                withAnimation {
+                                    scrollView.scrollTo("bottomView")
```
jaxdesmarais, line 25: Scrolling to the bottom of a view is most commonly used with table views or a view that has incrementing IDs to reference. In order to scroll to the bottom both when displaying this view as well as the bottom of the result after authorizing/capturing we need to duplicate this `scrollTo` logic. Bit odd but it seemed like the solution for this from what I found online. 🤷‍♂️

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
```diff
@@ -6,8 +6,7 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
 
     @Published var state: CurrentState = .idle
     @Published var intent: Intent = .authorize
-    @Published var createOrderResult: Order?
-    @Published var transactionResult: Order?
+    @Published var order: Order?
```
jaxdesmarais, line 9: We can get rid of the duplicate `Order` above now that we have a singular result view 🎉

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
```diff
@@ -73,18 +68,23 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
     }
 
-    func getPayPalClient() async throws -> PayPalWebCheckoutClient {
+    func getPayPalClient() async throws -> PayPalWebCheckoutClient? {
         do {
             let config = try await configManager.getCoreConfig()
             let payPalClient = PayPalWebCheckoutClient(config: config)
             return payPalClient
+        } catch {
+            updateState(.error(message: error.localizedDescription))
```
jaxdesmarais, line 88: Updated this logic since we wouldn't display anything if we didn't get a config back, no we will return our error view in this case

Demo/Demo/SwiftUIComponents/CommonComponents/LabelViewText.swift
sshropshire: FWIW on Android we have a similar `PropertyView`.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
sshropshire: ux: on Android we do display multiple snapshots of the `Order` at different points in time. Not sure how it's designed on iOS, but adding some context about why these two values are separate.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
```diff
@@ -105,6 +103,20 @@ class PayPalWebViewModel: ObservableObject, PayPalWebCheckoutDelegate {
...
+    func resetState() {
+        updateState(.idle)
+        order = nil
+        checkoutResult = nil
+        orderID = nil
+    }
+
+    private func updateOrder(_ order: Order) {
+        DispatchQueue.main.async {
+            self.orderID = order.id
```
sshropshire: 🧶 : Do we need the ID property since we have `order` as the source of truth?

sshropshire (approved):

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
jaxdesmarais: We were mainly using this as a psudo state on iOS to display slightly different results. Since we have 1 result that is appended to as we get more data, we don't need 2 properties.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
jaxdesmarais: We don't necessarily but it seemed easier to extract this into a constant vs doing `if let orderID = order.id` when we needed to access it in the `paymentButtonTapped` and `createTransaction` method. If we don't find it helpful we can extract as needed.

richherrera: I think we could convert `orderId` into a computed property

```
var orderId: String? {
  order?.id
}
```

richherrera (approved):

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
jaxdesmarais: I dig it, done: https://github.com/paypal/paypal-ios/pull/227/commits/2b0c8db4aa1d593beb2afaffdeb2abbc57620d3c

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
KunJeongPark: Hey @jaxdesmarais , it looks really smooth! I love the toggling of the buttons there in the UI. I am wondering about replacing result view from create with authorize/capture result view. I liked having history of the states, I think that's why we retained the different order snapshots in the first place, to compare the results to make sure that things match up. Another thing that is kind of unexpected for me is that the result view after order creation appears after the toggle for the funding source.

Demo/Demo/SwiftUIComponents/CommonComponents/LabelViewText.swift
KunJeongPark: I like this LabelViewText!

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebViewModel.swift
jaxdesmarais: Just getting back to these comments from being on PTO. 😄 

> I am wondering about replacing result view from create with authorize/capture result view. I liked having history of the states, I think that's why we retained the different order snapshots in the first place, to compare the results to make sure that things match up.

We can use breakpoints in these cases but should not busy up our demo app with different views containing the same data type. We could also add a quick print statement of the states if needed in each method that returns an order.

> Another thing that is kind of unexpected for me is that the result view after order creation appears after the toggle for the funding source.

This made the most sense to keep the result view pinned in the same general place throughout the flow without it jumping around as we added subsequent views. Otherwise, we'd require unnecessary logic to move this view around throughout the flow. I had tested this initially and it was much cleaner to keep the result view relatively in 1 place.

KunJeongPark: Hey Jax welcome back! Yeah, I personally like comparing different states instead of new ones swapping out the other for comparison. Steven and I had talked about the one of primary uses for the demo app being able to surface issues or prove that a feature is working correctly.  To me, being able to have different states is helpful. It definitely is cleaner the way you refactored it.

To me, being able to see history more clearly without print statements, things appearing chronologically is more important than cleaner code for demo app.

PR #232, opened 2024-01-03
3DS Vault with Purchase (Card)

richherrera:
### Summary of changes

* Add extra validations to 3DS Contingency
* Parse and return `liabilityShift` value from 3DS
* Add `liabilityShift` to CardResult
* Handle success and failure in various scenarios while validating 3DS
* Handle success and failure from deeplink 3DS url

### Checklist

- [x] Added a changelog entry

### Authors
@richherrera 
@KunJeongPark


Sources/CardPayments/CardClient.swift
```diff
@@ -74,13 +74,20 @@ public class CardClient: NSObject {
             do {
                 let result = try await checkoutOrdersAPI.confirmPaymentSource(cardRequest: request)
                 
-                if let url: String = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                if result.status == "PAYER_ACTION_REQUIRED",
+                let url = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                    
+                    guard let url = URL(string: url), url.absoluteString.contains("flow=3ds"), url.absoluteString.contains("helios") else {
+                        self.notifyFailure(with: CardClientError.threeDSecureURLError)
```
KunJeongPark, line 82: I'm wondering now if we should just check for 
`if let url = result.links?.first(where: { $0.rel == "payer-action" })?.href, url.contains("flow=3ds"), url.contains("helios") {` as conditions for going into 3DS flow and have the same guard in the 3DS for URL. The condition we are adding is to make sure that the payer-action is 3ds flow. 

I think we should distinguish intent to go into 3DS flow vs invalid URL error (guard in the 3DS function)

I think what we had originally was good, we just append conditions:
```
    if result.status == "PAYER_ACTION_REQUIRED",
    let url: String = result.links?.first(where: { $0.rel == "payer-action" })?.href,
    url.contains("3ds"), url.contains("helios") {
           analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:challenge-required")
           startThreeDSecureChallenge(url: url, orderId: result.id)
     } else {
            analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:succeeded")                    
            let cardResult = CardResult(orderID: result.id, deepLinkURL: nil)
            notifySuccess(for: cardResult)
    }
```
oh I forgot, the first string should be "flow=3ds" although 3ds should be good enough.

KunJeongPark: I think if we do want to reuse the startThreeDSecureChallenge function to return to different notify delegate functions for vaulting without purchase, we should not include result.status, I think for vault without purchase, it is "APPROVED". I have to double check my notes.

KunJeongPark: I got it. What we currently makes sense, we are just validating the status field with 3DS team. (whether without error, the only options are "PAYER_ACTION_REQUIRED" and "APPROVED")

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
+* CardPayments
+  * Add `liabilityShift` property from `CardResult`
```
scannillo:
```suggestion
  * Add `liabilityShift` property to `CardResult`
```

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
...
+  * Add `CardClientError.threeDSVerificationError` for invalid verification
+  * Add `CardClientError.missingDeeplinkgURLError` for missing deeplink URL from 3DS
```
scannillo:
```suggestion
  * Add `CardClientError.missingDeeplinkURLError` for missing deeplink URL from 3DS
```

Also nitpick - since 3DS is a protocol instead of a service, `from 3DS` is a little odd to me. Maybe `from 3DS service`, or just drop the `from 3DS`.

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +129,28 @@ public class CardClient: NSObject {
...
+                guard let url = url else {
+                    self.notifyFailure(with: CardClientError.missingDeeplinkgURLError)
+                    return
+                }
+                
+                if url.absoluteString.contains("state=undefined") && url.absoluteString.contains("code=undefined") {
+                    let liabilityShift = self.getQueryStringParameter(url: url.absoluteString, param: "liability_shift")
+                    let cardResult = CardResult(orderID: orderId, deepLinkURL: url, liabilityShift: liabilityShift)
+                    self.notifySuccess(for: cardResult)
+                } else if url.absoluteString.contains("error=error") {
```
scannillo: Would this be more flexible if it checked for just that the error key was present, and not that it also equaled to `error`? In case helios made some change to return us proper result details, our SDK wouldn't break.

KunJeongPark: ah that makes sense. so if they in the future returned specific "error=someError1"? That's a good call.

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +129,28 @@ public class CardClient: NSObject {
...
                 }
                 
-                let cardResult = CardResult(orderID: orderId, deepLinkURL: url)
-                self.notifySuccess(for: cardResult)
+                guard let url = url else {
+                    self.notifyFailure(with: CardClientError.missingDeeplinkgURLError)
+                    return
+                }
+                
+                if url.absoluteString.contains("state=undefined") && url.absoluteString.contains("code=undefined") {
```
scannillo: If we convert our URL to a [URLComponents](https://developer.apple.com/documentation/foundation/urlcomponents) type we can use it's built in [queryItems array](https://developer.apple.com/documentation/foundation/urlcomponents/1779966-queryitems) versus parsing from the raw string directly. 

Something like ...

```
let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
for queryItem in components.queryItems {
    if queryItem.name == "error" {
        // do stuff
    }
}
                
```

KunJeongPark: we can use getQueryStringParameter function that we copied from another module and check for nil value there.

CHANGELOG.md
richherrera: Thanks, done: https://github.com/paypal/paypal-ios/commit/7945475a59342158efe525f24e0e777b4fc5eced

CHANGELOG.md
richherrera: Nice catch, done: https://github.com/paypal/paypal-ios/commit/7945475a59342158efe525f24e0e777b4fc5eced

Sources/CardPayments/CardClient.swift
richherrera: you're right, good call, done: https://github.com/paypal/paypal-ios/pull/232/commits/ba6397f0ae564fe7b9f9c97d53c3b40b31740d4b

Sources/CardPayments/CardClient.swift
richherrera: thanks team, excellent advice, done: https://github.com/paypal/paypal-ios/pull/232/commits/ba6397f0ae564fe7b9f9c97d53c3b40b31740d4b

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -9,8 +9,10 @@ public struct CardResult {
...
     public let orderID: String
 
-    // TODO: parse contents of this URL once we are clear on values returned from 3ds
     /// This is the deep link url returned from 3DS authentication
     @_documentation(visibility: private)
     public let deepLinkURL: URL?
+    
+    /// Liability shift value returned from 3DS verification
+    public let liabilityShift: String?
 }
```
KunJeongPark, line 18: I'm thinking about the results we return for card vaulting 3DS flow also. For non-3ds contingency flow, we have the "APPROVED" status from confirm-payment-source. Should we return "APPROVED" status for that flow and liabilityShift for 3DS flow?

KunJeongPark: So CardResult/CardVaultResult would have optional liabilityShift and optional status field.
What are your thoughts?

KunJeongPark: We used to have "status" field for CardResult but we removed it since status doesn't get updated after 3DS flow.

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +131,29 @@ public class CardClient: NSObject {
                     }
                 }
                 
-                let cardResult = CardResult(orderID: orderId, deepLinkURL: url)
-                self.notifySuccess(for: cardResult)
+                guard let url = url else {
```
jaxdesmarais:
```suggestion
                guard let url else {
```

jaxdesmarais:
```suggestion
                guard let url else {
```

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +131,29 @@ public class CardClient: NSObject {
...
+                    return
+                }
+                
+                if self.getQueryStringParameter(url: url.absoluteString, param: "state") != nil
+                && self.getQueryStringParameter(url: url.absoluteString, param: "code") != nil {
+                    let liabilityShift = self.getQueryStringParameter(url: url.absoluteString, param: "liability_shift")
+                    let cardResult = CardResult(orderID: orderId, deepLinkURL: url, liabilityShift: liabilityShift)
+                    self.notifySuccess(for: cardResult)
+                } else if self.getQueryStringParameter(url: url.absoluteString, param: "error") != nil {
+                    self.notifyFailure(with: CardClientError.threeDSVerificationError)
```
jaxdesmarais, line 144: Is the error here ever helpful? If so we should pass it through as an associated value in the `threeDSVerificationError`.

Sources/CardPayments/CardClient.swift
```diff
@@ -45,7 +45,7 @@ public class CardClient: NSObject {
                 let result = try await vaultAPI.updateSetupToken(cardVaultRequest: vaultRequest).updateVaultSetupToken
                 
                 // TODO: handle 3DS contingency with helios link & add unit tests
-                    if let link = result.links.first(where: { $0.rel == "approve" && $0.href.contains("helios") }) {
+                if let link = result.links.first(where: { $0.rel == "approve" && $0.href.contains("helios") }) {
```
jaxdesmarais, line 48: Could we update `link` to something more descriptive here? Maybe `threeDSApprovalLink` or `threeDSContingencyLink`?

Sources/CardPayments/CardClient.swift
```diff
@@ -45,7 +45,7 @@ public class CardClient: NSObject {
...
                     let url = link.href
                     print("3DS url \(url)")
```
jaxdesmarais, line 50: Nit for a future PR but we should remove this print from the client, I'm guessing this was just missed from testing something at some point

Sources/CardPayments/Models/CardResult.swift
richherrera: If we remove the `status` field from `CardResult` for some reason, I believe it's not necessary to add it back right now. If the Vaulting flow with 3DS requires it, we can reintroduce it later

Sources/CardPayments/CardClientError.swift
```diff
@@ -73,4 +82,22 @@ enum CardClientError {
...
+    static let missingDeeplinkURLError = CoreSDKError(
+        code: Code.missingDeeplinkURLError.rawValue,
+        domain: domain,
+        errorDescription: "Missing deeplink URL from 3DS."
+    )
+    
+    static let malformedDeeplinkURLError = CoreSDKError(
+        code: Code.malformedDeeplinkURLError.rawValue,
+        domain: domain,
+        errorDescription: "Malformed deeplink url."
```
jaxdesmarais:
```suggestion
        errorDescription: "Malformed deeplink URL."
```

Sources/CardPayments/Models/CardResult.swift
KunJeongPark: I think it's a bit weird to just return orderID when we know we have "approved" status for non 3DS flow. We had removed it because we didn't have reqs on 3DS clear at the time.

Sources/CardPayments/CardClient.swift
richherrera: Done https://github.com/paypal/paypal-ios/pull/232/commits/c6905e819066abe4cd6c6ab3d1acc53e2a6b41ad

Sources/CardPayments/CardClientError.swift
richherrera: Done https://github.com/paypal/paypal-ios/pull/232/commits/c6905e819066abe4cd6c6ab3d1acc53e2a6b41ad

Sources/CardPayments/CardClient.swift
richherrera: Sure, since this was just an indentation change, we'll consider it in a different PR, (Vault without purchase)

Sources/CardPayments/CardClient.swift
KunJeongPark: Do you mean that error value? We get "error=error" which just means existence of error, no specifics yet.

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +131,29 @@ public class CardClient: NSObject {
...
                 
-                let cardResult = CardResult(orderID: orderId, deepLinkURL: url)
-                self.notifySuccess(for: cardResult)
+                guard let url else {
+                    self.notifyFailure(with: CardClientError.missingDeeplinkURLError)
+                    return
+                }
+                
+                if self.getQueryStringParameter(url: url.absoluteString, param: "state") != nil
+                && self.getQueryStringParameter(url: url.absoluteString, param: "code") != nil {
```
jaxdesmarais: It looks like we are doing several additional checks throughout this flow that we are not doing in the [Android PR](https://github.com/paypal/paypal-android/pull/227/files). Is there a reason these checks are different across platforms?

Sources/CardPayments/CardClient.swift
jaxdesmarais: Gotcha, yeah, if the error is just the raw string "error" probably not helpful to merchants, was mainly curious if it's ever more descriptive of the underlying issue.

Sources/CardPayments/CardClientError.swift
```diff
@@ -28,6 +28,15 @@ enum CardClientError {
...
         /// 6. GraphQLClient is unexpectedly nil
         case nilGraphQLClientError
+        
+        /// 7. An error from 3DS verification
+        case threeDSVerificationError
+        
+        /// 8. Missing Deeplink URL from 3DS
+        case missingDeeplinkURLError
+        
+        /// 9. Missing Deeplink URL from 3DS
```
jaxdesmarais: It looks like this docstring is a copy of the above, should it be different?

Sources/CardPayments/CardClient.swift
KunJeongPark: It's consistent with JS implementation with just generic error being communicated for now.

Sources/CardPayments/CardClient.swift
KunJeongPark: I will sync with Steven and Rich today about this.

Sources/CardPayments/CardClientError.swift
KunJeongPark: Good catch!

richherrera: Thanks, done: https://github.com/paypal/paypal-ios/pull/232/commits/2896471c1533de72880afd18545a33a374f9cfc0

Sources/CardPayments/Models/CardResult.swift
KunJeongPark: We can align with Android and just keep what we are returning and discuss maybe adding it later. I'm good with that.

Sources/CardPayments/CardClient.swift
KunJeongPark: So our reasoning for adding this additional check is for the case that the url is not properly formatted for error or success, we don't want to return success. I added a comment on the Android PR.

jaxdesmarais: Thanks for raising that on the Android PR! More just wanted to make sure we were using the same logic across the platforms and it sounds like what exists on iOS in this PR _should_ exist on Android as well.

Sources/CardPayments/CardClient.swift
```diff
@@ -74,13 +74,22 @@ public class CardClient: NSObject {
             do {
                 let result = try await checkoutOrdersAPI.confirmPaymentSource(cardRequest: request)
                 
-                if let url: String = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                if result.status == "PAYER_ACTION_REQUIRED",
+                let url = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+
```
jaxdesmarais:
```suggestion
```

Sources/CardPayments/CardClient.swift
```diff
@@ -74,13 +74,22 @@ public class CardClient: NSObject {
             do {
                 let result = try await checkoutOrdersAPI.confirmPaymentSource(cardRequest: request)
                 
-                if let url: String = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                if result.status == "PAYER_ACTION_REQUIRED",
+                let url = result.links?.first(where: { $0.rel == "payer-action" })?.href {
+                    
+                    guard getQueryStringParameter(url: url, param: "flow") == "3ds",
+                    url.contains("helios"),
+                    let url = URL(string: url) else {
```
jaxdesmarais: super nit, but this spacing looks off - can we `cmd` + `i` it to make sure it's correct?

Sources/CardPayments/CardClient.swift
jaxdesmarais: same nit here on spacing, maybe it's just me but `cmd` + `i` should let us know for sure. 😓

richherrera: Done: https://github.com/paypal/paypal-ios/pull/232/commits/ad6cebf4df635c50a6ced1627f5e81666a1f84ec

Sources/CardPayments/CardClient.swift
richherrera: Done: https://github.com/paypal/paypal-ios/pull/232/commits/ad6cebf4df635c50a6ced1627f5e81666a1f84ec
Thanks 👍

jaxdesmarais (approved):
Thanks for addressing all my many comments/questions as well as making sure we are aligned across iOS and Android!

Sources/CardPayments/CardClient.swift
richherrera: It's a pleasure. Thanks for the feedback. Android has the same logic as ours.

Sources/CardPayments/CardClient.swift
scannillo, line 50: I believe we are waiting to display the link in an ASWebAuthSession until the url scheme redirect is working? If so, maybe we can update the TODO note to call that out.

Sources/CardPayments/CardClient.swift
```diff
@@ -127,11 +130,29 @@ public class CardClient: NSObject {
...
+                
+                if self.getQueryStringParameter(url: url.absoluteString, param: "state") != nil
+                    && self.getQueryStringParameter(url: url.absoluteString, param: "code") != nil {
+                    let liabilityShift = self.getQueryStringParameter(url: url.absoluteString, param: "liability_shift")
+                    let cardResult = CardResult(orderID: orderId, deepLinkURL: url, liabilityShift: liabilityShift)
+                    self.notifySuccess(for: cardResult)
+                } else if self.getQueryStringParameter(url: url.absoluteString, param: "error") != nil {
+                    self.notifyFailure(with: CardClientError.threeDSVerificationError)
+                } else {
+                    self.notifyFailure(with: CardClientError.malformedDeeplinkURLError)
```
scannillo, line 146: 👏 These new error cases are really helpful

Sources/CardPayments/CardClientError.swift
```diff
@@ -28,6 +28,15 @@ enum CardClientError {
 
         /// 6. GraphQLClient is unexpectedly nil
         case nilGraphQLClientError
+        
+        /// 7. An error from 3DS verification
+        case threeDSVerificationError
+        
+        /// 8. Missing Deeplink URL from 3DS
+        case missingDeeplinkURLError
```
scannillo, line 36: Take it or leave it comment - we usually call the URL that we get _back_ from a service the `returnURL`

scannillo (approved):
🚀

Sources/CardPayments/CardClient.swift
richherrera: Sure, this change was just part of fixing the indentation, we already have a draft where we are addressing this https://github.com/paypal/paypal-ios/pull/238, we will gladly resolve your comments in that PR as well.

Sources/CardPayments/CardClient.swift
richherrera: Thanks for the advice!!!

PR #233, opened 2024-01-03
Remove duplicate image assets & unused PaymentButtonImage file

scannillo:
### Reason for changes

- Was doing investigation work for picking up DTBTSDK-3248 to support VoiceOver and came across some cleanup to be done

### Summary of changes

- Remove unused `PaymentButtonImage.swift`
- Remove unused/duplicate `PayPalLogo` and `PayPalCreditLogo` assets
- Add 1x universal image size for `paypal_color` image asset

### Checklist

- ~Added a changelog entry~
- [ ] I verified in our demo app that all of the button varieties look OK, but would appreciate if someone else can do those verifications as well as we don't have any UI tests at the moment

### Authors
@scannillo


richherrera (approved):
🧼🧽

KunJeongPark (approved):

PR #235, opened 2024-01-04
Payment Button Label Changes

stechiu:
### Reason for changes

Button Redesign for QL calls for new labels and updated image assets.

### Summary of changes

- Added new labels for Payment Button
- Changed label section to Wheel Picker Style in Demo app
- Changed Credit image asset
- Changed logos for different button sizes
- Changed image asset names

### Design Requirements
https://www.figma.com/file/gn4AXdmJUOMVlrCMpyhfrf/QL---Checkout-Buttons?type=design&node-id=1439-12490&mode=design&t=1rYqUZkbsXdsBURn-0

https://paypal.atlassian.net/wiki/spaces/Checkout/pages/697558273/Button+Designs+-+Requirements

### Other
- The `mini` and `miniWithLogomark` are rectangles in the redesign. They are left as circles and will be addressed in another PR
- This PR doesn't include the optional tagline below the button
Screenshot 2024-01-03 at 4 29 34 PM

### Visuals

| Button | Figma | PR | 
| - | - | - |
| PayPal Button (The labels are optional) | Screenshot 2024-01-03 at 3 18 54 PM<br>Screenshot 2024-01-03 at 3 18 32 PM | <video src="https://github.com/paypal/paypal-ios/assets/127455800/cae3243e-b5fb-4340-8e22-4e6959a974da" width="300" /> |
| Pay Later Button | Screenshot 2024-01-03 at 3 21 22 PM | <video src="https://github.com/paypal/paypal-ios/assets/127455800/ced0b247-b5f8-42b7-b07f-5e60a03ecf70" width="300" /> |
| PayPal Credit | Screenshot 2024-01-03 at 3 22 25 PM | <video src="https://github.com/paypal/paypal-ios/assets/127455800/2f565943-f2b8-47b4-8fbb-de4507dc5d5e" width="300" /> |
| Logomark (mini) | Screenshot 2024-01-03 at 3 23 33 PM | <video src="https://github.com/paypal/paypal-ios/assets/127455800/0b2c0383-7224-421f-a952-3ed501e83ffe" width="300" /> |

### To do

- Update changelog 
- There are 3 different designs for Pay Later. This PR uses Proposal 2a. I'm waiting for clarification from Patrick who is OOO until Jan 4.
Screenshot 2024-01-03 at 3 09 35 PM

### Checklist

- [x] Added a changelog entry

### Authors
@stechiu


richherrera (approved):
LGTM!!!

KunJeongPark:
Looks great Steph! I will let Patrick do the final approval.

Sources/PaymentButtons/Assets.xcassets/Redesign Logos/paypal_monogram_credit.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
```
scannillo: Do we need a separate `Redesign Logos` and `Logos` directory within Assets?

Sources/PaymentButtons/Assets.xcassets/Redesign Logos/paypal_credit.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
```
scannillo: In the [VoiceOver PR](https://github.com/paypal/paypal-ios/pull/236), I took the approach of using the `.imageset` name to determine which logo is being displayed. If it was prefixed with `paypal_`, I assumed it only had the PP logo / PP text, while if it was prefixed with `credit_` then it contained the PP logo & credit text.

If we want to use that pattern, then we would need to update the names of these assets to move `credit_` to the front.

I am down though for a better/more clear way for us to know in code what content each image type has. Is there a more clear naming convention we can use? Also the logic which appends the `imageAssetString` together in `ImageAsset.swift` is a little confusing IMO. It might be more clear to just have the fully hard coded strings that match the `.imageset` filenames vs concatenating. What do folks think?

Sources/PaymentButtons/Assets.xcassets/Redesign Logos/paypal_monogram_credit.imageset/Contents.json
stechiu: That's true. I wasn't sure if we needed to keep the old logos for anything so I separated them so that it would be easier to go through the `ImageAsset` logic. I'll go ahead and remove the old logos

scannillo: Gotcha - yeh I think we can 🧨 nuke the old ones!

Sources/PaymentButtons/Assets.xcassets/Redesign Logos/paypal_credit.imageset/Contents.json
stechiu: I also had trouble with the logic when I was updating the file and was going off of what was originally there. I assume the concatenation was to reduce the amount of duplicate code. I'd love for them to be hardcoded to match `.imageset` filenames rather than concatenating so I'll update to that!

stechiu: Screenshot 2024-01-04 at 3 04 38 PM

Here are the changes I made

scannillo: Awesome! Yeh, I agree - a bit easier to read and less error prone.

🏈 After we get the main parts of this work done for QL, would be great to add some UI/snapshot tests for these buttons

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+* PaymentButtons
+  * Add new size `.miniWithWordmark`
+  * Add new labels, such as `.addMoneyWith`, `.bookWith`, and `buyNowWith` for `PayPalButton.Label`
```
scannillo: We should call out the enum cases for `PaymentButtonLabel` we removed (`.checkout` and `.buyNow`) since those removals will break a merchant app.

Nit - but I would also include the full list of the enum cases added, even though it's long

scannillo: By break a merchant app, I mean if they were to upgrade to V2 and their code was:

```
PayPalButton.Representable(
    color: .black,
    edges: .softEdges,
    size: .expanded,
    label: .buyNow
)
```

it would not compile since the `.buyNow` option is gone

stechiu: Good point! Thanks, I'll make the changes

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+* PaymentButtons
+  * Add new size `.miniWithWordmark`
+  * Add new `PayPalButton` labels: `.addMoneyWith`, `.bookWith`, `.buyWith`, `.buyNowWith`, `.checkoutWith`, `.continueWith`, `.contributeWith`, `.orderWith`, `.payLaterWith`, `.payWith`, `.reloadWith`, `.rentWith`, `.reserveWith`, `.subscribeWith`, `.supportWith`, `.tipWith`, `.topUpWith`
+* Breaking Changes
+  * Remove `.checkout`, `.buyNow`, `.payLater`
```
scannillo: Nit - for adding the full typename associated with each enum case

```suggestion
  * Add new `PaymentButtonSize` case `.miniWithWordmark`
  * Add new `PayPalButton.Label` cases: `.addMoneyWith`, `.bookWith`, `.buyWith`, `.buyNowWith`, `.checkoutWith`, `.continueWith`, `.contributeWith`, `.orderWith`, `.payLaterWith`, `.payWith`, `.reloadWith`, `.rentWith`, `.reserveWith`, `.subscribeWith`, `.supportWith`, `.tipWith`, `.topUpWith`
* Breaking Changes
  * Remove `PayPalButton.Label` cases: `.checkout`, `.buyNow`, `.payLater`
```

Sources/PaymentButtons/PaymentButtonLabel.swift
```diff
@@ -3,30 +3,96 @@ import Foundation
...
         case suffix
     }
 
     var position: Position {
         switch self {
-        case .checkout, .buyNow, .payLater:
-            return .suffix
-
-        case .payWith:
+        case .addMoneyWith,
```
scannillo: Looks like the linter is complaining here. We probably simplify and do something like:

```
    var position: Position {
        return self == .payLater ? .suffix : .prefix
    }
```

stechiu: That's so much cleaner, thanks for the suggestion!

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -69,7 +69,27 @@ extension PaymentButtonFundingSource: CaseIterable {
...
-        [.none, .checkout, .buyNow, .payWith]
+        [
+            .addMoneyWith,
+            .bookWith,
+            .buyNowWith,
+            .buyWith,
+            .checkoutWith,
+            .continueWith,
+            .contributeWith,
+            .none,
```
scannillo: Nit - usually none comes first in the list of all cases

scannillo (approved):

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
stechiu: Thanks, will make the update!

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -69,7 +69,27 @@ extension PaymentButtonFundingSource: CaseIterable {
 extension PayPalButton.Label: CaseIterable {
     
     public static var allCases: [PayPalButton.Label] {
-        [.none, .checkout, .buyNow, .payWith]
+        [
```
scannillo, line 72: Can't we get `allCases` for free ([see sample](https://sarunw.com/posts/swift-caseiterable/#getting-all-cases-of-an-enum-with-swift-caseiterable))?

scannillo: Feel free to 🏈 punt on that one though - looks like an opportunity to cleanup our demo app in a follow-up PR

stechiu: Oh right, I overlooked using map filter and that would've made things easier... I'm going to open up a new PR for adding the optional tagline under the PayPal Button so I'll include that in there and demo app cleanup

stechiu: When I use
```
    public static var allCases: [PayPalButton.Label] {
        PayPalButton.Label.allCases.map { $0 }
    }
```
there's a warning on `allCases` stating "Function call causes an infinite recursion". Correct me if I'm wrong but I believe it's because `PayPalButton.Label` has associated value and `CaseIterable` will automatically synthesize `allCases` for enum cases without associated value. If there's an associated value, then the enum values need to be typed out

PR #237, opened 2024-01-08
Updated with PayPalOpen font

stechiu:
### Reason for changes

Need to update fonts for Payment Buttons to `PayPalOpen` font as the primary font as part of QL.

| Figma | PR |
| ----------- | ----------- |
| Screenshot 2024-01-08 at 12 01 09 PM | Screenshot 2024-01-08 at 12 01 30 PM |

### Summary of changes

- Added new fonts
- Replaced Payment fonts with new fonts

### Checklist

- [ ] Added a changelog entry

@stechiu


stechiu:
Since the change is in the font typeface and isn't breaking, nor do Merchants need to make any changes, does the change still need to be included in the CHANGELOG?

scannillo:
> Since the change is in the font typeface and isn't breaking

I would classify changing UI without a merchant having to opt-in to that change, as a breaking change. Even though it's super minor (in this case of a font that doesn't look too different), by the book it would be breaking.

We _could_ classify this is a "bug-fix to meet brand guidelines" if we want to get it into V1 😆. Sketchy? Thoughts?

Sources/PaymentButtons/PaymentButtonFont.swift
```diff
@@ -1,12 +1,15 @@
 import UIKit
 
 enum PaymentButtonFont {
-    /// The primary font to be employed in the UI for non special circumstances
-    public static let primaryFont: UIFont = .systemFont(ofSize: UIFont.systemFontSize)
+    /// The primary font used for non special circumstances
```
jaxdesmarais: I know this docstring existed before but it's a bit confusing to me, mainly the "non special circumstances" bit. Maybe we can just update to "The primary font used". Thoughts?

jaxdesmarais:
> We could classify this is a "bug-fix to meet brand guidelines" if we want to get it into V1 😆. Sketchy? Thoughts?

I think classifying this as a bug fix makes sense! I assume there is some "brand reputation" to actually using the correct fonts so could make sense for a minor version + changelog update.

Sources/PaymentButtons/PaymentButtonFont.swift
stechiu: Thanks! Yea it was a bit wordy and "The primary font used" would be enough

stechiu:
> > Since the change is in the font typeface and isn't breaking
> 
> I would classify changing UI without a merchant having to opt-in to that change, as a breaking change. Even though it's super minor (in this case of a font that doesn't look too different), by the book it would be breaking.
> 
> We _could_ classify this is a "bug-fix to meet brand guidelines" if we want to get it into V1 😆. Sketchy? Thoughts?

Got it, that makes sense! I read through the Semantic Versioning Guideline and must've overlooked UI related changes. In that case, we can have it go into V2 where we have other breaking changes instead.

scannillo:
> must've overlooked UI related changes

You definitely didn't! UI specific stuff isn't clearly defined in there.

I like Jax's thoughts - I'm also down for lumping into V1!

Sources/PaymentButtons/PaymentButtonFont.swift
```diff
@@ -1,12 +1,15 @@
 import UIKit
 
 enum PaymentButtonFont {
-    /// The primary font to be employed in the UI for non special circumstances
-    public static let primaryFont: UIFont = .systemFont(ofSize: UIFont.systemFontSize)
+    /// The primary font
+    public static let paypalPrimaryFont = UIFont(name: "PayPalOpen-Regular", size: 14) ??
```
scannillo, line 5: It looks like we aren't using the `-Medium` or `-Bold` font options? Are we planning to at some point?

Sources/PaymentButtons/PaymentButtonFont.swift
```diff
@@ -1,12 +1,15 @@
...
+        .systemFont(ofSize: UIFont.systemFontSize)
 
-    /// Usually a sub-heading or alternative text font that is used in some scenarios
-    public static let secondaryFont: UIFont = .systemFont(ofSize: UIFont.smallSystemFontSize)
+    /// The secondary font for sub-heading or alternative text
+    public static let paypalSecondaryFont = UIFont(name: "PayPalOpen-Regular", size: 12) ??
```
richherrera: I'm not sure, but I believe the Fonts are being added at the application level (Demo App, e.g: Demo/Resources/Fonts....). I think they should be added to PayPal and set as the target membership for PaymentButtons (Sources/PaymentButtons) 🤔

Sources/PaymentButtons/PaymentButtonFont.swift
stechiu: Not that I'm aware of unless they'll be used in other parts of the SDK. I can remove for now and we can always add them in the future as needed

Sources/PaymentButtons/PaymentButtonFont.swift
stechiu: Good call, I thought I had the files in `PayPal` but it ended up in `Demo`. Updated thanks!

Sources/PaymentButtons/PaymentButtonFont.swift
stechiu: Actually, since we're changing to 2 button sizes and not incorporating taglines, we only need primary font. Otherwise the other font sizes were needed for different button sizes and tagline

Sources/PaymentButtons/PaymentButtonSize.swift
```diff
@@ -17,14 +17,9 @@ public enum PaymentButtonSize: Int, CustomStringConvertible {
...
-            return PaymentButtonFont.secondaryFont
-
-        case .expanded:
-            return PaymentButtonFont.primaryFont
-
-        case .full:
-            return PaymentButtonFont.systemFont18
+        case .mini, .collapsed, .expanded, .full:
+            return UIFont(name: "PayPalOpen-Regular", size: 14) ??
+                .systemFont(ofSize: UIFont.systemFontSize)
```
stechiu: FYI: Button sizes will reduce to `.mini` and `.expanded` in another PR

Demo/Demo/Info.plist
```diff
@@ -2,6 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UIAppFonts</key>
+	<array>
+		<string>PayPalOpen-Regular.otf</string>
+		<string>PayPalOpen-Medium.otf</string>
+		<string>PayPalOpen-Bold.otf</string>
```
sshropshire: qq: Do we need both Medium and Bold fonts? Checking to see if we could save merchants a few bytes.

stechiu: They aren't needed for the redesign, so I'll remove

Demo/Demo/Info.plist
```diff
@@ -2,6 +2,12 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>UIAppFonts</key>
```
scannillo: Should these be added to the Demo app's info.plist?

stechiu: I tried removing it from the info.plist and it ends up defaulting to SF Pro. Did some research and in order to add new fonts, we need to add the font files and register it. I did, however, remove medium and bold now since they're unused

stechiu: Ah I didn't know earlier, but I need to add the font file in Demo app for it to show in the simulator, so I've gone ahead and done that

jaxdesmarais: Is this just due to how we pull in the SDK like a direct/carthage integration or will Cocoapods and SPM merchants need to do the same?

richherrera: I found an interesting post related to ([Add Font In Customize Framework Swift 4.2](https://medium.com/@GoalStack/add-font-in-customize-framework-swift-4-2-150a28bd8482)), maybe we can follow this approach

CHANGELOG.md
```diff
@@ -5,15 +5,18 @@
...
 * PayPalWebPayments
   * Add `vault(url:)` method to `PayPalWebCheckoutClient`
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
-* PaymentButtons
-  * Add `custom` case for `PaymentButtonEdges`
-  * Support VoiceOver by adding button accessibility labels
+
+* Breaking Changes
```
scannillo: We can remove the `Breaking Changes` label if this is going in as a patch change on `main` (V1)

Sources/PaymentButtons/PaymentButtonSize.swift
```diff
@@ -17,14 +17,8 @@ public enum PaymentButtonSize: Int, CustomStringConvertible {
...
-        case .mini, .collapsed:
-            return PaymentButtonFont.secondaryFont
-
-        case .expanded:
-            return PaymentButtonFont.primaryFont
-
-        case .full:
-            return PaymentButtonFont.systemFont18
+        case .mini, .collapsed, .expanded, .full:
+            return PaymentButtonFont.paypalPrimaryFont
```
jaxdesmarais: I don't think we need this switch at all, can just do:
```swift
var font: UIFont {
    PaymentButtonFont.paypalPrimaryFont
}
```

stechiu: Thanks, I'll get that fixed

CHANGELOG.md
```diff
@@ -3,22 +3,23 @@
...
   * Support VoiceOver by adding button accessibility labels
+  * Add `custom` case for `PaymentButtonEdges`
+  * Support VoiceOver by adding button accessibility labels
+  * Font typeface changed to "PayPalOpen" to meet brand guidelines
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Add `CardClientError.threeDSVerificationError` for invalid verification
   * Add `CardClientError.missingDeeplinkURLError` for missing deeplink URL
   * Add `CardClientError.malformedDeeplinkURLError` for malformed or invalid deeplink
+* PayPalWebPayments
```
jaxdesmarais, line 18: Do we need to merge the latest from `main` in? I think some of this CHANGELOG is from other PRs.

stechiu: The latest on `main` was merged in. What part of this CHANGELOG shouldn't be here?

jaxdesmarais: Ah got it, sorry I was confused by the diff but it looks like the order of things moved around a bit to keep all the button stuff together

Sources/PaymentButtons/PaymentButtonSize.swift
```diff
@@ -16,16 +16,7 @@ public enum PaymentButtonSize: Int, CustomStringConvertible {
...
-        case .mini, .collapsed:
-            return PaymentButtonFont.secondaryFont
-
-        case .expanded:
-            return PaymentButtonFont.primaryFont
-
-        case .full:
-            return PaymentButtonFont.systemFont18
-        }
+        return PaymentButtonFont.paypalPrimaryFont
```
jaxdesmarais: Can remove the keyword return and rely on implicit return here:
```suggestion
        PaymentButtonFont.paypalPrimaryFont
```

Demo/Demo/Info.plist
```diff
@@ -2,6 +2,10 @@
...
```
jaxdesmarais: Do we have a ticket to add documentation for this for merchants? It seems like this is something that a merchant would need to do to use our fonts. What happens if they don't add this to their `Info.plist`?

stechiu: This line was added in the CHANGELOG under Payment Buttons so that merchants know about the font change: `Font typeface changed to "PayPalOpen" to meet brand guidelines`

The font shouldn't be in `Info.plist` so I've removed it. The font is registered in the SDK under the `PaymentButtonFont` file so that merchants won't need to add the font themselves

jaxdesmarais (approved):

Sources/PaymentButtons/PaymentButtonFont.swift
```diff
@@ -1,12 +1,30 @@
...
+        guard
+            let pathForResourceString = frameworkBundle.path(forResource: name, ofType: fileExtension),
+            let fontData = NSData(contentsOfFile: pathForResourceString),
+            let dataProvider = CGDataProvider(data: fontData),
+            let fontRef = CGFont(dataProvider) else { return }
+        print("Font file path:", pathForResourceString)
 
-    /// Usually a sub-heading or alternative text font that is used in some scenarios
-    public static let secondaryFont: UIFont = .systemFont(ofSize: UIFont.smallSystemFontSize)
+        if CTFontManagerRegisterGraphicsFont(fontRef, &errorRef) == false {
```
richherrera: **NIT:**
```suggestion
        if !CTFontManagerRegisterGraphicsFont(fontRef, &errorRef) {
```

richherrera (approved):
🚀

warmkesselj (approved):

PR #238, opened 2024-01-08
Vault Without Purchase 3DS - Cards

KunJeongPark:
## Summary of changes

- Modify SetupToken to take Card payment source with parameters for 3DS 
- Added `SCA` option in CreateSetupToken
- Add 3ds contingency check in vault function
- Add new function `startVaultThreeDSecureChallenge` to send result to vaultDelegate
- Add new test cards for testing 3DS for vault without pay
- Add unit tests for 3DS flow in vault function
- Add doc strings for vault function
- Make `CardVaultResult` public
- keep optional status field in CardVaultResult - nil for 3ds contingency flow where status is not updated in sdk
- Added `didAttemptThreeDSecureAuthentication` boolean property in `CardVaultResult` to distinguish 3ds contingency result
- Remove deepLinkURL property in `CardVaultResult`

No 3DS contingency success:

https://github.com/paypal/paypal-ios/assets/9273272/04aaf4ae-607d-48e7-b2b0-22036f02fb5b

3DS contingency success:

https://github.com/paypal/paypal-ios/assets/9273272/899ea725-8db1-4898-9c33-f15382a1562a

In another ticket, we will add option to do a get call in merchant server and based on status (APPROVED), 
user will be able to vault the card.

### Checklist

- [x] Added a changelog entry


jaxdesmarais:
I know we recently discussed some stuff around 3DS in standup. Since this is branched off of main I assume it is ready to be released once merged? If not, we should put this work in a feature branch along with the `liabilityShift` that exists on main.

scannillo:
Was also going to ask if this is ready to be reviewed? Agreed with Jax, if it's targeting `main` that means we're confident in releasing it (unless there are "DO NOT MERGE"  or Draft labels)

KunJeongPark:
I thought we could come to a decision about what we are relaying to merchants with this PR as point of discussion.
I wanted to start point of discussion. There is really not big change, just returning  a string value "3dsVerificationAttempted" for 3ds contingency path if no errors outside of verification was encountered. 
We need a clear way to communicate to merchant when they need get call on their end to check 3ds verification status.

For merchants not using 3ds, they still get the same results but with "APPROVED" status that they get from 
confirm-payment-source endpoint. 

For vault with purchase, all we have to do is revert the parsing of liabilityShift, if that decision is made. 
And we want to be consistent in status value there. "APPROVED"  or "3dsVerificationAttempted" if there were no errors encountered from SDK. 

What we have in main atm, is liabilityShift parsed and returned, error/success of 3ds discerned in SDK for vault with purchase and vault without purchase is in non public mode. We are not returning any status, even in cases where there is non 3ds contingency approval. My aim in this PR is to get agreement on right now value that we return to merchant so that we can be clear on specs, and align on ticket reverting liabilityShift value if we are doing that.

For now in main, we need clear consensus of what we are ok with for right now. What we have right now in main for vault with purchase was in question from our last meeting with web team. 

Did you mean revert the liabilityShift parsing feature and put that on 3ds branch?

Sources/CardPayments/Models/CardVaultResult.swift
```diff
@@ -4,12 +4,15 @@ import CorePayments
...
-    
-    /// status of the updated setup token
-    public let status: String
+
+    /// setupToken status
+    public let status: String?
+
+    /// This is the deep link url returned from 3DS authentication
+    @_documentation(visibility: private)
+    public let deepLinkURL: URL?
```
scannillo: I thought we decided against exposing this URL directly

Sources/CardPayments/CardClient.swift
```diff
@@ -154,6 +164,43 @@ public class CardClient: NSObject {
...
+                    case ASWebAuthenticationSessionError.canceledLogin:
+                        self.notifyVaultCancellation()
+                        return
+                    default:
+                        self.notifyVaultFailure(with: CardClientError.threeDSecureError(error))
+                        return
+                    }
+                }
+
+                let cardVaultResult = CardVaultResult(setupTokenID: setupTokenID, status: "3dsVerificationAttempted", deepLinkURL: url)
```
scannillo: `status` has a meaning in PayPal orders API world. It can be one of these values ([see docs](https://developer.paypal.com/docs/api/orders/v2/#definition-order)):
Screenshot 2024-02-13 at 5 01 35 PM

---

Because of this, I don't think our SDK should be defining any new `status` values as this could confuse merchants. Instead, I think the `CardVaultResult` could have a boolean property (like `threeDSecureAttempted`). Our docs and docstrings can callout that if that value is returned as true, a merchant must do the additional "get order info" API call on their server.

scannillo: * This boolean property would be in addition to `status`

KunJeongPark: That's true. I was just talking to Nirvan about that. The only thing is that status value is  unknown in our SDK after 3DS verification. That's why we've considered "unknown" or something else.

Sources/CardPayments/Models/CardVaultResult.swift
KunJeongPark: Yes, should we remove it altogether? We still have it in CardRequest.

Sources/CardPayments/CardClient.swift
KunJeongPark: So what should we do about status when updated status is not known after 3DS verification?
keep status optional returning nil value in 3ds contingency flow and return boolean true?

jaxdesmarais: > So what should we do about status when updated status is not known after 3DS verification?
keep it optional and return boolean true?

If we don't know the status at this point in time since it's not returned in the URL, we shouldn't return anything for status to the merchant. So yeah I think going with your approach of keeping it optional and returning a bool is the best we can do for merchants. Hopefully as time goes on we can get feedback from merchants on improving the experience and relay those to the API team.

KunJeongPark: That's a good call.

Sources/CardPayments/CardClient.swift
```diff
@@ -176,13 +176,13 @@ public class CardClient: NSObject {
             context: self,
             sessionDidDisplay: { [weak self] didDisplay in
                 if didDisplay {
-                    // should we send different analytics for vault 3DS?
+                    // TODO: analytics for card vault
                     self?.analyticsService?.sendEvent("card-payments:3ds:challenge-presentation:succeeded")
```
KunJeongPark, line 158: Should I send same 3DS analytics for vault without purchase flow? 
For other analytics events, should I do them here or create another ticket?

KunJeongPark: Will create another ticket for vault analytics

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
```diff
@@ -3,6 +3,7 @@ import SwiftUI
 struct CardVaultView: View {
 
     @StateObject var cardVaultViewModel = CardVaultViewModel()
+    @State var sca: String? = "SCA_WHEN_REQUIRED"
```
richherrera, line 6: Since we are assigning a value from the beginning, the variable does not need to be optional, what do you think?

Sources/CardPayments/CardClient.swift
scannillo: +1 to Jax's comment - if we don't have a `status` from the API we hit, we shouldn't set it

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
jaxdesmarais: We can likely just hardcode it below unless there is specific reason we need to extract it into a hardcoded state var. Thinking we can likely just do this below:
```
PaymentSourceType.card(verification: "SCA_WHEN_REQUIRED")
```

Demo/Demo/SwiftUIComponents/CommonComponents/CardFormView.swift
```diff
@@ -22,6 +22,12 @@ struct CardFormView: View {
                 FloatingLabelTextField(placeholder: "Card Number", text: $cardNumberText)
                     .onChange(of: cardNumberText) { newValue in
                         cardNumberText = cardFormatter.formatFieldWith(newValue, field: .cardNumber)
+                        // 4 digit cvv for amex
+                        if newValue.starts(with: "34") || newValue.starts(with: "37") {
+                            cvvText = "1234"
+                        } else {
+                            cvvText = "123"
+                        }
```
jaxdesmarais: We could instead fetch the card type and check that it's amex or not instead of [duplicating the logic from here](https://github.com/paypal/paypal-ios/blob/ecdb5cebee714c220f64d8e8a1b1077421b676ac/Demo/Demo/Card/CardType.swift#L35-L36). Could do:
```suggestion
cvvText = CardType.unknown.getCardType(newValue) == .americanExpress ? "1234" : "123"
```
That way if the values for Amex ever change we just need to update in 1 place. What do you think?

Demo/Demo/ViewModels/CardVaultViewModel.swift
```diff
@@ -31,10 +31,14 @@ class CardVaultViewModel: VaultViewModel, CardVaultDelegate {
         return enabled
     }
 
-    func setUpTokenSuccessResult(vaultResult: CardPayments.CardVaultResult) {
+    func setUpdateTokenSuccessResult(vaultResult: CardPayments.CardVaultResult) {
```
jaxdesmarais: this naming may clarify this method a bit more since the naming is confusing. What do you think:
```suggestion
    func updateSetupToken(vaultResult: CardPayments.CardVaultResult) {
```

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,29 @@ public class CardClient: NSObject {
...
         Task {
             do {
                 let result = try await vaultAPI.updateSetupToken(cardVaultRequest: vaultRequest).updateVaultSetupToken
-                
-                // TODO: handle 3DS contingency with helios link & add unit tests
-                if let link = result.links.first(where: { $0.rel == "approve" && $0.href.contains("helios") }) {
-                    let url = link.href
+
+                if result.status == "PAYER_ACTION_REQUIRED",
+                let url = result.links.first(where: { $0.rel == "approve" })?.href {
```
jaxdesmarais:
```suggestion
                let urlString = result.links.first(where: { $0.rel == "approve" })?.href {
```

Sources/CardPayments/CardClient.swift
```diff
@@ -154,6 +164,43 @@ public class CardClient: NSObject {
         return url.queryItems?.first { $0.name == param }?.value
     }
 
+    private func startVaultThreeDSecureChallenge(
+        url: URL,
+        setupTokenID: String
+    ) {
+
```
jaxdesmarais:
```suggestion
```

jaxdesmarais: Is there a way to use `startThreeDSecureChallenge` for both vault and non vault flow? It looks like the main difference is that 1 takes an order ID and the other a setupTokenID. I wonder if we can just switch on which of those values isn't nil and handle the differences otherwise in a singular method. Would that be possible?

Sources/CardPayments/CardVaultDelegate.swift
```diff
@@ -17,4 +17,19 @@ public protocol CardVaultDelegate: AnyObject {
     ///   - client: the CardClient associated with delegate
     ///   - didFinishWithError: the error returned by the Card vault flow
     func card(_ cardClient: CardClient, didFinishWithVaultError vaultError: CoreSDKError)
+
+    /// Notify that the Card vault flow has been cancelled
+    /// - Parameters:
+    ///   - client: the CardClient associated with delegate
+    func cardVaultDidCancel(_ cardClient: CardClient)
```
jaxdesmarais: Is there a way to more similarly match the naming patterns of these new delagate methods to the ones above?

Sources/CardPayments/Models/CardVaultResult.swift
```diff
@@ -4,12 +4,14 @@ import CorePayments
...
-@_documentation(visibility: private)
 public struct CardVaultResult {
     
     /// setupTokenID of token that was updated
     public let setupTokenID: String
-    
-    /// status of the updated setup token
-    public let status: String
+
+    /// setupToken status
```
jaxdesmarais:
```suggestion
    /// setup token status
```

Sources/CardPayments/CardClient.swift
KunJeongPark: I thought about that and main concern was returning results via different delegates.
I thought it's too confusing and error prone to toggle. And would we be sending different analytics for 3ds for two flows?

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: I thought it might be clearer to have on top, in case default values change. Easier to find.
Clarifying in a way that it is default, not randomly hard coded for some reason.

Sources/CardPayments/CardClient.swift
KunJeongPark: Yes, Steven brought this up as well. I agree. I was wanting to clarify that this state is not updated
after 3ds verification, so I thought of "UNKOWN" or something else. 
But setting it nil makes sense.

Demo/Demo/SwiftUIComponents/CommonComponents/CardFormView.swift
KunJeongPark: I didn't realize we had that!

Demo/Demo/ViewModels/CardVaultViewModel.swift
KunJeongPark: The function is just setting the state to update the view, so imo it's confusing to use "updateSetupToken" which is kind of name for our endpoint as well. And I like having wrapper functions for results, especially as we change the values we return to merchant.

Sources/CardPayments/CardClient.swift
jaxdesmarais: I guess my question is more, do we only need different result delegates, 1 for vault, 1 for non-vault. It seems like most of the other delegates: 3dsWillLaunch, cancellation, failure shouldn't matter if the merchant is vaulting or not. And with that, does it make more sense to just add 3ds data to 1 result reguardless of vault data? We could potentially add `VaultResult` to the `CardResult` and return extra data there if vaulted.

It just seems to me like a lot of this logic is separated out when it may not need to be since under the hood both the vault and non-vault flow do the same thing: take in a string, launch a URL, and return a result. Curious to hear your thoughts though!

Demo/Demo/ViewModels/CardVaultViewModel.swift
jaxdesmarais: Gotcha, since this and the error case are doing something similar, we could consider doing something like this:
```swift
    func updateSetupTokenResponseState(vaultResult: CardVaultResult?, vaultError: CoreSDKError?) {
        DispatchQueue.main.async {
            if let vaultError {
                self.state.updateSetupTokenResponse = .error(message: vaultError.localizedDescription)
            } else if let vaultResult {
                self.state.updateSetupTokenResponse = .loaded(
                    UpdateSetupTokenResult(
                        id: vaultResult.setupTokenID,
                        status: vaultResult.status,
                        threeDSecureAttempted: vaultResult.threeDSecureAttempted
                    )
                )
            }
        }
    }
```

Sources/CardPayments/CardVaultDelegate.swift
KunJeongPark: This one was because the other delegate function, there were clear ways to distinguish different delegate functions with the parameter: func card(_ cardClient: CardClient, `didFinishWithVaultResult` vaultResult: CardVaultResult)
I wanted to distinguish cancel for checkout flow vs cancel for vaulting.

Sources/CardPayments/CardClient.swift
KunJeongPark: I think we would need different delegates to return different results since they are different flows with different result values,  checkout flow vs non checkout flow. id in CardResult is orderID and id in CardVaultResult is setupTokenID. And they need different state properties for different steps in the app for different endpoints, for example.

Demo/Demo/ViewModels/CardVaultViewModel.swift
KunJeongPark: The delegate functions either have result or error, so I think it's clearer the way it is. Not strongly tied to it though.

Sources/CardPayments/CardClient.swift
richherrera: I think we should also take into consideration the analytics we should add, I think it is better to have different analytics depending on the flow, I agree that using a single function for both flows would be correct, however I also think it could be confusing and prone to errors, we could do an alaisis, even divide the CardClient (thinking out loud) for the future, what do you think? is there a good way of combining the two? any idea @jaxdesmarais?

jaxdesmarais: Ideally we'd send vault vs non vault in some sort of metadata parameter on our analytics. This would allow us to better compare a module's overall performance as well as drilling down into specific performance of vault vs non-vaulted flows. With sending separate event names for each flow, we end up needing to add events together to get a more holistic picture of a modules conversion which also leaves us prone to missing events.

It seems like as a whole in other modules we've drawn a pretty strict line around vault vs non-vault, so this could be something we consider tackling in v2 but overall I think in cases where we have the same code duplicated in 2 places with only ~5 lines of code different there is likely a way we can DRY up the codebase. Both are doing the exact same thing: take in a string, launch a URL, and return a result. Any time we are in a situation where we end up needing to not forget code exists in 2 places for launching URLs we could potentially open ourselves up for errors if we need to make updates. Not sure we have to tackle it in this PR but worth exploring.

jaxdesmarais: Curious to hear @scannillo thoughts on this too when she is back!

KunJeongPark: I mostly had in mind that vault without purchase does not involve checkout at all, thus a separate process.
I'd prefer to have a bit of repetition and clear separation of two flows.

I think you do have a good point. We can create a ticket to revisit this for later.

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,28 @@ public class CardClient: NSObject {
         self.vaultAPI = vaultAPI
         self.webAuthenticationSession = webAuthenticationSession
     }
-    
-    @_documentation(visibility: private)
+
+    /// Updates a setup token with a payment method. Performs
+    /// 3DS verification if required. Check verification status with  /v3/vault/setup-token/{id} in your server.
```
sshropshire:
```suggestion
    /// 3DS verification if required. Check verification status with /v3/vault/setup-token/{id} in your server.
```

sshropshire: > Check verification status with /v3/vault/setup-token/{id} in your server.

I am a bit concerned with this phrasing that merchants will think `vault()` causes the SDK to make a client-side request to `/v3/vault/setup-token/{id}`.

Sources/CardPayments/CardClient.swift
```diff
@@ -141,6 +150,42 @@ public class CardClient: NSObject {
         return url.queryItems?.first { $0.name == param }?.value
     }
 
+    private func startVaultThreeDSecureChallenge(
+        url: URL,
+        setupTokenID: String
+    ) {
```
sshropshire:
```suggestion
    private func startVaultThreeDSecureChallenge(url: URL, setupTokenID: String) {
```
🧶 : only if it's within the column limit. Take or leave, a bit clearer though.

Sources/CardPayments/CardVaultDelegate.swift
sshropshire: Maybe `cardDidCancelVault()` ?

KunJeongPark: I like that better. I was just referring to the flow CardVault, but it does read funny

Sources/CardPayments/CardClient.swift
KunJeongPark: Oh. I didn't think about that. How about  "If verification is performed, SDK returns a property didAttemptThreeDSecureAuthentication. If didAttempt3DSecureVerification is true, check verification status with  /v3/vault/setup-token/{id} in your server."?

sshropshire: Yeah we should make it as clear as we possibly can. Ultimately the developer.paypal docs will be the best place to explain the big-picture client / server interactions.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
jaxdesmarais: It looks like this value is already the default in `CreateSetupTokenView`. We shouldn't also need to pass it at all as a default here since it's just redundant, right?

Demo/Demo/ViewModels/CardVaultViewModel.swift
jaxdesmarais: Yes, but with both values optional in the helper method, we can pass down the appropriate data from the delegate. So the error delegate would send `updateSetupTokenResponseState(vaultResult: nil, vaultError: vaultError)` and the success delegate would send `updateSetupTokenResponseState(vaultResult: vaultResult, vaultError: nil)`.

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,29 @@ public class CardClient: NSObject {
...
-    
-    @_documentation(visibility: private)
+
+    /// Updates a setup token with a payment method. Performs
+    /// 3DS verification if required. If verification is performed, SDK returns a property didAttemptThreeDSecureAuthentication.
+    /// If didAttempt3DSecureVerification is true, check verification status with /v3/vault/setup-token/{id} in your server.
+    /// - Parameters:
+    ///   - cardVaultRequest: The request containing setupTokenID and card
+    /// - Returns: Card vault result
+    /// - Throws: PayPalSDK error if vault could not complete successfully
```
jaxdesmarais: It doesn't look like this method returns or throws anything, not sure if that was missed or if these docstrings are incorrect

KunJeongPark: I followed the pattern in `approveOrder` function. you are right. the functions don't return values,
they just call delegate functions.

jaxdesmarais: Seems like we should make a ticket to update the `approveOrder` docstring then. If the results are all returned in the delegate methods, we should remove the returns/throws docs from this method and instead note that merchants need to use the delagates for result handling.

KunJeongPark: I didn't know how to enter reply to your comment above so I'll put it down here: 
Jax: "It looks like this value is already the default in CreateSetupTokenView. We shouldn't also need to pass it at all as a default here since it's just redundant" 

Victoria: yes, if I don't set default here paymentSourceType will be initialized with verification field nil which
omits the verification_method in the setup token create call.
docs say SCA_WHEN_REQUIRED is the default, which I take to mean that in its absence, that is the value.
This was true on other endpoints, but I will ask on the vault channel just in case.

KunJeongPark: Jax: Yes, but with both values optional in the helper method, we can pass down the appropriate data from the delegate. So the error delegate would send updateSetupTokenResponseState(vaultResult: nil, vaultError: vaultError) and the success delegate would send updateSetupTokenResponseState(vaultResult: vaultResult, vaultError: nil).

Victoria: it is actually updateUpdateSetupTokenResult if I follow the convention of referring to result of endpoint.
I will use `setUpdateSetupTokenResult`

Sources/CardPayments/CardClient.swift
scannillo: It's a good practice to leave a TODO with the JIRA ticket # listed as a comment

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -5,6 +5,8 @@ import AuthenticationServices
 @testable import CardPayments
 @testable import TestShared
 
+// swiftlint:disable type_body_length
+// swiftlint:disable file_length
```
scannillo: I think we can do this: 
```suggestion
// swiftlint:disable type_body_length, file_length
```

Sources/CardPayments/CardVaultDelegate.swift
```diff
@@ -17,4 +17,19 @@ public protocol CardVaultDelegate: AnyObject {
     ///   - client: the CardClient associated with delegate
     ///   - didFinishWithError: the error returned by the Card vault flow
     func card(_ cardClient: CardClient, didFinishWithVaultError vaultError: CoreSDKError)
+
+    /// Notify that the Card vault flow has been cancelled
+    /// - Parameters:
+    ///   - client: the CardClient associated with delegate
+    func cardDidCancelVault(_ cardClient: CardClient)
```
scannillo: This method will only get called if a 3DS flow was cancelled. To match the naming convention below, it should include `threeDSecure` in the name IMO

scannillo: Also it doesn't need `Vault` in it, since that's already in the class name

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
KunJeongPark: If I don't set default here, paymentSourceType with card will be initialized with verification field nil (until picker is toggled) which omits the verification_method in the setup token create call.
Docs say SCA_WHEN_REQUIRED is the default, which I take to mean that in its absence, that is the value.

But that assignment of sca on child view right now is not automatically initializing the paymentType's verification_method property with SCA_WHEN_REQUIRED. I like the idea of initializing the default value from parent view since parent view can decide that this is PayPal payment type also.

UnitTests/CardPaymentsTests/CardClient_Tests.swift
KunJeongPark: It is strange, I still got error body length when I applied your suggestion.

jaxdesmarais: I don't think you need the comma, should work with `// swiftlint:disable type_body_length file_length`

KunJeongPark: Thanks Jax!

Sources/CardPayments/CardVaultDelegate.swift
KunJeongPark: I guess it should be `cardThreeDSecureDidCancel`

sshropshire: Not `cardDidCancelThreeDSecure`?

KunJeongPark: 🤣 I just changed it from that. because other 3ds functions are `cardThreeDSecureDidFinish`, etc

KunJeongPark: So I think Sammy's point was that this cancel path doesn't exist for non 3DS contingency flow, meaning it goes with other cardThreeDSecure... functions

sshropshire: Oh got it yeah if there's already a convention in place it's best to match that.

Objectively though it has somewhat of a yoda vibe just callin it out 🤷 .

KunJeongPark: I agree with you, that's why I did what you suggested earlier, until I saw the other functions.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 ## unreleased
 * Breaking Changes
   * CardPayments
+    * Add `status` property to `CardVaultResult`
+    * Add `didAttemptThreeDSecureAuthentication` property to `CardVaultResult`
```
jaxdesmarais: We shouldn't have a "breaking changes" section on `main` as there should only be breaking changes on `next-major-version`.

Demo/Demo/SwiftUIComponents/CardVaultViews/CardVaultView.swift
jaxdesmarais: > Docs say SCA_WHEN_REQUIRED is the default, which I take to mean that in its absence, that is the value.

This is correct, if not provided based on what you said it sounds like it should default to `SCA_WHEN_REQUIRED`

> But that assignment of sca on child view right now is not automatically initializing the paymentType's verification_method property with SCA_WHEN_REQUIRED. I like the idea of initializing the default value from parent view since parent view can decide that this is PayPal payment type also.

I agree that only one place should know about this initialization vs having to pass it around. This sounds like some tech debt that we should address in a follow up PR. We shouldn't need to provide a default in multiple places for our demo app to behave as expected.

Demo/Demo/SwiftUIComponents/VaultViews/CreateSetupTokenView.swift
```diff
@@ -34,8 +44,8 @@ struct CreateSetupTokenView: View {
                                 selectedMerchantIntegration: selectedMerchantIntegration,
                                 paymentSourceType: paymentSourceType
                             )
-                        } catch {
-                            print("Error in getting setup token. \(error.localizedDescription)")
+                    } catch {
+                        print("Error in getting setup token. \(error.localizedDescription)")
```
jaxdesmarais: This spacing update should be reverted, the spacing was correct before since this catch should like up with the `do` above.

Demo/Demo/ViewModels/CardVaultViewModel.swift
```diff
@@ -31,29 +31,46 @@ class CardVaultViewModel: VaultViewModel, CardVaultDelegate {
         return enabled
     }
 
-    func setUpTokenSuccessResult(vaultResult: CardPayments.CardVaultResult) {
+    func setUpdateSetupTokenResult(vaultResult: CardPayments.CardVaultResult? = nil, vaultError: CorePayments.CoreSDKError? = nil) {
```
jaxdesmarais:
```suggestion
    func setUpdateSetupTokenResult(vaultResult: CardVaultResult? = nil, vaultError: CoreSDKError? = nil) {
```

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,27 @@ public class CardClient: NSObject {
         self.vaultAPI = vaultAPI
         self.webAuthenticationSession = webAuthenticationSession
     }
-    
-    @_documentation(visibility: private)
+
+    /// Updates a setup token with a payment method. Performs
+    /// 3DS verification if required. If verification is performed, SDK returns a property didAttemptThreeDSecureAuthentication.
+    /// If didAttempt3DSecureVerification is true, check verification status with /v3/vault/setup-token/{id} in your server.
```
jaxdesmarais: nit to cleanup docstrings:
```suggestion
    /// If `didAttempt3DSecureVerification` is `true`, check verification status with `/v3/vault/setup-token/{id}` in your server.
```

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,27 @@ public class CardClient: NSObject {
...
+    /// - Parameters:
+    ///   - cardVaultRequest: The request containing setupTokenID and card
```
jaxdesmarais:
```suggestion
    ///   - vaultRequest: The request containing setupTokenID and card
```

Sources/CardPayments/CardClient.swift
```diff
@@ -37,19 +37,27 @@ public class CardClient: NSObject {
...
-                
-                // TODO: handle 3DS contingency with helios link & add unit tests
-                if let link = result.links.first(where: { $0.rel == "approve" && $0.href.contains("helios") }) {
-                    let url = link.href
-                    print("3DS url \(url)")
+
+                if result.status == "PAYER_ACTION_REQUIRED",
+                let urlString = result.links.first(where: { $0.rel == "approve" })?.href {
+                    guard urlString.contains("helios"),
+                        let url = URL(string: urlString) else {
```
jaxdesmarais: This seems short enough to single line:
```suggestion
                    guard urlString.contains("helios"), let url = URL(string: urlString) else {
```

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -111,6 +164,101 @@ class CardClient_Tests: XCTestCase {
...
+                XCTAssertEqual(error.code, CardClientError.Code.threeDSecureError.rawValue)
+                XCTAssertEqual(error.localizedDescription, "Mock web session error description.")
+                expectation.fulfill()
+            },
+            cancel: { _ in
+                XCTFail("Invoked cancel() callback. Should invoke error().")
+                expectation.fulfill()
+            },
+            threeDSWillLaunch: { _ in XCTAssert(true) },
+            threeDSLaunched: { _ in XCTAssert(true) })
```
jaxdesmarais:
```suggestion
            threeDSLaunched: { _ in XCTAssert(true) }
        )
```

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -111,6 +164,101 @@ class CardClient_Tests: XCTestCase {
...
+            error: { _, error in
+                XCTFail(error.localizedDescription)
+                expectation.fulfill()
+            },
+            cancel: { _ in
+                XCTAssert(true)
+                expectation.fulfill()
+            },
+            threeDSWillLaunch: { _ in XCTAssert(true) },
+            threeDSLaunched: { _ in XCTAssert(true) })
```
jaxdesmarais:
```suggestion
            threeDSLaunched: { _ in XCTAssert(true) }
        )
```

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -111,6 +164,101 @@ class CardClient_Tests: XCTestCase {
...
+                XCTAssertTrue(result.didAttemptThreeDSecureAuthentication)
+                expectation.fulfill()
+            },
+            error: { _, error in
+                XCTFail(error.localizedDescription)
+                expectation.fulfill()
+            },
+            cancel: { _ in XCTFail("Invoked cancel() callback. Should invoke success().") },
+            threeDSWillLaunch: { _ -> Void in XCTAssert(true) },
+            threeDSLaunched: { _ -> Void in XCTAssert(true) })
```
jaxdesmarais:
```suggestion
            threeDSLaunched: { _ -> Void in XCTAssert(true) }
        )
```

CHANGELOG.md
scannillo: +1: also adding classes & properties is not breaking, but rather a minor version bump (semver.org)

CHANGELOG.md
```diff
@@ -12,6 +14,10 @@
   * Add `custom` case for `PaymentButtonEdges`
   * Support VoiceOver by adding button accessibility labels
   * Font typeface changed to "PayPalOpen" to meet brand guidelines
+* CardPayments
```
scannillo, line 11: We already have a `CardPayments` section above, let's consolidate

KunJeongPark: I think we've always separated breaking changes vs new features

CHANGELOG.md
KunJeongPark: addressed in 35aa9df5e5749a3dd5f9ca84af3a3bab8e6da325

CHANGELOG.md
KunJeongPark: addressed in 35aa9df5e5749a3dd5f9ca84af3a3bab8e6da325

Demo/Demo/ViewModels/CardVaultViewModel.swift
KunJeongPark: I was following convention of other delegate functions such as ` func card(_ cardClient: CardPayments.CardClient, didFinishWithResult result: CardPayments.CardResult)`

jaxdesmarais: The delegates shouldn't need to follow that pattern either

KunJeongPark: I've applied your suggestion

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
```diff
@@ -6,6 +6,11 @@ struct UpdateSetupTokenView: View {
 
     let setupToken: String
     let cardSections: [CardSection] = [
+        // source: https://developer.paypal.com/api/rest/sandbox/card-testing/#link-testcardnumbers
+        CardSection(title: "Successful Step Up Authentication - Visa", numbers: ["4005 5192 0000 0004"]),
+        // source: https://developer.paypal.com/api/rest/sandbox/card-testing/#link-testcardnumbers
```
sshropshire, line 11:
```suggestion
```
nit: duplicate `source` link.

Demo/Demo/SwiftUIComponents/VaultViews/CreateSetupTokenView.swift
```diff
@@ -25,6 +26,15 @@ struct CreateSetupTokenView: View {
             .frame(maxWidth: .infinity)
             .font(.headline)
             FloatingLabelTextField(placeholder: "Vault Customer ID (Optional)", text: $vaultCustomerID)
+            if case .card = paymentSourceType {
```
sshropshire, line 29: random: I can see that this is essentially `paymentSourceType == .card` but the Swift syntax for associated values is 🤯

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: oh! Good catch

Demo/Demo/SwiftUIComponents/VaultViews/CreateSetupTokenView.swift
KunJeongPark: I know. I am not fan of this one, it's weird.

sshropshire (approved):

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: Oh I remember now. I was trying to document where each card was coming from since some of them I got from 3DS team.

jaxdesmarais (approved):
Thanks as always for addressing my random questions/thoughts!

Demo/Demo/SwiftUIComponents/CardVaultViews/UpdateSetupTokenView.swift
KunJeongPark: Some of the cards below were working before until vault team added validation for vault cards, so some cards that were configured for 3DS are not working for vaulting. I put in a request to vault team to get the ones below top two (which are working) validated for vaulting

sshropshire: It's a take or leave 🧶 . If it helps us out in the future then it's good to keep.

KunJeongPark: More a note to myself. So I know to keep circling back on that request to vault team.

PR #243, opened 2024-01-16
Removed Buttons colors for V2

stechiu:
### Reason for changes

We will remove the deprecated colors in V2

### Summary of changes

- Removed `.black`, `.silver`, `.blue`, and `.darkBlue` as per QL button redesign requirements

### Checklist

- [x] Added a changelog entry

- @stechiu


UnitTests/PaymentButtonsTests/PayPalCreditButton_Tests.swift
```diff
@@ -14,7 +14,7 @@ class PayPalCreditButton_Tests: XCTestCase {
         let sut = PayPalCreditButton()
         XCTAssertEqual(sut.edges, PaymentButtonEdges.softEdges)
         XCTAssertEqual(sut.size, PaymentButtonSize.collapsed)
-        XCTAssertEqual(sut.color, PaymentButtonColor.darkBlue)
+        XCTAssertEqual(sut.color, PaymentButtonColor.white)
```
scannillo: Just confirming, should white or gold be the default color?

stechiu: Good point, I just checked Figma and gold is the recommended/default color. Will make that update :)

warmkesselj:
@stechiu does this enum embedded in PayPalButton need updating? 

```
    /// Available colors for PayPalButton.
    public enum Color: String {
        case gold
        case white
        case black
        case silver
        case blue

        var color: PaymentButtonColor {
            PaymentButtonColor(rawValue: rawValue) ?? .gold
        }
    }

**Comment by stechiu** [comment]
> @stechiu does this enum embedded in PayPalButton need updating?
> 
> ```
>     /// Available colors for PayPalButton.
>     public enum Color: String {
>         case gold
>         case white
>         case black
>         case silver
>         case blue
> 
>         var color: PaymentButtonColor {
>             PaymentButtonColor(rawValue: rawValue) ?? .gold
>         }
>     }
>     
> ```

Missed this one, thanks!

File: Sources/PaymentButtons/PayPalCreditButton.swift [hunk 2]
```diff
@@ -26,7 +24,7 @@ public final class PayPalCreditButton: PaymentButton {
     ///   - size: Size of the button. Default to collapsed if not provided.
     public convenience init(
         insets: NSDirectionalEdgeInsets? = nil,
-        color: Color = .darkBlue,
+        color: Color = .gold,
```
scannillo, line 27: Looks like the docstring on line 22 also needs to be updated to `gold` vs `dark blue`

Sources/PaymentButtons/PayPalCreditButton.swift
```diff
@@ -59,7 +57,7 @@ public extension PayPalCreditButton {
         ///   - size: Size of the button. Default to collapsed if not provided.
         public init(
             insets: NSDirectionalEdgeInsets? = nil,
-            color: PayPalCreditButton.Color = .darkBlue,
+            color: PayPalCreditButton.Color = .gold,
```
scannillo, line 60: Same comment re: docstring

jaxdesmarais:
It looks like we will also need to clean up the demo app with these changes. See: https://github.com/paypal/paypal-ios/blob/157712a6b10849ac105acdac0e24be995211657a/Demo/Demo/Extensions/PaymentButtonEnums%2BExtension.swift#L14

Sources/PaymentButtons/PaymentButton+ImageAsset.swift
```diff
@@ -34,14 +34,8 @@ extension PaymentButton {
         }
 
         switch color {
-        case .gold, .white, .silver:
```
richherrera, line 37: From what I see, the switch seems unnecessary. I think we can remove it. What do you think?

Sources/PaymentButtons/PaymentButtonColor.swift
```diff
@@ -8,47 +8,20 @@ public enum PaymentButtonColor: String {
...
-            return UIColor(hexString: "#0070BA")
-
-        case .darkBlue:
-            return UIColor(hexString: "#073990")
         }
     }
 
     var fontColor: UIColor {
         switch self {
-        case .gold, .white, .silver:
```
richherrera, line 47: Same here, I think we can save a few lines of code 😄 by just returning `.black`. What do you think?

richherrera (approved):
Beside the feedback from the rest of the team, it LGTM!!!

Sources/PaymentButtons/PayPalCreditButton.swift
stechiu: Good catch, thanks!

Sources/PaymentButtons/PaymentButton+ImageAsset.swift
stechiu: Good point thanks!

Sources/PaymentButtons/PaymentButtonColor.swift
```diff
@@ -8,48 +8,18 @@ public enum PaymentButtonColor: String {
...
 
     var fontColor: UIColor {
-        switch self {
-        case .gold, .white, .silver:
-            return .black
-
-        case .blue, .black, .darkBlue:
-            return .white
-        }
+        return .black
```
jaxdesmarais: We can remove this `return` and rely on implicit returns since this is the only line in the block

warmkesselj (approved):
LGTM

PR #247, opened 2024-02-02
Payment Button Sizing Bug

stechiu:
### Reason for changes

There was a sizing issue with the Payment Button when used in SwiftUI apps
- The button couldn't be horizontally centered
- The button frame took up the entirety of the screen size rather than hugging the button
- If a button was added to a Form Section, the button wouldn't resize to fit the Section dimensions; instead, the size remained static

### Summary of changes

- This PR fixes the above issues
- A dev won't be able to set the height to smaller than 38px or width shorter than 300px
- PayLater and Credit are also displayed correctly in SwiftUI
- The `mini` button logo wasn't the correct size before, so that has been fixed

| Bug | PR |
| ----------- | ----------- |
| Screenshot 2024-02-01 at 8 08 06 PM | Screenshot 2024-02-01 at 8 18 58 PM |

### Checklist

- [x] Added a changelog entry

- @stechiu , @nirvan543

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,6 @@
     * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
     * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
     * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
-    * Rename `PayPalButton.Representable` to `PayPalButtonView`, `PayPalPayLaterButton.Representable` to `PayPalPayLaterButtonView`, and `PayPalCreditButton.Representable` to `PayPalCreditButtonView` for SwiftUI
 
 ## 1.1.0 (2023-11-16)
 * PayPalNativePayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,6 @@
 # PayPal iOS SDK Release Notes
 
 ## unreleased
-* PaymentButtons
-  * Add `custom` case for `PaymentButtonEdges`
-  * Support VoiceOver by adding button accessibility labels
-  * Add `custom` case for `PaymentButtonEdges`
-  * Support VoiceOver by adding button accessibility labels
-  * Font typeface changed to "PayPalOpen" to meet brand guidelines
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Add `CardClientError.threeDSVerificationError` for invalid verification
@@ -18,6 +12,18 @@
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
+* PaymentButtons
+  * Add new `PaymentButtonSize` case `.standard`
+  * Add `custom` case for `PaymentButtonEdges`
+  * Support VoiceOver by adding button accessibility labels
+  * Add new `PaymentButtonSize` case `.miniWithWordmark`
+  * `PayPalButtonSize` `.mini` changed to rectangular button to meet brand guidelines
+  * Font typeface changed to "PayPalOpen" to meet brand guidelines
+* Breaking Changes
+  * PaymentButtons
+    * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
+    * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
+    * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
 
 ## 1.1.0 (2023-11-16)
 * PayPalNativePayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 # PayPal iOS SDK Release Notes
 
 ## unreleased
+* PaymentButtons
+  * Add `custom` case for `PaymentButtonEdges`
+  * Support VoiceOver by adding button accessibility labels
+  * Add `custom` case for `PaymentButtonEdges`
+  * Support VoiceOver by adding button accessibility labels
+  * Font typeface changed to "PayPalOpen" to meet brand guidelines
 * CardPayments
   * Add `liabilityShift` property to `CardResult`
   * Add `CardClientError.threeDSVerificationError` for invalid verification
@@ -12,23 +18,6 @@
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
-* PaymentButtons
-  * Add new `PaymentButtonSize` case `.standard`
-  * Add `custom` case for `PaymentButtonEdges`
-  * Support VoiceOver by adding button accessibility labels
-  * Add new `PaymentButtonSize` case `.miniWithWordmark`
-  * Add new `PayPalButton.Labels` cases: `.addMoneyWith`, `.bookWith`, `.buyWith`, `.buyNowWith`,
-  `.checkoutWith`, `.continueWith`, `.contributeWith`, `.orderWith`, `.payLaterWith`, `.payWith`,
-  `.reloadWith`, `.rentWith`, `.reserveWith`, `.subscribeWith`, `.supportWith`, `.tipWith`,
-  `.topUpWith`
-  * `PayPalButtonSize` `.mini` changed to rectangular button to meet brand guidelines
-  * Font typeface changed to "PayPalOpen" to meet brand guidelines
-* Breaking Changes
-  * PaymentButtons
-    * Remove `PayPalButton.Labels` cases: `.checkout`, `.buyNow`, `.payLater`
-    * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
-    * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
-    * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
 
 ## 1.1.0 (2023-11-16)
 * PayPalNativePayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
     * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
     * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
     * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
+    * Update `PayPalButton` to `PayPalButtonView`, `PayPalPayLaterButton` to `PayPalPayLaterButtonView`, and `PayPalCreditButton` to `PayPalCreditButtonView` in SwiftUI files
 
 ## 1.1.0 (2023-11-16)
 * PayPalNativePayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@
     * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
     * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
     * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
-    * Update `PayPalButton` to `PayPalButtonView`, `PayPalPayLaterButton` to `PayPalPayLaterButtonView`, and `PayPalCreditButton` to `PayPalCreditButtonView` in SwiftUI files
+    * Rename `PayPalButton.Representable` to `PayPalButtonView`, `PayPalPayLaterButton.Representable` to `PayPalPayLaterButtonView`, and `PayPalCreditButton.Representable` to `PayPalCreditButtonView` for SwiftUI
 
 ## 1.1.0 (2023-11-16)
 * PayPalNativePayments
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -36,7 +36,7 @@ extension PayPalCreditButton.Color: CaseIterable {
 extension PaymentButtonEdges: CaseIterable {
 
     public static var allCases: [PaymentButtonEdges] {
-        [.rectangle, .rounded, .rounded, .custom(10)]
+        [.hardEdges, .rounded, .rounded, .custom(10)]
     }
 
     static func allCasesAsString() -> [String] {
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -36,7 +36,7 @@ extension PayPalCreditButton.Color: CaseIterable {
 extension PaymentButtonEdges: CaseIterable {
 
     public static var allCases: [PaymentButtonEdges] {
-        [.hardEdges, .rounded, .rounded, .custom(10)]
+        [.hardEdges, .softEdges, .rounded, .custom(10)]
     }
 
     static func allCasesAsString() -> [String] {
--- a/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
+++ b/Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
@@ -3,7 +3,7 @@ import PaymentButtons
 extension PayPalPayLaterButton.Color {
 
     public static var allCases: [PayPalPayLaterButton.Color] {
-        [.gold, .white]
+        [.gold, .white, .black, .silver, .blue]
     }
 
     static func allCasesAsString() -> [String] {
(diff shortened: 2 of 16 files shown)


Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -113,69 +113,65 @@ public final class PayPalButton: PaymentButton {
...
+
+    public func updateUIView(_ uiView: PaymentButton, context: Context) {
+        context.coordinator.action = action
     }
 }
 
 // MARK: PayPalButton Preview
 
-struct PayPalButtonView: View {
+struct PayPalButtonUIView: View {
```
nirvan543, line 170: Do we need this view? I see that we are referencing `PayPalCreditButtonView` in demo app. Same question for the Credit and Pay Later buttons.

nirvan543: Do we need this view? I see that we are using `PayPalCreditButtonView` in the demo app. Same question for the Credit and Pay Later buttons.

Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -113,69 +113,65 @@ public final class PayPalButton: PaymentButton {
...
 
     var body: some View {
-        PayPalButton.Representable()
+        PayPalButtonView()
+            .fixedSize(horizontal: false, vertical: true)
```
nirvan543: I noticed that `PayPalPayLaterButtonUIView` and `PayPalCreditButtonUIView` don't call `.fixedSize(...)`. Do we need it here or do the other views also need to call `.fixedSize(...)`?

Sources/PaymentButtons/PayPalButton.swift
nirvan543: Follow-up: Correct me if I'm wrong, but I believe we saw instances where we needed to call `.fixedSize(...)` in the demo app. If that's the case, should we use this wrapper view in the demo app instead of the `PayPalButtonView` so as to eliminate the merchant's need to call `.fixedSize(...)` themselves?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
     * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
     * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
     * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
+    * Update `PayPalButton` to `PayPalButtonView`, `PayPalPayLaterButton` to `PayPalPayLaterButtonView`, and `PayPalCreditButton` to `PayPalCreditButtonView` in SwiftUI files
```
jaxdesmarais:
```suggestion
    * Rename `PayPalButton.Representable` to `PayPalButtonView`, `PayPalPayLaterButton.Representable` to `PayPalPayLaterButtonView`, and `PayPalCreditButton.Representable` to `PayPalCreditButtonView` for SwiftUI
```

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebButtonsView.swift
```diff
@@ -27,15 +27,15 @@ struct PayPalWebButtonsView: View {
 
                 switch selectedFundingSource {
                 case .paypalCredit:
-                    PayPalCreditButton.Representable(color: .gold, size: .standard) {
-                        payPalWebViewModel.paymentButtonTapped(funding: .paypalCredit)
+                    PayPalCreditButtonView(color: .gold, edges: .rounded, size: .standard) {
+                        payPalWebViewModel.paymentButtonTapped(funding: .paylater)
```
jaxdesmarais: It looks like the name of this button is `PayPalCreditButtonView` so I think this is supposed to be `. paypalCredit`
```suggestion
                        payPalWebViewModel.paymentButtonTapped(funding: . paypalCredit)
```

Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -113,69 +113,65 @@ public final class PayPalButton: PaymentButton {
     deinit {}
 }
 
-public extension PayPalButton {
```
jaxdesmarais, line 63: I am not opposed to this change if we decide on a better name, but it seems like using `Representable` is a common practice for SDKs that offer both SwiftUI and UIKit. For example, here is Stripe's example SwiftUI view for their card form: https://github.com/stripe/stripe-ios/blob/392f6d08fcb4f075c5f36a376c14477e7087a6e3/Example/UI%20Examples/UI%20Examples/Source/SwiftUICardFormView.swift#L19.

Not saying that we _have_ to go with this option but this is why we originally went with using `Representable`.

jaxdesmarais (commented):
We should also add the min/max sizes to our public docstrings for the buttons.

Sources/PaymentButtons/PayPalButton.swift
nirvan543: I personally have not seen `.Representable` used before. I usually expect a view to be self-contained rather than being defined within a class. I'm curious if there are more examples of this pattern. I have not seen any aside from our's and the Stripe example linked above.

jaxdesmarais: There aren't a ton of SDKs that offer both a SwiftUI and UIKit UI element out of the box. When we originally went with using `Representable` we attempted to find other examples but Stripe was the only one we were able to find (open source and not just a question on StackOverflow) that offered 1 elements for both UIKit and SwiftUI.

Apple Pay _just_ added a button for SwiftUI with iOS 16+ (`PayWithApplePayButton` - SwiftUI and `PKPaymentRequest` - UIKit). This could be the pattern that we move forward with but at the time we went with `Representable` only 1 SDK (Stripe) offered UIKit + SwiftUI elements at all, so we went forward with the approach that existed at the time.

We also originally attempted to subclass the main button class with `UIViewRepresentable`, but there was apparently a bug that allowed us to do this originally (PR #84) for context to fix this and move to `Representable`. That to say, I am not opposed to whatever approach we decide but at the time of writing this code originally we had very few examples to base our code on. 😄 

Do you have any examples of an SDK using 1 namespace for both their UIKit and SwiftUI elements? It certainly _could_ be possible again but basing my knowledge off of the work done over a year and half ago when SwiftUI was still quite new.

Sources/PaymentButtons/PayPalButton.swift
stechiu: The view was here before and is used so we can preview the button element on its own, separate from when it's used with other elements in `SwiftUIPaymentButtonDemo`. In terms of if we need it, technically we don't since we have the demo app to see the button.

afaik, we can't call `PayPalButtonUIView` in `SwiftUIPaymentButtonDemo` because the latter is expecting a UIViewRepresentable and now a View. I did try to do so just now and I'm not able to add `PayPalButtonUIView`. 

I think Merchant devs would know to used `.fixedSize()` since it's a common issue in SwiftUI, but in case they don't, I can add that to the dev doc sample code. What do you think?

CHANGELOG.md
stechiu: Thanks!

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -36,7 +36,7 @@ extension PayPalCreditButton.Color: CaseIterable {
 extension PaymentButtonEdges: CaseIterable {
 
     public static var allCases: [PaymentButtonEdges] {
-        [.hardEdges, .softEdges, .rounded, .custom(10)]
+        [.hardEdges, .rounded, .rounded, .custom(10)]
```
richherrera: .rounded is duplicated, is this intentional?

Sources/PaymentButtons/PayPalButton.swift
sshropshire: `Representable` has worked well for us up to this point, but I do think it makes sense to revisit this until it feels 100% right.

SwiftUI and other declarative UI frameworks put a premium on readability. I'd vote for reserving the simplest name for SwiftUI like Apple does with `PayWithApplePayButton` <-> `PayPalButton`.

I also like the PK (Payment Kit) module prefix they use for the UIKit side of things. Maybe we could do something similar like `PPPayPalButton` or `PYPayPalButton`?

nirvan543: @jaxdesmarais Absolutely that makes sense and there is nothing wrong with the `.Representable` approach. However, I feel if we can come up with a better name to delineate the two, I think that can be considered best practice in today's SwiftUI environment.

I like the idea @sshropshire suggested. For the UIView, we can perhaps prefix the class names (with a non-Apple prefix) and leave the SwiftUI view with no prefix. For example:

UIKit: `PPPayPalButton` (or something similar)
SwiftUI: `PayPalButton`

I'd like to hear @scannillo's thoughts on the prefix idea because she did mention this:

https://github.com/paypal/paypal-ios/pull/84#issuecomment-1291422816

jaxdesmarais: Yep, like I said above, I am by no means saying using `Representable` is the right approach now. Just that it _was_ the right approach at the time this code was written. Since there was some debate on the initial "why" I was just attempting to provide context from that point in time since most of the current team wasn't around when that decision was made. :)

sshropshire: It's cool yeah I remember us choosing that because of Stripe.

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebButtonsView.swift
```diff
@@ -27,15 +27,15 @@ struct PayPalWebButtonsView: View {
...
                         payPalWebViewModel.paymentButtonTapped(funding: .paypalCredit)
                     }
                 case .paylater:
-                    PayPalPayLaterButton.Representable(color: .gold, edges: .softEdges, size: .standard) {
+                    PayPalPayLaterButtonView(color: .gold, edges: .rounded, size: .standard) {
                         payPalWebViewModel.paymentButtonTapped(funding: .paylater)
                     }
                 case .paypal:
-                    PayPalButton.Representable(color: .gold, size: .standard) {
+                    PayPalButtonView(color: .gold, size: .standard) {
```
jaxdesmarais: Do we want to add `edges: .rounded` here to match the above buttons?

Demo/Demo/SwiftUIComponents/SwiftUIPaymentButtonDemo.swift
```diff
@@ -109,25 +109,25 @@ struct SwiftUIPaymentButtonDemo: View {
...
-                    PayPalButton.Representable(
+                    PayPalButtonView(
                         color: PayPalButton.Color.allCases[colorsIndex],
                         edges: selectedEdge,
                         size: selectedSize,
                         label: selectedLabel
                     )
                     .id(buttonID)
-                    .frame(maxWidth: .infinity)
+                    .frame(width: 300)
```
jaxdesmarais, line 119: Should we add this frame to the other buttons as well?

Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -113,69 +113,64 @@ public final class PayPalButton: PaymentButton {
...
+            insets: insets,
+            label: label?.label
+        )
+        self.action = action
+    }
 
-        public func makeUIView(context: Context) -> UIView {
-            let view = UIView()
+    // MARK: - UIViewRepresentable methods
+    // TODO: Make unit test for UIVRepresentable methods: https://engineering.paypalcorp.com/jira/browse/DTNOR-623
```
jaxdesmarais: We don't want links on public GitHub to internal resources:
```suggestion
    // TODO: Make unit test for UIVRepresentable methods: ticket DTNOR-623
```

Sources/PaymentButtons/PayPalPayLaterButton.swift
```diff
@@ -38,67 +38,61 @@ public final class PayPalPayLaterButton: PaymentButton {
...
+            insets: insets,
+            label: .payLater
+        )
+        self.action = action
+    }
 
-            return view
-        }
+    // MARK: - UIViewRepresentable methods
+    // TODO: Make unit test for UIVRepresentable methods: https://engineering.paypalcorp.com/jira/browse/DTNOR-623
```
jaxdesmarais: Same here:
```suggestion
    // TODO: Make unit test for UIVRepresentable methods: ticket DTNOR-623
```

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
stechiu: hmm yea I'm not sure why `.rounded` shows up twice. Will update

Demo/Demo/SwiftUIComponents/PayPalWebPayments/PayPalWebButtonsView.swift
stechiu: I'll add it in

Sources/PaymentButtons/PayPalButton.swift
stechiu: This looks like a bad merge... I didn't type that and the link leads to a 404 error. Will remove

Sources/PaymentButtons/PayPalPayLaterButton.swift
stechiu: This looks like a bad merge... I didn't type that and the link leads to a 404 error. Will remove

Sources/PaymentButtons/PayPalButton.swift
stechiu: Seems like we don't need it here since it doesn't effect the parent view `SwiftUIPaymentButtonDemo`

Sources/PaymentButtons/PayPalButton.swift
nirvan543: Oh I see. This is the pre-preview struct. Can we rename these preview structs to indicate they they are meant to be for the preview? Also, can we move the struct inside the preview struct as well?

Example:

```
struct PayPalButtonView_Preview: PreviewProvider {

    static var previews: some View {
        PayPalButtonView()
    }
    
    private struct PayPalButtonWrapper: View {
        var body: some View {
            PayPalButton.Representable()
        }
    }
}
```

Sources/PaymentButtons/PayPalButton.swift
nirvan543: I'm open to revisiting this for when we have an idea on a name. For now, the way it works is perfectly fine.

stechiu: Same, I'll revert it back to `.representable` and start a conversation on Slack since what we decide will impact other parts of the codebase where we build the UI in UIKit and wrap it to display in SwiftUI view

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
     * Remove `PaymentButtonColor` `.black`, `.silver`, `.blue`, and `.darkBlue`
     * Add `PaymentButtonColor.gold` for `PayPalCreditButton`
     * Remove `PaymentButtonSize` cases: `.miniWithWordmark`, `.collapsed`, `.expanded`, `.full`
+    * Rename `PayPalButton.Representable` to `PayPalButtonView`, `PayPalPayLaterButton.Representable` to `PayPalPayLaterButtonView`, and `PayPalCreditButton.Representable` to `PayPalCreditButtonView` for SwiftUI
```
jaxdesmarais: Not sure if this change should be reverted and done in a separate PR based on this comment: https://github.com/paypal/paypal-ios/pull/247#discussion_r1484672567

stechiu: I'll be adding in `.Representable` again so this log will be removed

Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -132,7 +132,7 @@ public extension PayPalButton {
         public init(
             insets: NSDirectionalEdgeInsets? = nil,
             color: PayPalButton.Color = .gold,
-            edges: PaymentButtonEdges = .softEdges,
+            edges: PaymentButtonEdges = .rounded,
```
jaxdesmarais: Lets update all of the docstrings for the inits to note the new default

jaxdesmarais (commented):
Just a small comment about updating our docstrings in all 3 inits to note the default in the inits. Also, do we need to make any changes on main to fix the sizing issue for currently released versions of the SDK since this change is going into next major version?

stechiu:
> Just a small comment about updating our docstrings in all 3 inits to note the default in the inits. Also, do we need to make any changes on main to fix the sizing issue for currently released versions of the SDK since this change is going into next major version?

@jaxdesmarais After thinking about it, this PR could go into `main` since the change isn't breaking and fixes an existing issue. I'll point the branch to `main`

stechiu:
> > Just a small comment about updating our docstrings in all 3 inits to note the default in the inits. Also, do we need to make any changes on main to fix the sizing issue for currently released versions of the SDK since this change is going into next major version?
> 
> @jaxdesmarais After thinking about it, this PR could go into `main` since the change isn't breaking and fixes an existing issue. I'll point the branch to `main`

Well it looks like pointing to `main` means introducing all the changes made from `next-major-version` into `main`... If we prefer this to go into `main`, it's probably best to close this PR and open a new one that is branched from `main`. What do you think?

jaxdesmarais:
> After thinking about it, this PR could go into main since the change isn't breaking and fixes an existing issue. I'll point the branch to main

Yep, now that we separated out the representable changes into a follow up PR we should be able to make the changes in this PR in `main`. Agree it's likely easier to close this PR and open a new PR off of main so you don't end up with all the changes from next major version.

stechiu:
> > After thinking about it, this PR could go into main since the change isn't breaking and fixes an existing issue. I'll point the branch to main
> 
> Yep, now that we separated out the representable changes into a follow up PR we should be able to make the changes in this PR in `main`. Agree it's likely easier to close this PR and open a new PR off of main so you don't end up with all the changes from next major version.

Ok sounds good! I'll make a new PR where I branch from `main`. Thanks for your help

Sources/PaymentButtons/PayPalButton.swift
stechiu: Oops I just saw this. I'll make the change!

PR #253, opened 2024-02-15
Revert `PaymentButton` colors and updated to `standard` size on `next-major-version` branch

stechiu:
### Reason for changes

We want to revert the color changes on the `next-major-version` branch until we get final designs from Design team.

Relevant PR where colors were removed from `next-major-version` https://github.com/paypal/paypal-ios/pull/243

### Summary of changes

- Reverted to original colors
- Updated sizes to `standard` so there are now only 2 sizes for merchants to choose from

### Checklist

- [x] Added a changelog entry

- @stechiu


Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -8,6 +8,9 @@ public final class PayPalButton: PaymentButton {
     public enum Color: String {
         case gold
         case white
+        case black
+        case silver
+        case blue
```
richherrera, line 13: Could you attach the PR that has the changes we are reverting? It would help a lot for the review, I think here we need to revert the changes to the default color parameter on _init's._

stechiu: Here's the relevant PR where colors were removed from `next-major-version` https://github.com/paypal/paypal-ios/pull/243

stechiu: The inits should be updated with the original colors, but just check against the original to make sure

Sources/PaymentButtons/PayPalCreditButton.swift
```diff
@@ -25,7 +25,7 @@ public final class PayPalCreditButton: PaymentButton {
     ///   - size: Size of the button. Default to standard if not provided.
     public convenience init(
         insets: NSDirectionalEdgeInsets? = nil,
-        color: Color = .white,
+        color: Color = .darkBlue,
```
jaxdesmarais, line 28: We should update the docstring above this init as well

stechiu: Missed that, thanks! I've updated the docstrings in `PayPalButton` and `PayPalPayLaterButton` as well

jaxdesmarais (approved):

richherrera (approved):

PR #254, opened 2024-02-15
[iOS] - Revert URL Parsing, CardResult change properties

KunJeongPark:
### Summary of changes

- CardResult to return order status and boolean threeDSecureAttempted
- Revert parsing of return url

No 3DS Contingency:
https://github.com/paypal/paypal-ios/assets/9273272/6988c66b-9071-40e7-a3a8-9379382179f8

3DS Contingency:

https://github.com/paypal/paypal-ios/assets/9273272/548e7e9f-a647-4548-a519-9eca97dadc3d

### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/CardPayments/CardClientError.swift
+++ b/Sources/CardPayments/CardClientError.swift
@@ -82,22 +82,4 @@ enum CardClientError {
         domain: domain,
         errorDescription: "GraphQLClient is unexpectedly nil."
     )
-    
-    static let threeDSVerificationError = CoreSDKError(
-        code: Code.threeDSVerificationError.rawValue,
-        domain: domain,
-        errorDescription: "3DS Verification is returning an error."
-    )
-    
-    static let missingDeeplinkURLError = CoreSDKError(
-        code: Code.missingDeeplinkURLError.rawValue,
-        domain: domain,
-        errorDescription: "Missing deeplink URL from 3DS."
-    )
-    
-    static let malformedDeeplinkURLError = CoreSDKError(
-        code: Code.malformedDeeplinkURLError.rawValue,
-        domain: domain,
-        errorDescription: "Malformed deeplink URL."
-    )
 }
--- a/Sources/CardPayments/Models/CardResult.swift
+++ b/Sources/CardPayments/Models/CardResult.swift
@@ -9,10 +9,9 @@ public struct CardResult {
     /// The order ID associated with the transaction
     public let orderID: String
 
-    /// This is the deep link url returned from 3DS authentication
-    @_documentation(visibility: private)
-    public let deepLinkURL: URL?
-    
+    /// status of the order
+    public let status: String?
+
     /// Liability shift value returned from 3DS verification
-    public let liabilityShift: String?
+    public let threeDSecureAttempted: Bool
 }
--- a/Sources/CardPayments/Models/CardResult.swift
+++ b/Sources/CardPayments/Models/CardResult.swift
@@ -12,6 +12,6 @@ public struct CardResult {
     /// status of the order
     public let status: String?
 
-    /// 3DS verification was attempted. Use v3/setup-tokens/{id} in your server to get verification results.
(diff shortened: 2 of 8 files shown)


Sources/CardPayments/Models/CardResult.swift
```diff
@@ -9,10 +9,9 @@ public struct CardResult {
...
-    /// This is the deep link url returned from 3DS authentication
-    @_documentation(visibility: private)
-    public let deepLinkURL: URL?
-    
+    /// status of the order
+    public let status: String?
+
     /// Liability shift value returned from 3DS verification
-    public let liabilityShift: String?
+    public let threeDSecureAttempted: Bool
```
braintreeps:
```suggestion
    public let didAttemptThreeDSecureAuthentication: Bool
```
super nit: An `is` prefix may be more strictly in line with naming conventions for `Bool` properties, but "did" here reads well IMO, even though that naming convention is more common with delegate methods.

braintreeps (approved):

sshropshire:
🤦‍♂️ Braintreeps was me. I was trying to help with some permissions stuff but forgot to log out.

Sources/CardPayments/Models/CardResult.swift
KunJeongPark: I like it! I'd like to hear what others say about it. We also have the same property for CardVaultResult in vault without purchase PR: https://github.com/paypal/paypal-ios/blob/3e0d2636c43b4b7f8d7d55421d59bf4c550f9d51/Sources/CardPayments/Models/CardVaultResult.swift#L16

Sources/CardPayments/CardClientError.swift
```diff
@@ -82,22 +82,4 @@ enum CardClientError {
         domain: domain,
         errorDescription: "GraphQLClient is unexpectedly nil."
     )
-    
-    static let threeDSVerificationError = CoreSDKError(
```
jaxdesmarais, line 86: I think we will want to remove these from the CHANGELOG

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -9,10 +9,9 @@ public struct CardResult {
...
 
-    /// This is the deep link url returned from 3DS authentication
-    @_documentation(visibility: private)
-    public let deepLinkURL: URL?
-    
+    /// status of the order
+    public let status: String?
+
     /// Liability shift value returned from 3DS verification
-    public let liabilityShift: String?
```
jaxdesmarais, line 17: Just adding a comment so we don't forget to remove this from the CHANGELOG and add the new property. We should also update this docstring.

Sources/CardPayments/Models/CardResult.swift
richherrera: I like it, agree with `didAttemptThreeDSecureAuthentication`

Sources/CardPayments/CardClientError.swift
KunJeongPark: yes, I haven't done CHAGNELOG yet since I thought we'd probably need to settle on property names first.

Sources/CardPayments/Models/CardResult.swift
KunJeongPark: oh the docString also

richherrera (approved):
LGTM!! Thanks

Sources/CardPayments/Models/CardResult.swift
```diff
@@ -9,10 +9,9 @@ public struct CardResult {
...
-    /// This is the deep link url returned from 3DS authentication
-    @_documentation(visibility: private)
-    public let deepLinkURL: URL?
-    
-    /// Liability shift value returned from 3DS verification
-    public let liabilityShift: String?
+    /// status of the order
+    public let status: String?
+
+    /// 3DS verification was attempted. Use v3/setup-tokens/{id} in your server to get verification results.
```
KunJeongPark, line 15: oops wrong endpoint in docstrings

PR #256, opened 2024-02-20
Update Environment URLs

sshropshire:
### Summary of changes

 - Use `api-m` style URLs for communicating with PayPal APIs
 - Equivalent Android PR: https://github.com/paypal/paypal-android/pull/241

 ### Checklist

- @sshropshire

### DIFF
--- a/Sources/CorePayments/Networking/Enums/Environment.swift
+++ b/Sources/CorePayments/Networking/Enums/Environment.swift
@@ -8,9 +8,9 @@ public enum Environment {
     var baseURL: URL {
         switch self {
         case .sandbox:
-            return URL(string: "https://api.sandbox.paypal.com")!
+            return URL(string: "https://api-m.sandbox.paypal.com")!
         case .live:
-            return URL(string: "https://api.paypal.com")!
+            return URL(string: "https://api-m.paypal.com")!
         }
     }
 
--- a/Sources/CorePayments/Networking/TrackingEventsAPI.swift
+++ b/Sources/CorePayments/Networking/TrackingEventsAPI.swift
@@ -13,7 +13,7 @@ class TrackingEventsAPI {
     // MARK: - Initializer
     
     init(coreConfig merchantConfig: CoreConfig) {
-        // api.sandbox.paypal.com does not currently send FPTI events to BigQuery/Looker
+        // api-m.sandbox.paypal.com does not currently send FPTI events to BigQuery/Looker
         self.coreConfig = CoreConfig(clientID: merchantConfig.clientID, environment: .live)
         self.networkingClient = NetworkingClient(coreConfig: coreConfig)
     }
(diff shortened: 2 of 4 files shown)


KunJeongPark:
Steven, what do you think about the fact that `confirm-payment-source` and `updateSetupToken` return url's without `-m`

sshropshire:
> Steven, what do you think about the fact that `confirm-payment-source` and `updateSetupToken` return url's without `-m`

If the API response contains links with `api` instead of `api-m` it's probably fine. I'm going off that document that Tim shared that says to prefer the `-m` URLs for communicating with PayPal APIs.

Sources/CorePayments/Networking/Enums/Environment.swift
```diff
@@ -8,9 +8,9 @@ public enum Environment {
     var baseURL: URL {
         switch self {
         case .sandbox:
-            return URL(string: "https://api.sandbox.paypal.com")!
+            return URL(string: "https://api-m.sandbox.paypal.com")!
         case .live:
-            return URL(string: "https://api.paypal.com")!
+            return URL(string: "https://api-m.paypal.com")!
         }
```
KunJeongPark, line 14: I c, so it's more efficient to use this in our network calls overall?

KunJeongPark: Do we need to check all features with new url? I can just run through demo for all the features.
We have no way of testing live environment.

KunJeongPark: Seems to work fine on all demo features. Thanks Steven!

KunJeongPark (approved):

Sources/CorePayments/Networking/Enums/Environment.swift
sshropshire: > I c, so it's more efficient to use this in our network calls overall?

Yeah it seems like it will be because of some CDN network edge optimization.

richherrera (approved):

jaxdesmarais:
Just worth calling out that we recently removed the `-m` on [BT iOS](https://github.com/braintree/braintree_ios/pull/1175) and [BT Android](https://github.com/braintree/braintree_android/pull/896). Not sure if it's worth digging more into the why. It seems like Sammy discovered it wasn't needed, but maybe that mainly applies to FPTI?

scannillo:
> Just worth calling out that we recently removed the -m on https://github.com/braintree/braintree_ios/pull/1175 and https://github.com/braintree/braintree_android/pull/896. Not sure if it's worth digging more into the why. It seems like Sammy discovered it wasn't needed, but maybe that mainly applies to FPTI?

Yeh, I wasn't able to track down someone on the API side who could speak to what it did differently than without the `-m`. I saw a Confluence doc, but it was a few years old without much detail. Would love to see what doc Tim was referencing 👀 

The main reason though was that PaymentReady doesn't work with `-m`. Harish, who owns the PaymentReady API, hasn't heard of `-m`.

warmkesselj (approved):

PR #257, opened 2024-02-26
Demo app - Make Buliding urlRequests Consistent

KunJeongPark:
### Summary of changes

-  Cleaned up creating setupToken and paymentToken in demo app
    using the same buildBaseURL and buildURLRequest.
-  Included option to exclude httpBody for get calls in buildURLRequest function in preparation for get setup token 
    endpoint call for vault without purchase 3Ds, get order call for vault with purchase 3DS. 

### Checklist

- @KunJeongPark


warmkesselj (approved):

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -16,41 +14,39 @@ final class DemoMerchantAPI {
...
-                environment: DemoSettings.environment,
-                selectedMerchantIntegration: selectedMerchantIntegration
-            )
-            
-            let data = try await data(for: urlRequest)
+            let requestBody = CreateSetupTokenParam(customer: VaultCustomer(id: customerID), paymentSource: paymentSourceType)
+
+            guard let url = buildBaseURL(with: "/setup_tokens", selectedMerchantIntegration: selectedMerchantIntegration)
+
+            else {
```
```suggestion
            guard let url = buildBaseURL(with: "/setup_tokens", selectedMerchantIntegration: selectedMerchantIntegration) else {
```

jaxdesmarais: Also do we need to pass `selectedMerchantIntegration` into this method at all? Can't we just use `DemoSettings.merchantIntegration` in the `buildBaseURL` method directly?

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -16,41 +14,39 @@ final class DemoMerchantAPI {
...
-            let request = PaymentTokenRequest(setupToken: setupToken)
-            let urlRequest = try createPaymentTokenUrlRequest(
-                paymentTokenRequest: request,
-                environment: DemoSettings.environment,
-                selectedMerchantIntegration: selectedMerchantIntegration
-            )
-            let data = try await data(for: urlRequest)
+            let requestBody = PaymentTokenParam(paymentSource: PaymentTokenParam.PaymentSource(setupTokenID: setupToken))
+            guard let url = buildBaseURL(with: "/payment_tokens", selectedMerchantIntegration: selectedMerchantIntegration)
+            else {
```
```suggestion
            guard let url = buildBaseURL(with: "/payment_tokens", selectedMerchantIntegration: selectedMerchantIntegration) else {
```

jaxdesmarais: Same question here around not passing around `selectedMerchantIntegration` in this or the top level method and instead accessing the `DemoSettings` directly.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -147,9 +143,11 @@ final class DemoMerchantAPI {
         urlRequest.httpMethod = method
         urlRequest.addValue("application/json", forHTTPHeaderField: "Content-Type")
 
-        if let json = try? encoder.encode(body) {
-            print(String(data: json, encoding: .utf8) ?? "")
-            urlRequest.httpBody = json
+        if method != "GET" {
```
jaxdesmarais: This can be a punt for this PR but we should consider extracting our methods out into an enum vs checking against raw strings

Demo/Demo/Models/PaymentTokenParam.swift
```diff
@@ -0,0 +1,21 @@
...
+
+struct PaymentTokenParam: Encodable {
+
+    let paymentSource: PaymentSource
+
+    struct PaymentSource: Encodable {
+
+        let token: Token
+
+        public init(setupTokenID: String) {
```
richherrera:
```suggestion
       init(setupTokenID: String) {
```

Demo/Demo/Models/CreateSetupTokenParam.swift
```diff
@@ -1,5 +1,16 @@
 import Foundation
 
+struct CreateSetupTokenParam: Encodable {
+
+    var customer: VaultCustomer?
```
richherrera:
```suggestion
    let customer: VaultCustomer?
```

Demo/Demo/Networking/DemoMerchantAPI.swift
KunJeongPark: We can, but I thought of buildBaseURL just taking in values and building them.  maybe we might move it out of UserDefaults at some point. This value is passed in from main Feature controller view through DemoSettings, and I've left the functions just get the values from params so if we change how we store the values, we can just do it in one spot.

Demo/Demo/Networking/DemoMerchantAPI.swift
KunJeongPark: We can, but I thought of buildBaseURL just taking in values and building them. maybe we might move it out of UserDefaults at some point. This value is passed in from main Feature controller view through DemoSettings, and I've left the functions just get the values from params so if we change how we store the values, we can just do it in one spot.

Demo/Demo/Networking/DemoMerchantAPI.swift
```diff
@@ -147,9 +143,11 @@ final class DemoMerchantAPI {
...
+            if let json = try? encoder.encode(body) {
+                print(String(data: json, encoding: .utf8) ?? "")
+                urlRequest.httpBody = json
+            }
         }
```
richherrera:
```suggestion
        if method != "GET", let json = try? encoder.encode(body) {
                print(String(data: json, encoding: .utf8) ?? "")
                urlRequest.httpBody = json
        }
```

Demo/Demo/Networking/DemoMerchantAPI.swift
jaxdesmarais: I am not sure I follow. My suggestion is to update `buildBaseURL` from:
```swift
// current
private func buildBaseURL(with endpoint: String, selectedMerchantIntegration: MerchantIntegration = .direct) -> URL? {
    return URL(string: DemoSettings.environment.baseURL + selectedMerchantIntegration.path + endpoint)
}

// updated
private func buildBaseURL(with endpoint: String) -> URL? {
    URL(string: DemoSettings.environment.baseURL + DemoSettings.merchantIntegration.path + endpoint)
}
```
It seems like we pass `DemoSettings.merchantIntegration` all the way down from our views when we shouldn't need to. This PR seems to be a good opportunity to address that in or at least part of it.

richherrera (approved):
LGTM!! Just small suggestions :)

Demo/Demo/Networking/DemoMerchantAPI.swift
KunJeongPark: Yes, that was intentional in case we move it out of user defaults, so we can change it to one spot.
In demo app, there is selection for merchant integration from the main feature view. So if we don't use UserDefaults, it would have to be passed down through params.

jaxdesmarais: Is there a plan to move that out of `UserDefaults`? I've not seen any conversations around that so not sure if you can point me in the direction of that conversation I can get caught up!

KunJeongPark: Oh no,  I meant just the value for selectedMerchantIntegration. No, there isn't. 
I guess it defeats the purpose of having it in UserDefaults if it is passed down the view hierarchy.

jaxdesmarais: 🙈 yeah that's what I am thinking as well. We can always move back to passing it through at a later state if we decide to move it out of `UserDefaults` but agree it defeats the purpose quite a bit currently. We don't have to tackle it _everywhere_ in this PR but I think addressing the current surface area of this PR make sense. What do you think?

sshropshire: This pattern comes from the Android Demo app where the UX is a bit different. Grabbing this value directly from `UserDefaults` would make sense on iOS.

We should have flexibility to change iOS and Android Demo apps independently. The UI requirements are fuzzy around testing different merchant integrations.

PR #258, opened 2024-02-28
Vault Without Purchase Analytics

vradopp:
### Reason for changes

We want to add analytics to the flow to track merchant’s usage of vault without purchase.

### Summary of changes
Added analytics events for vault without purchase flows

**PayPal Wallet Vault Events**
`paypal-web-payments:vault-wo-purchase:started`
`paypal-web-payments:vault-wo-purchase:canceled`
`paypal-web-payments:vault-wo-purchase:succeeded`
`paypal-web-payments:vault-wo-purchase:failed`

**Card Vault Events**
`card-payments:vault-wo-purchase:started`
`card-payments:vault-wo-purchase:challenge-required`
`card-payments:vault-wo-purchase:challenge:canceled`
`card-payments:vault-wo-purchase:succeeded`
`card-payments:vault-wo-purchase:failed`

### Checklist

- [ ] Added a changelog entry

- vradopp

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -44,6 +44,8 @@ public class CardClient: NSObject {
     /// - Parameters:
     ///   - vaultRequest: The request containing setupTokenID and card
     public func vault(_ vaultRequest: CardVaultRequest) {
+        analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
+        analyticsService?.sendEvent("card-payments:vault-wo-purchase:started")
         Task {
             do {
                 let result = try await vaultAPI.updateSetupToken(cardVaultRequest: vaultRequest).updateVaultSetupToken
@@ -54,6 +56,7 @@ public class CardClient: NSObject {
                         self.notifyVaultFailure(with: CardClientError.threeDSecureURLError)
                         return
                     }
+                    analyticsService?.sendEvent("card-payments:vault-wo-purchase:challenge-required")
                     startVaultThreeDSecureChallenge(url: url, setupTokenID: vaultRequest.setupTokenID)
                 } else {
                     let vaultResult = CardVaultResult(setupTokenID: result.id, status: result.status, didAttemptThreeDSecureAuthentication: false)
@@ -190,10 +193,12 @@ public class CardClient: NSObject {
     }
     
     private func notifyVaultSuccess(for vaultResult: CardVaultResult) {
+        analyticsService?.sendEvent("card-payments:vault-wo-purchase:succeeded")
         vaultDelegate?.card(self, didFinishWithVaultResult: vaultResult)
     }
 
     private func notifyVaultFailure(with vaultError: CoreSDKError) {
+        analyticsService?.sendEvent("card-payments:vault-wo-purchase:failed")
         vaultDelegate?.card(self, didFinishWithVaultError: vaultError)
     }
 
@@ -203,6 +208,7 @@ public class CardClient: NSObject {
     }
 
     private func notifyVaultCancellation() {
+        analyticsService?.sendEvent("card-payments:vault-wo-purchase:challenge:canceled")
         vaultDelegate?.cardThreeDSecureDidCancel(self)
     }
 }
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -104,6 +104,10 @@ public class PayPalWebCheckoutClient: NSObject {
     /// - Returns: PayPalVaultResult
     /// - Throws: PayPalSDK error if vaulting could not complete successfully
     public func vault(url: URL) {
+        if let setupToken = getQueryStringParameter(url: url.absoluteString, param: "approval_session_id") {
+            analyticsService = AnalyticsService(coreConfig: config, setupToken: setupToken)
+        }
+        analyticsService?.sendEvent("paypal-web-payments:vault-wo-purchase:started")
         webAuthenticationSession.start(
             url: url,
             context: self,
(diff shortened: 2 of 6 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -104,6 +104,10 @@ public class PayPalWebCheckoutClient: NSObject {
     /// - Returns: PayPalVaultResult
     /// - Throws: PayPalSDK error if vaulting could not complete successfully
     public func vault(url: URL) {
```
scannillo, line 106: Outside the scope of this PR - but why doesn't `PayPalWebCheckoutClient.vault()` accept a request object? The entry point to all of our other SDK features use the request object pattern (see [CardClient.vault()](https://github.com/paypal/paypal-ios/blob/15fa228521118f64ba9685f496d5c44c51631fb6/Sources/CardPayments/CardClient.swift#L46), [PayPalWebClient.start()](https://github.com/paypal/paypal-ios/blob/15fa228521118f64ba9685f496d5c44c51631fb6/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift#L35), [PayPalNativeCheckoutClient.start()](https://github.com/paypal/paypal-ios/blob/15fa228521118f64ba9685f496d5c44c51631fb6/Sources/PayPalNativePayments/PayPalNativeCheckoutClient.swift#L43)).

The request object pattern helps us easily add request params as needed as features scale versus needing to add new overload methods. 

cc: @KunJeongPark

scannillo: IMO a `PayPalVaultRequest` with a required `setupTokenID` & `url` would help future proof our analytics (in case the URL changes the `approval_session_id` param name), as well as align us better to the Card vaulting feature's public API.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -104,6 +104,10 @@ public class PayPalWebCheckoutClient: NSObject {
...
+        if let setupToken = getQueryStringParameter(url: url.absoluteString, param: "approval_session_id") {
+            analyticsService = AnalyticsService(coreConfig: config, setupToken: setupToken)
+        }
```
scannillo, line 109: Also slightly unrelated to analytics (but revealed by analytics 😆) - should we boot a merchant from the flow if their URL doesn't contain this value? Currently merchants can pass us any URL and we'll open it in an ASWebSession.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire: I agree we've never discussed this as a group. A `PayPalVaultRequest` object with a `url` property would be consistent with other methods. I'm personally 50 / 50 on adding a `setupTokenID` since it's an extra property, but if it's critical to analytics we should add it in.

Sources/CardPayments/CardClient.swift
```diff
@@ -44,6 +44,8 @@ public class CardClient: NSObject {
     /// - Parameters:
     ///   - vaultRequest: The request containing setupTokenID and card
     public func vault(_ vaultRequest: CardVaultRequest) {
+        analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
+        analyticsService?.sendEvent("card-payments:vault-wo-purchase:started")
```
jaxdesmarais, line 48: Should we extract analytics out into a constants file on PPCP: [example from the BT SDK](https://github.com/braintree/braintree_ios/blob/main/Sources/BraintreeCard/BTCardAnalytics.swift)?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire: +1 for being precautious. The only risk here would be if the format of the URL we get from PPaaS changed one day to something we don't expect, it could break existing integrations.

jaxdesmarais: Agree, seems like a pretty big risk vector. You could pass in `stealyourcreditcard.com` and we'd be like "sounds good!" and open it right up.

sshropshire: Y'all know what we would need to allowlist so we can patch this?

jaxdesmarais: Luckily this hasn't been released yet on iOS, so we should be able to resolve this. We also should figure out when we want to release since our unreleased section is real long and we probably should be doing smaller releases along the way.

sshropshire: Yeah on second thought maybe restricting the TLD to `paypal.com` at least? That way we know for certain it's going to us over TLS.

scannillo (approved):
Actual changes look good - comments left are outside the scope of analytics 👍

Sources/CardPayments/CardClient.swift
sshropshire: This is a good call. Making DTPPCPSDK-1984 to address this project wide.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire: Captured in DTPPCPSDK-1985.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire: Captured in DTPPCPSDK-1986.

sshropshire (approved):

PR #260, opened 2024-03-01
Create Request Object for PayPalWebCheckoutClient.vault()

KunJeongPark:
### Summary of changes

- Create PayPalVaultRequest
- Refactor PayPalWebCheckout.vault(url: String) to PayPalWebCheckout.vault(_ vaultRequest: PayPaylVaultRequest)
- Adjust PayPalClient_Tests with modified vault function argument
- Change demo app to pass in setupTokenID and url in PayPalVaultRequest into vault function

### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,9 @@
   * Add `cardThreeDSecureDidCancel()` to `CardVaultDelegate`
   * Add `cardThreeDSecureWillLaunch()` to `CardVaultDelegate`
   * Add `cardThreeDSecureDidFinish()` to `CardVaultDelegate`
-* PayPalWebPayments
-  * Add `vault(url:)` method to `PayPalWebCheckoutClient`
+* PayPalWebPayments  
+  * Add `PayPalVaultRequest` type for interacting with `vault` method
+  * Add `vault` method to `PayPalWebCheckoutClient`
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
   * Add `cardThreeDSecureDidFinish()` to `CardVaultDelegate`
 * PayPalWebPayments  
   * Add `PayPalVaultRequest` type for interacting with `vault` method
-  * Add `vault` method to `PayPalWebCheckoutClient`
+  * Add `vault(_:)` method to `PayPalWebCheckoutClient`
   * Add `PayPalVaultResult` type to return vault result
   * Add `PayPalVaultDelegate` to handle results from vault flow
   * Add `PayPalWebCheckoutClientError.paypalVaultResponseError` for missing or invalid response from vaulting
--- a/Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
+++ b/Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
@@ -14,10 +14,11 @@ struct PayPalVaultView: View {
                         paymentSourceType: PaymentSourceType.paypal(usageType: "MERCHANT")
                     )
                     SetupTokenResultView(vaultViewModel: paypalVaultViewModel)
-                    if let url = paypalVaultViewModel.state.setupToken?.paypalURL {
+                    if let urlString = paypalVaultViewModel.state.setupToken?.paypalURL,
+                        let setupTokenID = paypalVaultViewModel.state.setupToken?.id {
                         Button("Vault PayPal") {
                             Task {
-                                await paypalVaultViewModel.vault(url: url)
+                                await paypalVaultViewModel.vault(url: urlString, setupTokenID: setupTokenID)
                             }
                         }
                         .buttonStyle(RoundedBlueButtonStyle())
--- a/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
@@ -6,7 +6,7 @@ class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
 
     let configManager = CoreConfigManager(domain: "PayPal Vault")
 
-    func vault(url: String) async {
+    func vault(url: String, setupTokenID: String) async {
         guard let paypalUrl = URL(string: url) else {
             return
         }
@@ -17,7 +17,8 @@ class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
             let config = try await configManager.getCoreConfig()
             let paypalClient = PayPalWebCheckoutClient(config: config)
             paypalClient.vaultDelegate = self
-            paypalClient.vault(url: paypalUrl)
+            let vaultRequest = PayPalVaultRequest(url: paypalUrl, setupTokenID: setupTokenID)
+            paypalClient.vault(vaultRequest)
         } catch {
             print("Error in vaulting PayPal Payment")
         }
--- a/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
@@ -7,7 +7,7 @@ class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
     let configManager = CoreConfigManager(domain: "PayPal Vault")
 
     func vault(url: String, setupTokenID: String) async {
-        guard let paypalUrl = URL(string: url) else {
+        guard let payPalURL = URL(string: url) else {
             return
         }
         DispatchQueue.main.async {
@@ -17,7 +17,7 @@ class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
             let config = try await configManager.getCoreConfig()
             let paypalClient = PayPalWebCheckoutClient(config: config)
             paypalClient.vaultDelegate = self
-            let vaultRequest = PayPalVaultRequest(url: paypalUrl, setupTokenID: setupTokenID)
+            let vaultRequest = PayPalVaultRequest(url: payPalURL, setupTokenID: setupTokenID)
             paypalClient.vault(vaultRequest)
         } catch {
             print("Error in vaulting PayPal Payment")
--- a/Sources/PayPalWebPayments/PayPalVaultRequest.swift
+++ b/Sources/PayPalWebPayments/PayPalVaultRequest.swift
@@ -12,7 +12,7 @@ public struct PayPalVaultRequest {
     /// Creates an instance of a PayPal vault request
     /// - Parameters:
     ///    - url: The PayPal URL for the approval web page
-    ///    - setupTokenID: An ID for the setup token to update
+    ///    - setupTokenID: An ID for the setup token associated with the vault
     public init(url: URL, setupTokenID: String) {
         self.url = url
         self.setupTokenID = setupTokenID
--- a/Sources/PayPalWebPayments/PayPalVaultRequest.swift
+++ b/Sources/PayPalWebPayments/PayPalVaultRequest.swift
@@ -3,10 +3,11 @@ import Foundation
 /// A request to vault a PayPal payment method
 public struct PayPalVaultRequest {
 
-    /// PayPal URL for the approval web page
+    /// PayPal approval URL returned as the `href` from the setup token API call
     public let url: URL
 
     /// ID for the setup token associated with the vault
+    /// Returned as  top level `id` from the setup token API call
     public let setupTokenID: String
 
     /// Creates an instance of a PayPal vault request
--- a/Sources/PayPalWebPayments/PayPalVaultRequest.swift
+++ b/Sources/PayPalWebPayments/PayPalVaultRequest.swift
@@ -1,0 +1,16 @@
+import Foundation
+
+/// A request to vault a PayPal payment method
+public struct PayPalVaultRequest {
+
+    /// PayPal URL for the approval web page
+    public let url: URL
+
+    /// ID for the setup token associated with the vault
+    public let setupTokenID: String
+
+    public init(url: URL, setupTokenID: String) {
+        self.url = url
+        self.setupTokenID = setupTokenID
+    }
+}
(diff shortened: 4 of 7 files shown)


CHANGELOG.md
```diff
@@ -16,8 +16,9 @@
   * Add `cardThreeDSecureDidCancel()` to `CardVaultDelegate`
   * Add `cardThreeDSecureWillLaunch()` to `CardVaultDelegate`
   * Add `cardThreeDSecureDidFinish()` to `CardVaultDelegate`
-* PayPalWebPayments
-  * Add `vault(url:)` method to `PayPalWebCheckoutClient`
+* PayPalWebPayments  
+  * Add `PayPalVaultRequest` type for interacting with `vault` method
+  * Add `vault` method to `PayPalWebCheckoutClient`
```
jaxdesmarais: nit to update this to match the public signature:
```suggestion
  * Add `vault(_:)` method to `PayPalWebCheckoutClient`
```

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
```diff
@@ -14,10 +14,11 @@ struct PayPalVaultView: View {
                         paymentSourceType: PaymentSourceType.paypal(usageType: "MERCHANT")
                     )
                     SetupTokenResultView(vaultViewModel: paypalVaultViewModel)
-                    if let url = paypalVaultViewModel.state.setupToken?.paypalURL {
+                    if let urlString = paypalVaultViewModel.state.setupToken?.paypalURL,
+                        let setupTokenID = paypalVaultViewModel.state.setupToken?.id {
                         Button("Vault PayPal") {
                             Task {
-                                await paypalVaultViewModel.vault(url: url)
+                                await paypalVaultViewModel.vault(url: urlString, setupTokenID: setupTokenID)
```
jaxdesmarais, line 21: This may be a silly question but is there a reason our view is pulling these values from the view model only to send it to the view model? Can we just call `await paypalVaultViewModel.vault()` with nothing then pull the setup token and id from the state within the view model's vault method?

Sources/PayPalWebPayments/PayPalVaultRequest.swift
```diff
@@ -0,0 +1,16 @@
...
+/// A request to vault a PayPal payment method
+public struct PayPalVaultRequest {
+
+    /// PayPal URL for the approval web page
+    public let url: URL
+
+    /// ID for the setup token associated with the vault
+    public let setupTokenID: String
+
+    public init(url: URL, setupTokenID: String) {
```
jaxdesmarais, line 17: Lets add some docstrings to this public init

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
KunJeongPark: It's just the way we have the demo app setup. I get the setup token from create API and update the state in viewModel. 

And then to present button to vault PP, I check for existence of  these values that were just stored in vault state and use the vault function in viewModel to launch the approval url. 

I think it'd be weird for vault function in viewModel to just fetch these values without context.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
KunJeongPark: addressed here: bbc4e576f8908793cfbad11761cf66a28862ca2e

CHANGELOG.md
KunJeongPark: addressed here: 7e4e12b86390de218fd37a3bdad50fda9f6d8cd3

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -100,16 +100,15 @@ public class PayPalWebCheckoutClient: NSObject {
...
     /// After setupToken successfullly attaches a payment method, you will need to create a payment token with the setup token
     /// - Parameters:
-    ///   - url: URL created from string value from setupToken API
-    /// - Returns: PayPalVaultResult
-    /// - Throws: PayPalSDK error if vaulting could not complete successfully
-    public func vault(url: URL) {
-        if let setupToken = getQueryStringParameter(url: url.absoluteString, param: "approval_session_id") {
+    ///   - vaultRequest: Request created with url for vault approval and setupTokenID
+    public func vault(_ vaultRequest: PayPalVaultRequest) {
+        if let setupToken = getQueryStringParameter(url: vaultRequest.url.absoluteString, param: "approval_session_id") {
```
vradopp: Should we use the setupToken from the `vaultRequest` instead?

KunJeongPark: Good point, I just resolved the merge conflict and didn't think about it. That would be much easier,

KunJeongPark: addressed here: 981bfc192e1bdc1fb7afc7b1e51719d94a5b9f00
and b597eb7df7e64e90c46812cabb145a9bf358a42b

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
jaxdesmarais: Yeah I guess to me it just seems like a pretty big anti pattern to send data from ViewModel > View > ViewModel since that data already exists in the view model itself. I'm not sure if using something like a `bool` in the view model would be a better way of indicating that we can vault and to show this view. Just worth thinking about if you have ideas.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
```diff
@@ -0,0 +1,20 @@
...
+    /// PayPal URL for the approval web page
+    public let url: URL
+
+    /// ID for the setup token associated with the vault
+    public let setupTokenID: String
+
+    /// Creates an instance of a PayPal vault request
+    /// - Parameters:
+    ///    - url: The PayPal URL for the approval web page
+    ///    - setupTokenID: An ID for the setup token to update
```
jaxdesmarais: The docstrings here vs above are different. Are we able to align the two?

KunJeongPark: Great catch!

KunJeongPark: Addressed here: 743b6afcafd0506de30da7eb932114a50df07b55

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
KunJeongPark: the property is an optional field and nil until it is populated. 
It's a pattern I've used in views before where I choose to present a UI element
depending on presence of some value in a published property. 

So rather than checking in the view model's vault function for presence of its values, I just send the values from the view where it checks the value anyway. I wouldn't try to vault until setup token is there already.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
```diff
@@ -0,0 +1,20 @@
+import Foundation
+
+/// A request to vault a PayPal payment method
+public struct PayPalVaultRequest {
+
+    /// PayPal URL for the approval web page
+    public let url: URL
```
scannillo, line 7: Slight tangent here - but do you have context on why we have to require the URL? I assume some limitation with the API. Do we have it documented somewhere?

KunJeongPark: PayPal vault works very differently than Card vault in that we just simply launch the url for API to be called outside of our SDK and we get results back. There is documentation for v3/vault/setup-tokens with PAYPAL WALLET in PPaas. Not sure if that is what you are asking.

It's also different from PayPal Checkout flow where we can just construct the url with an orderID.
The url from setup token API looks like this:  "href": "https://api-m.paypal.com/agreements/approve?approval_session_id=xxxx"

Demo/Demo/SwiftUIComponents/PayPalVaultViews/PayPalVaultView.swift
KunJeongPark: That we need to launch this url as opposed to mobile SDK directly calling API was researched by Steven
@sshropshire

sshropshire: This was guidance from the JS PPCP SDK. Apparently the URL given to us by PPaaS is lower latency than any URL we can create on our own using a setup token ID.

jaxdesmarais: Gotcha, my question wasn't around the URL at all. It was around passing view model data from the view model back to the view model, which is a bit of an odd pattern. I know we do it in many places, so should think about if there is a better way vs having the view model passed it's own data. Really just something to think about!

KunJeongPark: Oh I think I answered that part in the wrong section. That portion of my answer was in response to Sammy's question on the bottom.

KunJeongPark: Yeah, on your point Jax, I don't need to pass it, I just check existence of those values in the view, so I just passed them so I don't have to check for nil values in the function in the viewModel. I just thought it would be weird that I'd have `vault` function check for nil when I wouldn't get there if the `setup token` value was nil.

This is all on the demo app side, so merchant can choose to handle it differently.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
jaxdesmarais: Should we include that last piece in our docstring to make it more clear to us/merchants where this URL comes from? Something like:
```
/// PayPal approval URL returned as the `href` from the setup token API call
```

Sources/PayPalWebPayments/PayPalVaultRequest.swift
```diff
@@ -0,0 +1,20 @@
...
+
+    /// ID for the setup token associated with the vault
```
jaxdesmarais, line 9: Maybe similar here, should we specify what this param is returned as in the API call?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -100,16 +100,13 @@ public class PayPalWebCheckoutClient: NSObject {
...
     /// - Parameters:
-    ///   - url: URL created from string value from setupToken API
-    /// - Returns: PayPalVaultResult
-    /// - Throws: PayPalSDK error if vaulting could not complete successfully
-    public func vault(url: URL) {
-        if let setupToken = getQueryStringParameter(url: url.absoluteString, param: "approval_session_id") {
-            analyticsService = AnalyticsService(coreConfig: config, setupToken: setupToken)
-        }
+    ///   - vaultRequest: Request created with url for vault approval and setupTokenID
+    public func vault(_ vaultRequest: PayPalVaultRequest) {
```
jaxdesmarais, line 104: This isn't a breaking change because we haven't released this yet right?

Sources/PayPalWebPayments/PayPalVaultRequest.swift
KunJeongPark: Do we document things that happens on merchant's server side or do we leave that to our developer docs?
I like your suggestion of describing it in our SDK in this case because it's not self-explanatory where it comes from. For example, we don't include anything about the fact that orderID comes from merchant server side call in `CardRequest` at rely on developer docs to explain those portions.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: Yes, it is new with next release.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
KunJeongPark: same comment, do we typically mention in docStrings instructions we have on merchant server side?
or do we rely on developer docs.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
jaxdesmarais: I think it could help a merchant who maybe doesn't know where this URL comes from or which URL to pass in. I could certainly see a merchant passing in the wrong URL here so think giving some guidance would be beneficial.

Sources/PayPalWebPayments/PayPalVaultRequest.swift
jaxdesmarais: Same here, it looks like there are a _ton_ of IDs that come back in this response, I think pointing the merchant in the right direction would be helpful.

KunJeongPark: addressed here: dadbd5de17e6cc2524f7128ecedc7968d3dff3ef

Sources/PayPalWebPayments/PayPalVaultRequest.swift
KunJeongPark: addressed here: dadbd5de17e6cc2524f7128ecedc7968d3dff3ef

Demo/Demo/ViewModels/PayPalVaultViewModel.swift
```diff
@@ -17,7 +17,8 @@ class PayPalVaultViewModel: VaultViewModel, PayPalVaultDelegate {
             let config = try await configManager.getCoreConfig()
             let paypalClient = PayPalWebCheckoutClient(config: config)
             paypalClient.vaultDelegate = self
-            paypalClient.vault(url: paypalUrl)
+            let vaultRequest = PayPalVaultRequest(url: paypalUrl, setupTokenID: setupTokenID)
```
jaxdesmarais:
```suggestion
            let vaultRequest = PayPalVaultRequest(url: payPalURL, setupTokenID: setupTokenID)
```

Sources/PayPalWebPayments/PayPalVaultRequest.swift
```diff
@@ -0,0 +1,21 @@
...
+    /// Returned as  top level `id` from the setup token API call
+    public let setupTokenID: String
+
+    /// Creates an instance of a PayPal vault request
+    /// - Parameters:
+    ///    - url: The PayPal URL for the approval web page
+    ///    - setupTokenID: An ID for the setup token associated with the vault
```
jaxdesmarais: Can we also update these docstrings with the updated strings above?

jaxdesmarais (approved):

richherrera (approved):

PR #261, opened 2024-03-07
Add button analytics

vradopp:
### Reason for changes

Log payment button usage

### Summary of changes

- Added `button_type` which is an existing tag to track the type of button used
- Added events for `payment-button:tapped` and `payment-button:initialized` 

### Checklist

~~- [ ] Added a changelog entry~~

- vradopp

### DIFF
--- a/Sources/PaymentButtons/PaymentButton.swift
+++ b/Sources/PaymentButtons/PaymentButton.swift
@@ -26,8 +26,8 @@ public class PaymentButton: UIButton {
 
     // Use an empty config and default to live environment for button analytics
     private let analyticsService = AnalyticsService(
-        coreConfig: .init(clientID: "", environment: .live),
-        orderID: ""
+        coreConfig: .init(clientID: "N/A", environment: .live),
+        orderID: "N/A"
     )
 
     // MARK: - Init
--- a/Sources/PaymentButtons/PaymentButton.swift
+++ b/Sources/PaymentButtons/PaymentButton.swift
@@ -46,7 +46,7 @@ public class PaymentButton: UIButton {
         self.size = size
         self.insets = insets
         self.label = label
-        self.analyticsService.sendEvent("paypal-button:initialized", buttonType: fundingSource.rawValue)
+        self.analyticsService.sendEvent("payment-button:initialized", buttonType: fundingSource.rawValue)
         super.init(frame: .zero)
         self.addTarget(self, action: #selector(onTap), for: .touchUpInside)
     }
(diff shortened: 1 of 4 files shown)


Sources/PaymentButtons/PaymentButton.swift
```diff
@@ -20,6 +24,12 @@ public class PaymentButton: UIButton {
     static let bundle = Bundle(for: PaymentButton.self)
     #endif
 
+    // Use an empty config and default to live environment for button analytics
+    private let analyticsService = AnalyticsService(
+        coreConfig: .init(clientID: "", environment: .live),
+        orderID: ""
```
scannillo: Nit - is it more clear to pass `"N/A"` instead of `""`?

scannillo:
Changes look good! Another nit - but I think we do want a CHANGELOG entry here, it helps future us track which version we added these analytics in

Sources/PaymentButtons/PaymentButton.swift
```diff
@@ -36,7 +46,9 @@ public class PaymentButton: UIButton {
         self.size = size
         self.insets = insets
         self.label = label
+        self.analyticsService.sendEvent("paypal-button:initialized", buttonType: fundingSource.rawValue)
```
jaxdesmarais: General question: do we want to send a specific event for each `PaymentButton` subclass - so credit, pay later, paypal. Or are we sending the same event name for all buttons?

scannillo (approved):

Sources/PaymentButtons/PaymentButton.swift
vradopp: Its the same event name for all buttons but we also pass the `buttonType` which would be one of `PayPal` `Pay Later` or `Credit`

jaxdesmarais: 🤦 it certainly helps if I could read, sorry just totally missed the second half of this event. Take it or leave it, but is it worth considering `payment-button` as the naming here? Wondering incase we add something like a Venmo button in the future would we reused this class.

scannillo: I like that idea!

jaxdesmarais: @vradopp what do you think about updating the events in this PR to "payment-button" vs "paypal-button"?

vradopp: Sounds good to me! Just updated

stechiu (approved):
LGTM!

jaxdesmarais (approved):
🚀 thanks for making that update!

saperi22:
@vradopp 
> Added button_type which is an existing tag to track the type of button used

do you mind pointing me where to look for existing tags? Not sure how to verify if `button_type` is an existing tag.

jaxdesmarais:
> do you mind pointing me where to look for existing tags? Not sure how to verify if `button_type` is an existing tag.

I can help answer this! If you go to `go/fptiexplorer` and click on `tags` on the lefthand side you can search for approved FPTI tags. Generally we want to reuse already approved tags vs creating new tags.

PR #263, opened 2024-03-18
Bump Magnes for Privacy Manifest, Changelog

KunJeongPark:
### Summary of changes

- Updated Changelog to indicate privacy manifest files and Magnes 5.5.0 update
- Updated PPRiskMagnes.xcframework with latest version from Magnes team with updated version
Screenshot 2024-03-19 at 2 50 23 PM

Generated Privacy Report

### Checklist

- [x] Added a changelog entry

- @KunJeongPark @agedd


Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/Headers/PPRiskMagnes-Swift.h
```diff
@@ -1,4 +1,6 @@
-// Generated by Apple Swift version 5.4.2 (swiftlang-1205.0.28.2 clang-1205.0.19.57)
+#if 0
+#elif defined(__arm64__) && __arm64__
+// Generated by Apple Swift version 5.9.2 (swiftlang-5.9.2.2.56 clang-1500.1.0.2.5)
```
scannillo: I'm surprised CI isn't failing. I thought Swift 5.9.2 was introduced in Xcode 15.1, and our CI is building against Xcode 14.3?

KunJeongPark: How did this happen? The README in new Magnes copy said it was generated with Xcode 12. README must not have been updated.

jaxdesmarais: Yeah that's pretty interesting to me as well, are we actually using Magnes anywhere? I wonder if that's why it's not failing but agree quite odd. As far as their README goes they probably are never opening/updating that. There is a 0% chance afaik they could build with Xcode 12 and Swift 5.9.2, so we can ignore their README.

Just to confirm my understanding since I think you already looking into this but wanted to ask:
- Have we confirmed we can actually archive the SDK with this magnes version - wondering if that's when we will get an error? Certainly quite odd CI doesn't return an error
- The current version of Magnes in the SDK is dynamic as well as this one? Or are they both static? If it's different than iOS v6 (static) we should add to our v2 plan for PPCP to switch to static so we can avoid different requirements per SDK.

KunJeongPark: Yes current version in main is dynamic.
I was able to archive the demo app with Magnes and generate Magnes Privacy items. I will include it on top.
My plan, as I discussed with you earlier, @jaxdesmarais is to use data collector in demo app

KunJeongPark: We can send an email out to Magnes team asking for another dynamic framework version built with Xcode 14.3 or we can switch to static library. I saw email that was forwarded to me that there was ask for 14.3 built static version.

I don't know what was involved in original discussion in discussion of using dynamic vs static in PPCP.

jaxdesmarais: > we can switch to static library

This is a breaking change so should be done in a major version - if we have a doc going for PPCP v2 we should ensure we add this to the list to include

> I don't know what was involved in original discussion in discussion of using dynamic vs static in PPCP.

PPCP started before iOS v6, so there wasn't a discussion at the time, we just used what Magnes provided us. 😄

KunJeongPark: I think the privacy manifest should go out first, so we will ask Magnes for new dynamic version built with Xcode 14.3 as that is our minimum requirement for PPCP. Actually, I see that there was an asking for 14.1 version for BT 5.x, so I think we should ask for 14.1 built dynamic version. What do you think?

KunJeongPark: I just replaced Magnes with dynamic version that they sent built with Xcode 15.0
There is a PR that bumps min version to Xcode 15.0 (https://github.com/paypal/paypal-ios/pull/265)
that should be merged in before this.

jaxdesmarais (approved):
👏 that's for also adding this to the demo app so we are using this module somewhere!

Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/Headers/PPRiskMagnes-Swift.h
KunJeongPark: My only lingering question is that Magnes is built with Swift 5.9 and our Package.swift and Podspec files specify Swift 5.8. There are no problems in our pipeline and but I wanted to be absolutely sure that this is ok.

scannillo: > My only lingering question is that Magnes is built with Swift 5.9 and our Package.swift and Podspec files specify Swift 5.8. 

This is a great question. I think since we bumped to Xcode 15.0+ we can also bump to Swift 5.9+. Looks like that was missed in our README [here](https://github.com/paypal/paypal-ios?tab=readme-ov-file#languages) & our Podspec & Package.swift.

jaxdesmarais: But to clarify we don't _have_ to  merchants will not run into issues if we merge this as is. As long as our version is the same/lower than a merchants there will not be issues. We just cannot specify a higher version than our minimally supported version. The main advantage is for us in that if we bump those to Swift 5.9 we can use Swift 5.9 features in the SDK.

KunJeongPark: I just want to make sure there is no confusion for the merchant specifying Swift 5.8 in Package.swift and Podspec files and minimum Xcode version as 15.0+ (Swift 5.9) specified in our README and CHANGELOG.

jaxdesmarais: Xcode 15.0 supports all the way down to Swift 4, so there should be no confusion for merchants if we specify a lower version - they would only run into issues if we say we support 5.9 but actually specified 5.10.

We can update those in a follow up PR along with the README Sammy mentioned, but it's not _required_ to release this.

richherrera: **Swift's Forward Compatibility:** Swift aims to maintain forward compatibility within the same major version, meaning code compiled with an older version of Swift should run fine with binaries compiled with newer Swift versions within the same major version sequence. Swift 5.x versions strive for source compatibility, which extends to binary frameworks as well, assuming they don't leverage any newer language features that are unavailable in the older version.

**Module Stability:** Introduced in Swift 5.1, module stability allows for binary frameworks to be compiled with one version of the Swift compiler and then imported and used in another module compiled with a different Swift version. This is crucial for binary dependencies in ensuring that they work across Swift versions, as long as those versions support module stability.

KunJeongPark: It's a three liner change, so I can replace the comment in Package.swift and Podspec files and README on this PR or open another ticket, I don't have preference

jaxdesmarais: Lets do a follow up since it's not directly released to magnes, can just be a 1 pt ticket for README, podspec, Package.swift, and CHANGELOG. It'll be a small PR but will keep the context of this PR limited to Magnes.

richherrera (approved):

Frameworks/XCFrameworks/PPRiskMagnes.xcframework/ios-arm64/PPRiskMagnes.framework/Headers/PPRiskMagnes-Swift.h
KunJeongPark: Created new ticket  DTPPCPSDK-2104 to update README, Package.swift and Podspec files to show Swift 5.9

PR #277, opened 2024-06-06
Fix Lint Errors

jaxdesmarais:
### Summary of changes

- Fix lint failures on CI (see https://github.com/paypal/paypal-ios/actions/runs/9307131388/job/25617789168)

### Checklist

### Authors

- @jaxdesmarais

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -13,6 +13,7 @@ disabled_rules:
   - type_name # tests will have have the format <SUT>_Tests
   - xctfail_message
   - blanket_disable_command
+  - non_optional_string_data_conversion
 
 opt_in_rules:
   - array_init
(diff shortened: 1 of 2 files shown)


.swiftlint.yml
```diff
@@ -13,6 +13,7 @@ disabled_rules:
   - type_name # tests will have have the format <SUT>_Tests
   - xctfail_message
   - blanket_disable_command
+  - non_optional_string_data_conversion
```
jaxdesmarais, line 16: The places where this rule was being triggered ([example here](https://github.com/paypal/paypal-ios/blob/b46f71db38acf9847ca718714dbc698294bec892/Sources/FraudProtection/DeviceInspector.swift#L35)) were places where we were using `if let` or `guard let` to safely unwrap these values. There is some debate on reverting and/or changing this rule [here](https://github.com/realm/SwiftLint/issues/5263) so I've opted to disable it for us for now since we are already treating the triggers for this rule safely.

richherrera (approved):

sshropshire (approved):

PR #283, opened 2024-07-11
Eligibility Logic skeleton files

richherrera:
### Reason for changes

DTMOBILES-777

- Create the basic components to add the Eligibility logic

### Summary of changes

- Add `EligibilityAPI`, `EligibilityClient`, `EligibilityRequest`, `EligibilityResult` and `EligibilityVariables`

### Checklist

- [ ] Added a changelog entry

### Authors

- @richherrera


scannillo (approved):
Nice! Love starting w/ the public API. 

We'll need to add docstrings too - can we make a follow up ticket for doing that across iOS & Android (we can re-use the same for both platforms)

richherrera:
> Nice! Love starting w/ the public API.
> 
> We'll need to add docstrings too - can we make a follow up ticket for doing that across iOS & Android (we can re-use the same for both platforms)

I like that plan, thanks!!! created DTMOBILES-853

Sources/CorePayments/Eligibility/EligibilityRequest.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+
+public struct EligibilityRequest {
+
+    public let intent: String
+    public let currency: String
+    public let enableFunding: [String]
```
jaxdesmarais: Do we want to make these internal and just set them in the `init`?

agedd (approved):
lgtm! 🚀

jaxdesmarais:
> We'll need to add docstrings too - can we make a follow up ticket for doing that across iOS & Android (we can re-use the same for both platforms)

This _is_ in the acceptance criteria for the ticket along with some additional changes not reflected in this PR. Since it's also missing constructing the API call/response, adding to the demo app, adding a changelog entry, and adding unit tests from the AC. @richherrera was the plan to link multiple PRs to 1 ticket for this work?

Sources/CorePayments/Eligibility/EligibilityRequest.swift
richherrera: Address: [f63935d](https://github.com/paypal/paypal-ios/pull/283/commits/f63935ddc262bdc9da03fa90758b1a9de9b0d190)

richherrera:
> > We'll need to add docstrings too - can we make a follow up ticket for doing that across iOS & Android (we can re-use the same for both platforms)
> 
> This _is_ in the acceptance criteria for the ticket along with some additional changes not reflected in this PR. Since it's also missing constructing the API call/response, adding to the demo app, adding a changelog entry, and adding unit tests from the AC. @richherrera was the plan to link multiple PRs to 1 ticket for this work?

Exactly, that's the plan. I think reviewing PRs would be faster this way, integrating small pieces of code. Multiple PRs will be created to meet all the acceptance criteria of the ticket.

PR #291, opened 2024-10-10
Remove MXO

KunJeongPark:
### Summary of changes

- This PR removes the `PayPalNativePayments` module
- The PayPal Mobile Checkout library has been sunset and will not receive updates going forward

### Checklist

- [x] Added a changelog entry

- @KunJeongPark


jaxdesmarais (commented):
We should also remove `MOBILE_CHECKOUT_MIGRATION_GUIDE.md` as well as make sure it's fully removed from the demo app settings plist.

KunJeongPark:
> We should also remove `MOBILE_CHECKOUT_MIGRATION_GUIDE.md` as well as make sure it's fully removed from the demo app settings plist.

Good catch with migration guide! I don't think we are using those plist values anymore. I think they are remnants of old userDefaults settings. I will remove that plist file.

sshropshire (approved):

PR #292, opened 2024-10-14
Convert Delegation Pattern to Async-Await

KunJeongPark:
### Summary of changes

- convert from returning results via delegate functions to merchants to async await functions that return results or errors directly. 
- provide completion handler version of public functions

Discussion: 

- Created a [separate PR for completion handler first version ](https://github.com/paypal/paypal-ios/pull/293) that offers async await wrapper functions.
   
  Thinking ahead to including app switch features, the completion handler version has been tested in Braintree   
  iOS SDK. I did put up a [POC branch](https://github.com/braintree/braintree_ios/tree/async-app-switch-poc) 
  with async await pattern and saw it work by resuming continuation.

   **Pros** of using async await pattern first is setting a pattern where we don't have to refactor bottom up from completion  
  handler version in the future. Code is much easier to read and unit tests are much easier to implement.

  **Cons** I've experienced working on this PR:
1.   `ASWebAuthenticationSession` returns results via completion handlers, so we are having to wrap a
    significant portion of the public functions in `withCheckedThrowingContinuation`.
2. Higher risk of continuation resume errors. Continuations require strict guarantees that they will only be resumed once. 
3. Possibly more complex scenarios with states with app switch failure
4. As I stated above,  app switch pattern has been tested with completion handler in Braintree iOS SDK in production, not with async/await's continuation.

I am closing this one in favor of [completion handler version](https://github.com/paypal/paypal-ios/pull/293) 

  For the merchant, the integration experience will be the same with completion handler first or async await first approach.
  Both versions will offer both completion handler and async await versions of the public functions.

  Eliminating old delegation pattern with either approach definitely simplifies integration and future development for us.

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark


sshropshire (approved):
I do really like async/await where possible. It makes the code really clean from a maintenance perspective.

PR #296, opened 2024-11-07
Simplify Cancel Errors

KunJeongPark:
### Summary of changes

-  Simplify cancel errors for merchants

### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@
       * Add `vault(request:completion:(CardVaultResult?, Error?) -> Void)` to `CardClient`
       * Add `approveOrder(request:) async throws -> CardResult`
       * Add `vault(vaultRequest:) async throws -> CardVaultResult`   
-      * Add `.canceled` to `CardClientError`   
+      * Add `.threeDSecureCanceled` to `CardClientError`   
 * PayPalWebPayments
   * Deprecate `PayPalVaultRequest(url:setupTokenID:)`
   * Add `PayPalVaultRequest(setupTokenID:)`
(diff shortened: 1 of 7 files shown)


Sources/CardPayments/CardClientError.swift
```diff
@@ -39,7 +39,7 @@ enum CardClientError {
         case malformedDeeplinkURLError
 
         /// 10. Cancellation from 3DS verification
-        case threeDSCancellation
+        case canceledError
```
sshropshire: nit: For this error `threeDSecureCanceled` makes sense to me, since CardClient won't always perform 3DS verification. It like how it implicitly implies that 3DS was attempted.

sshropshire (approved):
One nit, overall this LGTM

Sources/CardPayments/CardClientError.swift
KunJeongPark: I thought about this and I was opting for brevity instead of the error name being more descriptive.
The only way card processing can be canceled is in 3DS. I wanted to keep it as simple as possible 
since cancellation as an error is new.

But I see your point that it is not clear that this is only for 3DS.

scannillo (approved):

CHANGELOG.md
```diff
@@ -22,7 +23,8 @@
       * Add `approveOrder(request:completion:(CardResult?, Error?) -> Void)` to `CardClient`
       * Add `vault(request:completion:(CardVaultResult?, Error?) -> Void)` to `CardClient`
       * Add `approveOrder(request:) async throws -> CardResult`
-      * Add `vault(vaultRequest:) async throws -> CardVaultResult`      
+      * Add `vault(vaultRequest:) async throws -> CardVaultResult`   
+      * Add `.threeDSecureCanceled` to `CardClientError`
```
scannillo, line 27: Nit, these are a rename vs an add. In case a merchant expects the old ones to still be there

KunJeongPark: I wanted to show that a new async/await function was added. 
I think rename might confuse them because it really is a new function.

PR #297, opened 2024-11-12
Cancel error handling

KunJeongPark:
### Summary of changes

- Provide helper methods to merchants to distinguish cancellation errors from 3DS in cards, PayPal Checkout and vault flows. 
- Our SDK is right now returning CoreSDKError to the users. To discern cancellation errors, users need to check domain and code, which are specified by enum values only at internal level, which is cumbersome.
- For next beta release, we should consider restructuring errors returned to merchants so they can easily check for types of errors.
- Demo app changes and unit tests for helper methods
- Cast errors from urlSessions as NetworkingClientError
- Modify completion handlers to return CoreSDKError instead of general error

Without helper methods, merchant has to look up our internal Client level Codes and domain to get matching number, string
```
if let error {
   if error.code.rawValue == 6 && error.domain == "PayPalClientErrorDomain" {
      // handle cancel error
  } else {
    // handle all other errors
  }
}
```

With helper method:
```
if let error {
   if PayPalError.isCheckoutCanceled(error) {
     // handle cancel error
   } else {
    // handle all other errors
   }
```

### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
@@ -16,7 +16,7 @@ class PayPalVaultViewModel: VaultViewModel {
             let vaultRequest = PayPalVaultRequest(setupTokenID: setupTokenID)
             paypalClient.vault(vaultRequest) { result, error in
                 if let error {
-                    if PayPalWebCheckoutClient.isVaultCanceled(error) {
+                    if PayPalWebCheckoutClientError.isVaultCanceled(error) {
                         DispatchQueue.main.async {
                             print("Canceled")
                             self.state.paypalVaultTokenResponse = .idle
--- a/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
@@ -16,7 +16,7 @@ class PayPalVaultViewModel: VaultViewModel {
             let vaultRequest = PayPalVaultRequest(setupTokenID: setupTokenID)
             paypalClient.vault(vaultRequest) { result, error in
                 if let error {
-                    if PayPalWebCheckoutClientError.isVaultCanceled(error) {
+                    if PayPalError.isVaultCanceled(error) {
                         DispatchQueue.main.async {
                             print("Canceled")
                             self.state.paypalVaultTokenResponse = .idle
--- a/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/PayPalVaultViewModel.swift
@@ -16,8 +16,15 @@ class PayPalVaultViewModel: VaultViewModel {
             let vaultRequest = PayPalVaultRequest(setupTokenID: setupTokenID)
             paypalClient.vault(vaultRequest) { result, error in
                 if let error {
-                    DispatchQueue.main.async {
-                        self.state.paypalVaultTokenResponse = .error(message: error.localizedDescription)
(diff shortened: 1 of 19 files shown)


Demo/Demo/ViewModels/PayPalVaultViewModel.swift
```diff
@@ -16,8 +16,15 @@ class PayPalVaultViewModel: VaultViewModel {
             let vaultRequest = PayPalVaultRequest(setupTokenID: setupTokenID)
             paypalClient.vault(vaultRequest) { result, error in
```
sshropshire, line 17: qq: For the `PayPalClient.vault()` callback can we do `error: CoreSDKError` instead of `error: Error`?

KunJeongPark: I followed the Braintree example where merchant doesn't have to know about error types unless they need to for specific purposes.

KunJeongPark: The reason for opening this PR was while I was working on migration guide, if the merchant wants to treat cancel error separately (I've seen these asks coming in from merchants on Braintree side), it is quite cumbersome the way we have the errors set up in our iOS SDK. They have to cast to CoreSDKError, then look for specific number for code and value for domain. I thought about exposing CardClientError as public so they can check those values CardClientError.domain, CardClientError.Code.rawValue, but that seemed very messy too because CardClientError internally just creates CoreSDKError.

KunJeongPark: I see your point, so that the merchant is aware of type of errors we are throwing. But unless they are wanting to distinguish an error as in cancel case, it is really not very user friendly besides codes they'd have to look up. Localized error description is the same for regular error and CoreSDKError.

sshropshire: Yeah to be honest the type wrangling seems unnecessary since we could just return `CoreSDKError`. It results in less code and feels a lot cleaner. Would it be worth it to consider diverging from the BT way?

KunJeongPark: they'd still have to check for code an domain, I guess I can make CardClientError and PayPalWebCheckoutError public so they can check CardClientError.domain, code.

KunJeongPark: I was thinking also, if we do restructuring of the errors, I think we should eventually,
this helper function abstracts away the implementation so merchants don't have to make changes in their code.

sshropshire: Yeah honestly the domain portion feels cumbersome too. It feels like it's only there because the `Error` type is so generic. I'd say go what feels best but I wouldn't get tied down into how BT does it if there's a better way.

It's hard for me to comment from the Android side. We're using sealed classes so we don't need `domain` and `errorCode`. Kotlin's type system captures all of the error type information. Merchants can use `switch` statements to get more detail on the exact error cause.

KunJeongPark: Yes, I initially just wrote that way on migration guide with switch statement and then I tried it. ❌
Per our discussion, I moved the helper functions to the CardClientError and PayPalClientError
and. kept the general Error return to the merchant's completion handlers.

I will push up changing name of CardClientError enum to -> CardError
and PayPalWebCheckoutError -> PayPalError

sshropshire (approved):

PR #298, opened 2024-11-14
Make PayPalError and CardError Code enum consistent with error naming pattern

KunJeongPark:
### Summary of changes

- Make PayPalError and CardError Code enum consistent with error naming pattern
- Changelog entries

### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/CardPayments/CardError.swift
+++ b/Sources/CardPayments/CardError.swift
@@ -39,7 +39,7 @@ public enum CardError {
         case malformedDeeplinkURLError
 
         /// 10. Cancellation from 3DS verification
-        case threeDSCancellationError
+        case threeDSecureCanceledError
     }
 
     static let unknownError = CoreSDKError(
@@ -68,8 +68,8 @@ public enum CardError {
         errorDescription: "An invalid 3DS URL was returned. Contact developer.paypal.com/support."
     )
 
-    static let threeDSecureCanceled = CoreSDKError(
-        code: Code.threeDSCancellationError.rawValue,
+    static let threeDSecureCanceledError = CoreSDKError(
+        code: Code.threeDSecureCanceledError.rawValue,
         domain: domain,
         errorDescription: "3DS verification has been canceled by the user."
     )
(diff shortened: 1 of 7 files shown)


sshropshire (approved):

Sources/CardPayments/CardError.swift
```diff
@@ -39,7 +39,7 @@ public enum CardError {
         case malformedDeeplinkURLError
 
         /// 10. Cancellation from 3DS verification
-        case threeDSCancellationError
+        case threeDSecureCanceledError
```
sshropshire, line 42: This is a random side nit: I noticed the `Error` suffix may be a bit redundant here since the type is `CardError`. I wonder if it's worth cleaning up in a future release.

KunJeongPark: Yes, I thought about that, and that was my initial reason for changing it in the static function name but noticed that there was enum and static function name match and didn't want to deviate just for these.

What I want for next release is to be able to do `if error == CardError.threeDSecureCanceled` or if `error == PayPalError.checkoutCanceled`.
Or something hierarchical like `error.source == .card(.threeDSecureCanceled)` if we keep the `CoreSDKError` <= this seems more cumbersome than what I have here though

I think it's matter of preference and depends on if it's something merchant would act on.
I can definitely see merchants wanting to separate cancel case as I did in the demo app.

sshropshire: Got it yeah. I think the Swift API Guidelines have a naming convention for factory methods if we want to go with something like `makeThreeDSecureCanceledError()` for the internal method in the future. But it's super nit. I do think brevity in the `case` name (within reason) is a big win for developer experience.

KunJeongPark: I think the first one is more aligned with what I see on Android.

sshropshire: Oh yeah that definitely will change on Android. They shouldn't have been named that way.

PR #301, opened 2024-11-19
Merge 2.0.0-beta1 to main

KunJeongPark:
### Summary of changes

- remove MXO
- from delegation pattern to completion handler / async await
- cancellation error helper functions

### Checklist

- [x] Added a changelog entry

- @KunJeongPark


sshropshire (approved):

Demo/Demo/Info.plist
```diff
@@ -3,7 +3,7 @@
 <plist version="1.0">
 <dict>
 	<key>CFBundleShortVersionString</key>
-	<string>1.5.0</string>
+	<string>2.0.0-beta1</string>
```
KunJeongPark, line 6: I think this is handled in the release script, but fixed it before I saw the yml file. I think it's fine.

Sources/CardPayments/CardClient.swift
```diff
@@ -87,33 +109,52 @@ public class CardClient: NSObject {
...
 
     private func startThreeDSecureChallenge(
         url: URL,
-        orderId: String
+        orderId: String,
+        completion: @escaping (CardResult?, CoreSDKError?) -> Void
     ) {
-        delegate?.cardThreeDSecureWillLaunch(self)
         
         webAuthenticationSession.start(
```
GMALKHA, line 159: if the webAuthenticationSession does not complete due to the network issues the sessionDidComplete might never be called .  maybe we should include some timeout mechanism here ?

GMALKHA (approved):

Sources/CardPayments/CardClient.swift
KunJeongPark: That's a good call out on issues we just discussed about some API responses being very slow or non responsive in Sandbox environment. I am not sure if this would be an issue in prod unless something is very wrong. 

Networking issues on customer's end are valid consideration. Let's create a ticket for investigating time out mechanisms and implementation for next release.

PR #303, opened 2024-11-22
Make CoreSDKError Equatable

KunJeongPark:
### Summary of changes

- Make `CoreSDKError` Equatable and make static constants in `CardError` and `PayPalError` public so merchants can do this:
```
if let error {
  if error == CardError.threeDSecureCanceledError {
      // handle cancel case
   } else {
    // handle other errors 
 }

```
### Checklist

- [x] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Demo/Demo/ViewModels/CardVaultViewModel.swift
+++ b/Demo/Demo/ViewModels/CardVaultViewModel.swift
@@ -47,7 +47,7 @@ class CardVaultViewModel: VaultViewModel {
                     )
                 )
             } else if let vaultError {
-                if CardError.isThreeDSecureCanceled(vaultError) {
+                if let error = vaultError as? CoreSDKError, error == CardError.threeDSecureCanceledError {
                     print("Canceled")
                     self.state.updateSetupTokenResponse = .idle
                 } else {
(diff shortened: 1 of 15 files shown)


KunJeongPark:
I will also make other Error enums (eg. `NetworkClientErrors`) and static constants public. 

I haven't seen a case where merchant needs to act on any error other than cancellation, but I think it's consistent to make all error types public.

Demo/Demo/ViewModels/CardVaultViewModel.swift
```diff
@@ -47,7 +47,7 @@ class CardVaultViewModel: VaultViewModel {
                     )
                 )
             } else if let vaultError {
-                if CardError.isThreeDSecureCanceled(vaultError) {
+                if let error = vaultError as? CoreSDKError, error == CardError.threeDSecureCanceledError {
```
sshropshire, line 50: q: is it possible to make `vaultError` a `CoreSDKError?` to prevent needing a cast here?

sshropshire (approved):
This is awesome! 🔥

Demo/Demo/ViewModels/CardVaultViewModel.swift
KunJeongPark: yes, good call. Actually, `clientID` fetching can also throw an error in the vault function above, so that's why I cast it here.

PR #304, opened 2024-11-25
[DTPPMOBILE-143] Deprecate Delegation Methods in Paypal SDK

GMALKHA:
### Reason for changes

The PayPal SDK is transitioned from the delegation pattern to the completion handler approach for handling asynchronous operation. We want to mark the existing delegation methods as deprecated in the 1.X to notify merchants about upcoming change. 

### Summary of changes
https://github.com/paypal/paypal-ios/pull/301
I used the migration-merged PR to deprecate all the methods and protocols that have been removed or renamed. 

### Checklist

- [ Deprecating Delegation]
- [ Deprecating Renamed methods] 

### Authors

- @GMALKHA

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,4 +34,5 @@ jobs:
         run: cd SampleApps/SPMTest && swift package resolve
       - name: Build & archive SPMTest
         run: xcodebuild -project 'SampleApps/SPMTest/SPMTest.xcodeproj' -scheme 'SPMTest' clean build archive CODE_SIGNING_ALLOWED=NO
+        run: pod lib lint --allow-warnings="deprecated"
 
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -43,6 +43,7 @@ public class CardClient: NSObject {
     /// If `didAttempt3DSecureVerification` is `true`, check verification status with `/v3/vault/setup-token/{id}` in your server.
     /// - Parameters:
     ///   - vaultRequest: The request containing setupTokenID and card
+    @available(*, deprecated, message: "This method is deprecated and will be removed in a future release. Use the new completion handler-based approach instead.")
     public func vault(_ vaultRequest: CardVaultRequest) {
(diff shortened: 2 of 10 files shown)


Sources/CardPayments/CardClient.swift
```diff
@@ -43,6 +43,7 @@ public class CardClient: NSObject {
     /// If `didAttempt3DSecureVerification` is `true`, check verification status with `/v3/vault/setup-token/{id}` in your server.
     /// - Parameters:
     ///   - vaultRequest: The request containing setupTokenID and card
+    @available(*, deprecated, message: "This method is deprecated and will be removed in a future release. Use the new completion handler-based approach instead.")
```
KunJeongPark: Maybe we can specify that it will be removed in 2.0.0 and link them to the migration guide in our repo. Something like => See our v2 migration guide in 
https://github.com/paypal/paypal-ios/

KunJeongPark:
Thank you for addressing this deprecation issue. It looks like you have some lint failures with line lengths. 
The Cocoapods failures on these deprecation warnings, let's look into if we can selectively silence these warnings.

Sources/CardPayments/CardClient.swift
GMALKHA: Thanks, Victoria! That’s a very good point. I was also thinking about that. Let me implement it.

KunJeongPark: Have you tried  adding options `run: pod lib lint --allow-warnings="deprecated"` in build.yml?

.github/workflows/build.yml
```diff
@@ -34,4 +34,5 @@ jobs:
         run: cd SampleApps/SPMTest && swift package resolve
       - name: Build & archive SPMTest
         run: xcodebuild -project 'SampleApps/SPMTest/SPMTest.xcodeproj' -scheme 'SPMTest' clean build archive CODE_SIGNING_ALLOWED=NO
+        run: pod lib lint --allow-warnings="deprecated"
```
KunJeongPark, line 37: Nice!

KunJeongPark (approved):

KunJeongPark:
🎉

PR #307, opened 2024-11-26
DTPP-11 Demo app organize folder structure

GMALKHA:
### Reason for changes
Organize folder structure
### Checklist

- [x] Organizing Demo app folder structure


Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -227,6 +227,78 @@
...
+			);
+			path = CardPaymentViewModel;
+			sourceTree = "<group>";
+		};
+		1001E2BC2CF644BE0023A03C /* PayPalVaultViewModel */ = {
+			isa = PBXGroup;
+			children = (
+				3BB60B542B1FA00C00A298CF /* PayPalVaultViewModel.swift */,
+				3BA0A58A2B1E240300330681 /* VaultViewModel.swift */,
+				3B2027402A8A72050007907E /* VaultState.swift */,
```
KunJeongPark: VaultViewModel is a parent class and VaultState is a published state in VaultViewModel, maybe they could be in their own folder along with common vault views - the ones that are done on merchant server?

KunJeongPark:
Looks so clean! Let me test things out, just to make sure nothing is missing and will give it a ✅
Also, we don't link the tickets directly in this repo, just the number like this: DTPPMOBILE-11
I link the PR in the ticket.

Demo/Demo.xcodeproj/project.pbxproj
KunJeongPark: Screenshot 2024-11-26 at 11 04 24 AM
The PayPalVault -> PayPalVaultViewModel => VaultState, VaultViewModel
PayPalVault -> views under VaultViews, these are common views for merchant side set up for all vault flows.

Could go under just "Vault" folder

KunJeongPark: oh I meant outside of the PayPalVault folder. Maybe just "Vault" folder with view models and views. There is PayPal vault, Card vault and other payment methods in the future.
And how about we order the views in sequence of expected events of the main view?

KunJeongPark: so for example under CardPayments, there are container views like CardPaymentView (which is not a good name, originally it was going to be container view for all the subviews like in CardVault) which has create order, order result result, the SDK portion, approve and result and container view for order completion with capture/authorize and result view

GMALKHA: > so for example under CardPayments, there are container views like CardPaymentView (which is not a good name, originally it was going to be container view for all the subviews like in CardVault) which has create order, order result result, the SDK portion, approve and result and container view for order completion with capture/authorize and result view

maybe i can change the name to CardViewComponents or CardPaymentsScreens ?

KunJeongPark:
Looks great Georgi! 🎉

KunJeongPark (approved):

Demo/Demo.xcodeproj/project.pbxproj
KunJeongPark: CardPaymentView => that actually is the main container view.
CreateOrderCardPaymentView
OrderCreateCardResultView
CardOrderApproveView
CardApprovalResultView
CardPaymentOrderCompletionView
CardOrderActionButton
CardOrderCompletionResultView

KunJeongPark: I think CardPaymentView is ok, consistent with PayPalVaultView. I like CardPaymentsScreens ✅. 
but we have PayPalWebButtonsView as main container view in PayPalWebPaymentsView.

KunJeongPark: Not too particular, I think having the correct order would make it easier to read.

KunJeongPark: Sure! I prefer CardPaymentsScreens over CardViewComponents because it's clearer about which module the container view is for. And that view, imo should be at the very top of the view folder.

PR #308, opened 2024-12-02
Silence Font Registration Warning

KunJeongPark:
### Summary of changes

- Move configure function with PayPal font registration to init function to silence failure to register
   already registered font
- Modify extension to UIFont to return on already registered font

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/PaymentButtons/PaymentButtonFont.swift
+++ b/Sources/PaymentButtons/PaymentButtonFont.swift
@@ -12,6 +12,10 @@ extension UIFont {
         var errorRef: Unmanaged<CFError>?
         let frameworkBundle = Bundle(for: PaymentButton.self)
 
+        guard UIFont(name: name, size: 10.0) == nil else {
+            return
+        }
+
         guard
             let pathForResourceString = frameworkBundle.path(forResource: name, ofType: fileExtension),
             let fontData = NSData(contentsOfFile: pathForResourceString),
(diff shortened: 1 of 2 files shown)


Sources/PaymentButtons/PaymentButtonFont.swift
```diff
@@ -12,6 +12,10 @@ extension UIFont {
         var errorRef: Unmanaged<CFError>?
         let frameworkBundle = Bundle(for: PaymentButton.self)
 
+        guard UIFont(name: name, size: 10.0) == nil else {
+            return
+        }
+
```
KunJeongPark, line 18: If there is failure to instantiate this font, it most likely means the font is not registered, so it goes through with registering this font, otherwise it returns without registering.

sshropshire (approved):

Sources/PaymentButtons/PaymentButtonFont.swift
GMALKHA: I was thinking if it would make sense to call it from the AppDelegate, so the registration logic is available globally throughout the app. This way i think we wouldn’t need to register it again in individual views later. 
 Just thinking out loud :)

GMALKHA: I think this font is used only in a few specific views, so it makes sense to initialize it only in the PaymentButtonFont.

GMALKHA (approved):

PR #311, opened 2024-12-03
[DTPP-142] PayPalWebPayments show stages of order

GMALKHA:
### Summary of changes
Refactored PaypalWebViewModel for better state tracking, updated PayPalButtonsView for progress handling, and fixed successView display issue.

### Checklist
- [x] Replaced centralized state management (updateState) with PayPalPaymentState.
- [x] Added individual LoadingState properties for createOrderResponse and approveResultResponse.
- [x] Updated the UI in PayPalButtonsView to check the appropriate LoadingState for progress or error handling.
- [x] Introduced properties like createOrderResponse, approveResultResponse, captureOrderResponse, and authorizedOrderResponse in PaypalPayments for managing states independently.
- [x] Verified that the correct view is displayed after creating an order.

-  @GMALKHA


Demo/Demo/CardPayments/CardPaymentViewModel/CardPaymentState.swift
```diff
@@ -49,3 +49,50 @@ struct CardPaymentState: Equatable {
         }
     }
 }
+
+struct PayPalPaymentState: Equatable {
```
GMALKHA: It needs to be moved to a separate file.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -44,25 +44,75 @@ class PayPalWebViewModel: ObservableObject {
...
                 selectedMerchantIntegration: DemoSettings.merchantIntegration
             )
-
-            updateOrder(order)
-            updateState(.success)
+            DispatchQueue.main.async {
+                self.state.createdOrderResponse = .loaded(order)
+            }
+//            updateOrder(order)
+//            updateState(.success)
```
KunJeongPark: did you mean to delete this

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebCreateOrderView.swift
```diff
@@ -40,7 +40,10 @@ struct PayPalWebCreateOrderView: View {
                     }
                 }
                 .buttonStyle(RoundedBlueButtonStyle())
-                if payPalWebViewModel.state == .loading && payPalWebViewModel.checkoutResult == nil && payPalWebViewModel.orderID == nil {
+//                if payPalWebViewModel.state == .loading && payPalWebViewModel.checkoutResult == nil && payPalWebViewModel.orderID == nil {
```
GMALKHA: Will need to refactore and remove commented lines

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebResultView.swift
```diff
@@ -5,44 +5,41 @@ struct PayPalWebResultView: View {
...
-                LabelViewText("Status:", bodyText: status)
-            }
+            
+            LabelViewText("Order ID:", bodyText: createOrderResponse.id)
+            
+            LabelViewText("Status:", bodyText: createOrderResponse.status)
 
             if let payerID = payPalWebViewModel.checkoutResult?.payerID {
-                LabelViewText("Payer ID:", bodyText: payerID)
+                LabelViewText("Payer ID:", bodyText: createOrderResponse.status)
```
GMALKHA: instead of status need to pass payerID

Demo/Demo/CardPayments/CardPaymentViewModel/CardPaymentState.swift
```diff
@@ -49,3 +49,50 @@ struct CardPaymentState: Equatable {
...
+
```
KunJeongPark: Shouldn't this be in PayPalWebPayments folder? (It can go in view model folder)

KunJeongPark:
I think I expected the result of the order creation on the view right below the create order form and above checkout with PayPal.

KunJeongPark:
Pressing on PayPal checkout button is stuck in loading indicator of Create order button. I'll take a closer look at what's going on.
I think what I expected was separate results view after each state. That was my intention for wanting this, so we can compare results screens of each step.

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebButtonsView.swift
```diff
@@ -39,7 +39,7 @@ struct PayPalWebButtonsView: View {
                             payPalWebViewModel.paymentButtonTapped(funding: .paypal)
                         }
                     }
-                    if payPalWebViewModel.state == .loading &&
+                    if payPalWebViewModel.state.createdOrderResponse == .loading &&
```
KunJeongPark: I think this should be approveResultResponse

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -44,25 +44,73 @@ class PayPalWebViewModel: ObservableObject {
...
+            print("❌ Failed to authorize order: \(error.localizedDescription)")
+        }
+    }
 
     func paymentButtonTapped(funding: PayPalWebCheckoutFundingSource) {
         Task {
             do {
-                self.updateState(.loading)
+                DispatchQueue.main.async {
+                    self.state.createdOrderResponse = .loading
```
KunJeongPark: This should be approveResultResponse

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -73,22 +121,26 @@ class PayPalWebViewModel: ObservableObject {
...
                         if let error {
-                            if error == PayPalError.checkoutCanceledError {
-                                print("Canceled")
-                                self.updateState(.idle)
-                            } else {
-                                self.updateState(.error(message: error.localizedDescription))
+                            DispatchQueue.main.async {
+                                if error == PayPalError.checkoutCanceledError {
+                                    print("Canceled")
+                                    self.state.createdOrderResponse = .idle
```
KunJeongPark: same here

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -73,22 +121,26 @@ class PayPalWebViewModel: ObservableObject {
...
+                                } else {
+                                    self.state.createdOrderResponse = .error(message: error.localizedDescription)
```
KunJeongPark: same here

KunJeongPark: Since you created different responses for each stage or the order, you need to repurpose the result view or have different result views for each step.

Right now, you have a resultView that takes createdOrderResponse

GMALKHA: Good catch . I will push a commit to refactor it.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -73,22 +121,26 @@ class PayPalWebViewModel: ObservableObject {
...
+                                    self.state.approveResultResponse = .idle
+                                } else {
+                                    self.state.approveResultResponse = .error(message: error.localizedDescription)
+                                }
                             }
                         } else {
-                            self.updateState(.success)
-                            self.checkoutResult = result
+                            DispatchQueue.main.async {
+                                self.state.createdOrderResponse = .loaded(Order(id: orderID, status: "COMPLETED"))
```
KunJeongPark: this needs to be approveResultResponse as well

KunJeongPark:
Looks good so far, nit picks:
1. this result view could be formatted the same as other views:


2. Loading indicator on the buttons on checkout with PayPal action
3. "Authorize Order" button seems unresponsive - this is because you are acting only on authorized option but observing capture order response on the PayPalWebTransactionView
4. Your viewModel's completeTransaction is only handling authorize option, not capture.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,15 +152,17 @@ class PayPalWebViewModel: ObservableObject {
...
+            }
             print("❌ failed to create PayPalWebCheckoutClient with error: \(error.localizedDescription)")
             return nil
         }
     }
 
     func completeTransaction() async throws {
         do {
-            updateState(.loading)
+            self.state.authorizedOrderResponse = .loading
```
KunJeongPark: This needs to be on main thread

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -44,51 +44,103 @@ class PayPalWebViewModel: ObservableObject {
...
+            DispatchQueue.main.async {
+                self.state.createdOrderResponse = .error(message: error.localizedDescription)
+            }
             print("❌ failed to fetch orderID with error: \(error.localizedDescription)")
         }
     }
+    
+    func captureOrder(orderID: String, selectedMerchantIntegration: MerchantIntegration) async {
+        do {
+            self.state.capturedOrderResponse = .loading
```
KunJeongPark: this needs to be on main

KunJeongPark:
I'm trying to figure out why your result views for order completion is not appearing while it is working on main.
I see, old updatOrder assigned self.order which assigned OrderID and we are not doing it anymore so completeOrder is not getting called(as it is checking for orderID) We don't need to check for orderID anymore since it is handled in the states.

GMALKHA:
> I'm trying to figure out why your result views for order completion is not appearing while it is working on main. I see, old updatOrder assigned self.order which assigned OrderID and we are not doing it anymore so completeOrder is not getting called(as it is checking for orderID) We don't need to check for orderID anymore since it is handled in the states.

Ah, that was actually something I was planning to refactor yesterday. Right, the orderId is no longer used, and we need to remove a check from there.

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebTransactionView.swift
```diff
@@ -19,7 +19,7 @@ struct PayPalWebTransactionView: View {
                 .buttonStyle(RoundedBlueButtonStyle())
                 .padding()
 
-                if payPalWebViewModel.state == .loading {
+                if payPalWebViewModel.state.capturedOrderResponse == .loading {
```
KunJeongPark: I would add || for authorizedOrderResponse loading

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,47 +154,41 @@ class PayPalWebViewModel: ObservableObject {
...
     }
 
     func completeTransaction() async throws {
         do {
-            updateState(.loading)
-
-            let payPalClientMetadataID = payPalDataCollector?.collectDeviceData()
-            if let orderID {
+            DispatchQueue.main.async {
+                self.state.authorizedOrderResponse = .loading
```
KunJeongPark: you need to check the intent and if it's authorize, then use this state, otherwise use the capturedOrderResponse. Same for all the other states. you can just call a function like set loading state which checks the intent and sets appropriate state.
Same with all the other state setting in this function.

GMALKHA:
> I'm trying to figure out why your result views for order completion is not appearing while it is working on main. I see, old updatOrder assigned self.order which assigned OrderID and we are not doing it anymore so completeOrder is not getting called(as it is checking for orderID) We don't need to check for orderID anymore since it is handled in the states.

The loading indicator wasn’t showing because it was tracking the orderID computed property, but it wasn’t assigned in the ViewModel and was always nil. I’ll push a commit to set the orderID

The approve result view is not appearing because the state for approveResultResponse always defaults to idle instead of transitioning to loaded. I'm looking into that.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,47 +155,80 @@ class PayPalWebViewModel: ObservableObject {
...
+            switch self.intent {
+            case .authorize:
+                self.state.authorizedOrderResponse = .loading
+            case .capture:
+                self.state.capturedOrderResponse = .loading
+            }
+        }
+    }
+    
+    private func setLoadingState(order: Order) {
```
KunJeongPark: I would rename this setOrderCompletionLoadedState or something specific

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,47 +155,80 @@ class PayPalWebViewModel: ObservableObject {
...
             return nil
         }
     }
 
     func completeTransaction() async throws {
         do {
-            updateState(.loading)
+            setLoadingState()
+            
+            try await processIntent(orderID: orderID ?? "")
```
KunJeongPark: I think you are calling the auth/capture twice.
once here and in completeOrder. Once you comment out the processIntent (which is essentially doing the same thing as completeOrder), I see the 
success view. You just need to manage the scroll down. And the spacing looks a bit big between the "Authorize/capture" button and the success view. But that's a really nit pick

I see lint errors from just not leaving a space in your structs or classes 
and one super long line

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -44,51 +44,106 @@ class PayPalWebViewModel: ObservableObject {
...
+                                    self.state.approveResultResponse = .idle
+                                } else {
+                                    self.state.approveResultResponse = .error(message: error.localizedDescription)
+                                }
                             }
                         } else {
-                            self.updateState(.success)
-                            self.checkoutResult = result
+                            DispatchQueue.main.async {
+                                self.state.approveResultResponse = .loaded(PayPalPaymentState.ApprovalResult(id: orderID, status: "COMPLETED"))
```
KunJeongPark: I think this is actually an error that was there before, the status on success url should be "APPROVED" not "COMPLETED"

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
GMALKHA: Gotcha, in this case, authorizeOrder and captureOrder functions are no longer needed since completeOrder is verifying the intent. these two methods can be removed.

KunJeongPark: I mean your processIntent function and call should be removed

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
KunJeongPark: Looks good now! I would just resolve the 7 lint warnings.

KunJeongPark:
Hey Georgi, it looks good so far, just doing some testing. Can you order the views in order they appear?
The main container view, and then order of appearance?

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -44,51 +50,64 @@ class PayPalWebViewModel: ObservableObject {
...
+                                print("✅ Checkout result: \(String(describing: result))")
+                            }
                         }
                     }
                 }
-                updateState(.success)
             } catch {
                 print("Error starting PayPalWebCheckoutClient")
-                updateState(.error(message: error.localizedDescription))
+                self.state.createdOrderResponse = .error(message: error.localizedDescription)
```
KunJeongPark: I think this needs to be on main also

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,47 +119,68 @@ class PayPalWebViewModel: ObservableObject {
             payPalDataCollector = PayPalDataCollector(config: config)
             return payPalClient
         } catch {
-            updateState(.error(message: error.localizedDescription))
+            DispatchQueue.main.async {
+                self.state.createdOrderResponse = .error(message: error.localizedDescription)
```
KunJeongPark, line 125: I guess this works, can be used for not getting configuration.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -100,47 +119,68 @@ class PayPalWebViewModel: ObservableObject {
...
             print("Error with \(intent) order: \(error.localizedDescription)")
         }
     }
-
-    func resetState() {
-        updateState(.idle)
-        order = nil
-        checkoutResult = nil
+    
+    private func setLoadingState() {
```
KunJeongPark, line 150: Nice!

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
KunJeongPark: Just a little fix here and it looks really good Georgi. Great work!
Thanks you!

KunJeongPark:
🎉 Great job Georgi!

KunJeongPark (approved):

PR #312, opened 2024-12-06
CardClient/ PayPalWebCheckout Analytics Audit

KunJeongPark:
### Summary of changes

Change analytic event names for accuracy and consistency with Android

### Checklist

- @KunJeongPark @sdhopati

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -45,7 +45,7 @@ public class CardClient: NSObject {
     ///   if it fails, `CardVaultResult will be `nil` and `error` will describe the failure
     public func vault(_ vaultRequest: CardVaultRequest, completion: @escaping (CardVaultResult?, CoreSDKError?) -> Void) {
         analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
-        analyticsService?.sendEvent("card-payments:vault-wo-purchase:started")
+        analyticsService?.sendEvent("card-payments:vault:started")
         Task {
             do {
                 let result = try await vaultAPI.updateSetupToken(cardVaultRequest: vaultRequest).updateVaultSetupToken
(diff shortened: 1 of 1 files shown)


Sources/CardPayments/CardClient.swift
```diff
@@ -113,19 +113,15 @@ public class CardClient: NSObject {
                         return
                     }
                 
-                    analyticsService?.sendEvent("card-payments:3ds:confirm-payment-source:challenge-required")
+                    analyticsService?.sendEvent("card-payments:approve-order:auth-challenge-required")
```
sshropshire, line 116:
```suggestion
                    analyticsService?.sendEvent("card-payments:approve-order:payer-action-required")
```
I wonder if we should start aligning our naming conventions with PPaaS more? I've been back and forth for a while, but it may be helpful for migrating merchants with direct integrations to our SDK since they'll be familiar with the terminology.

It may be out of scope for this PR, but it applies throughout the SDK e.g. we could rename `CardClient.approveOrder()` to `CardClient.confirmPaymentSource()` etc. Does that sound worthwhile?

KunJeongPark: It was a bit confusing. For some reason we had both in iOS, `confirm-payment-source` events and `approve-order` final events. So we simplified it to just `approve-order`, following Android implementation.

I thought about the naming and it'd be a bit confusing to have convention of API name because for instance, PayPal web module just presents the pages. I think following function name as you did makes sense.

sshropshire: Right yeah it's the same for us on Android. PayPal always shows an auth challenge. On the other hand, for Android Card we have a 2-step process since not all payment source confirmations require 3DS. They may succeed on the first `CardClient.approveOrder()`. It's also a 2-step process on Android mostly because of process kill.

KunJeongPark:
We are going to close this for now to list events we are sending for both iOS and Android and align on which events to send.

PR #316, opened 2025-01-17
Return Result type instead of tuple of optionals

KunJeongPark:
### Summary of changes

- Return result type with .success and .failure rather than optional of result and errors.

### Checklist

- [x] Added a changelog entry

- @KunJeongPark


sshropshire (approved):
🔥

PR #321, opened 2025-03-19
Card Form in PayPalWebPayments

KunJeongPark:
### Summary of changes

- BCDC using existing PayPalWebPayments module instead of creating a new one
- Add button changes for BCDC, logo image and new `CardButton`
- Demo app changes to enter in card as funding source
- Add card in `PayPalWebCheckoutFundingSource` enum
   this enables `start` function to distinguish the flows for API call (we need to get specs for)
- UpdateClientConfig GraphQL endpoint for card fundingSource and unit tests
  

### Checklist

- [x] Added a changelog entry

- @KunJeongPark


nirvan543:
I like including the branded card checkout within the `PayPalWebPayments` module because:

- It's a web-based payment method option and it fits with the other web-based payment method options (PayPal, Pay Later, Credit)
- Since it's branded card processing and it's done via web, it naturally fits as an extension to the existing `PayPalWebPayments` module.
- Doing it this way will reduce the amount of code we need to add (both to the SDK and demo app) and thus reduces the amount of code that the merchant has to do (and reduces their file size).
- Since this is a short-term solution, having it as an extension to the existing module feels right.

When we build the native card fields solution, we can put it in it's own module because it will encompass BCDC, ACDC, and the other payment method options (PayPal, Pay Later, Credit, and more 👀 ).

Sources/PaymentButtons/CardButton.swift
```diff
@@ -0,0 +1,131 @@
...
+/// Configuration for Card button
+public final class CardButton: PaymentButton {
+
+    /// Available colors for CardButton.
+    public enum Color: String {
+        case gold
+        case white
+        case black
+        case silver
+        case blue
```
nirvan543: This button will only ever be black so we can remove the other options.

Sources/PaymentButtons/CardButton.swift
```diff
@@ -0,0 +1,131 @@
...
+        /// Display "Pay with" on the left side of the button's logo
+        case payWith = "Pay with"
+
+        /// Display "Debit or Credit Card" on the left side of the button's logo
+        case card = "Debit or Credit Card"
+
+        var label: PaymentButtonLabel? {
+            PaymentButtonLabel(rawValue: rawValue)
+        }
+    }
```
nirvan543: We can remove every other option for the label and keep "Debit or Credit Card". BCDC doesn't support the other labels.

Sources/PaymentButtons/CardButton.swift
nirvan543: I lied, we also support a white BCDC button. I'll find the black card icon for this.

KunJeongPark:
@nirvan543 yes i think it makes sense to have it here.
We just need to call the APi for the card funding option.

I considered  having it as a separate library thinking that might be a new feature from the merchant/customer's point of view and something we could keep the same function signature for when we implement native.

I'm thinking about how we'd present this in our docs. If you want to be able to use card forms, we have guest checkout
option available through PayPalWebPayments - is that confusing?

Is it confusing for people who want to use this module for PayPal checkout flows only to see "card" value in 
PayPalCheckoutFundingSource enum.

But I agree, this flow makes more sense from code perspective. As long as it doesn't cause confusion for the merchants, this definitely makes more sense.

UnitTests/PaymentButtonsTests/CardButton_Tests.swift
```diff
@@ -0,0 +1,38 @@
+import XCTest
+@testable import PaymentButtons
+
+class CardButton_Tests: XCTestCase {
+
+    // MARK: - CardButton for UIKit
+
+    func testInit_whenCardButtonCreated_hasUIImageFromAssets() {
+        let sut = CardButton()
+        XCTAssertEqual(sut.imageView?.image, UIImage(named: "card-black"))
```
KunJeongPark, line 10: This test, I found out is not testing what it's supposed to test on all the other buttons as well.
It just passes because both values are nil. I will make a ticket to fix these tests.

UnitTests/PaymentButtonsTests/CardButton_Tests.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    func testInit_whenPayPalButtonCreated_hasDefaultUIValuess() {
+        let sut = CardButton()
+        XCTAssertEqual(sut.edges, PaymentButtonEdges.softEdges)
+        XCTAssertEqual(sut.size, PaymentButtonSize.collapsed)
+        XCTAssertEqual(sut.color, PaymentButtonColor.black)
+        XCTAssertEqual(sut.label, PaymentButtonLabel.card)
```
KunJeongPark, line 18: Because this button logo doesn't have text, i defaulted it to having the label.
So collapsed and mini don't show "Debit or Credit Card" label text
Full and expanded show the label text.
All the other buttons default to collapsed, those buttons have PayPal text embedded in the logo, I'm wondering default size on the card button should be the collapsed or something else.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -40,55 +49,66 @@ public class PayPalWebCheckoutClient: NSObject {
...
-        let baseURLString = config.environment.payPalBaseURL.absoluteString
-        let payPalCheckoutURLString =
+        Task {
+            if request.fundingSource == .card {
+                do {
+                    let configResult = try await clientConfigAPI.updateClientConfig(request: request)
+                    print("configResult: \(configResult)")
+                } catch {
+                    print("error in calling graphQL: \(error.localizedDescription)")
+                    self.notifyCheckoutFailure(with: PayPalError.payPalURLError, completion: completion)
```
KunJeongPark: Probably should not return with failure. Not sure about situation where this call would fail.

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
```diff
@@ -33,6 +33,17 @@ extension PayPalCreditButton.Color {
     }
 }
 
+extension CardButton.Color {
```
sshropshire, line 36: q: Are we able to use `CaseIterable` to autogenerate this code? link: [docs.swift.org](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/enumerations/#Iterating-over-Enumeration-Cases)

Sources/CorePayments/Networking/Enums/HTTPHeader.swift
```diff
@@ -7,4 +7,5 @@ public enum HTTPHeader: String {
     case authorization = "Authorization"
     case contentType = "Content-Type"
     case origin = "Origin"
+    case paypalClientContext = "paypal-client-context"
```
sshropshire:
```suggestion
    case payPalClientContext = "paypal-client-context"
```
nit: I'm honestly back and forth on which version is best.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -40,55 +49,66 @@ public class PayPalWebCheckoutClient: NSObject {
...
+                }
+            }
+            let baseURLString = config.environment.payPalBaseURL.absoluteString
+            let payPalCheckoutURLString =
             "\(baseURLString)/checkoutnow?token=\(request.orderID)" +
             "&fundingSource=\(request.fundingSource.rawValue)"
```
```suggestion
                "\(baseURLString)/checkoutnow?token=\(request.orderID)" +
                "&fundingSource=\(request.fundingSource.rawValue)"
```
quickfix: indentation.

sshropshire (approved):
🔥

Demo/Demo/Extensions/PaymentButtonEnums+Extension.swift
KunJeongPark: The extensions in the demo app were done that way because those enums are declared in the SDK and we can't add CaseIterable conformance from outside the module. This iteration is not something the user would do, but kind of specific to our demo app to showcase all possible button appearances

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
```diff
@@ -26,7 +24,7 @@ class UpdateClientConfigAPI {
 
     // MARK: - Internal Methods
 
-    func updateClientConfig(request: PayPalWebCheckoutRequest) async throws -> ClientConfigResponse {
+    public func updateClientConfig(orderID: String, fundingSource: String) async throws -> ClientConfigResponse {
```
KunJeongPark: This can be modified to pass in different values for user experience flow for other payment methods.

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
```diff
@@ -0,0 +1,71 @@
...
+
+    /// Exposed for injecting MockNetworkingClient in tests
+    init(coreConfig: CoreConfig, networkingClient: NetworkingClient) {
+        self.coreConfig = coreConfig
+        self.networkingClient = networkingClient
+    }
+
+    // MARK: - Internal Methods
+
+    public func updateClientConfig(orderID: String, fundingSource: String) async throws -> ClientConfigResponse {
```
nirvan543: Should we make the `fundingSource` type be the enum from CorePayments? That way, we can ensure the value is valid.

Sources/PaymentButtons/PaymentButton+ImageAsset.swift
```diff
@@ -10,7 +10,11 @@ extension PaymentButton {
     
     var imageAccessibilityLabel: String {
         // NEXT_MAJOR_VERSION: - To be replaced with translation strings.
-        fileName.starts(with: "credit") ? "PayPal Credit" : "PayPal"
+        if fundingSource == .card {
+            return "Credit Card"
+        }
+
+        return fileName.starts(with: "credit") ? "PayPal Credit" : "PayPal"
```
nirvan543, line 17: What about Pay Later?

KunJeongPark: I didn't change that line but appears to be an oversight.

KunJeongPark: I think this is an image accessibility label, not the text portion of the button, so I think they would be the same for both PayPal and Pay Later.
These are still the old buttons with image and text label separate

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
KunJeongPark: In the start function, we call updateClientConfig by passing in raw value of  PayPalWebCheckoutRequest's fundingSource enum. I think I prefer to keep this a sting and to pass in raw value of the enums from each module

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
```diff
@@ -0,0 +1,68 @@
...
+                )
+            }
+        """
+
+        let variables = UpdateClientConfigVariables(
+            orderID: orderID,
+            fundingSource: fundingSource,
+            integrationArtifact: "MOBILE_SDK",
+            userExperienceFlow: "INCONTEXT",
+            productFlow: "HERMES",
```
sshropshire, line 54:
```suggestion
            productFlow: "MOBILE_SDK",
```
~Should this be `MOBILE_SDK` as well? I had to make this change on Android.~

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -40,55 +49,66 @@ public class PayPalWebCheckoutClient: NSObject {
...
+                self.notifyCheckoutFailure(with: PayPalError.payPalURLError, completion: completion)
+                return
+            }
 
-                if let url = url {
-                    guard let orderID = self.getQueryStringParameter(url: url.absoluteString, param: "token"),
-                    let payerID = self.getQueryStringParameter(url: url.absoluteString, param: "PayerID") else {
-                        self.notifyCheckoutFailure(with: PayPalError.malformedResultError, completion: completion)
-                        return
+            webAuthenticationSession.start(
```
sshropshire, line 74: 🧶 : Should `webAuthenticationSession.start()` be called outside of the `Task {}` block?

UnitTests/CorePaymentTests/UpdateClientConfigAP_Tests.swift
```diff
@@ -0,0 +1,89 @@
...
+            XCTAssertEqual(error.code, 123)
+            XCTAssertEqual(error.localizedDescription, "error-desc")
+        }
+    }
+
+    func testUpdateSetupToken_whenSuccess_returnsParsedUpdateSetupTokenResponse() async throws {
+        let successsResponseJSON = """
+        {
+            "data": {
+            "updateClientConfig": true
```
sshropshire:
```suggestion
                "updateClientConfig": true
```
🧶 : indentation

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
KunJeongPark: No, we talked to Risk and Compliance and they wanted HERMES which is the default
They felt integration_artifact is the value that should distinguish our SDK

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: In iOS, we are awaiting the API call before launching the web session. In order to make sure that this happens in sequential order, the web session has to be started within the Task block - since only inside a Task, we can use await to suspend the execution until asynchronous call completes.

sshropshire: Interesting ok. We're doing async like web on Android, and a synchronous presentation of the CCT. It might be a better user experience if the `ASWebAuthentication` is shown immediately, but that's up to y'all on iOS. Just a suggestion.

PR #347, opened 2025-08-28
Add UpdateClientConfig to paypal vault

KunJeongPark:
### Reason for changes
Call graphQL endpoint UpdateClientConfig for PayPalWebClient vault function
to propagate our CCO values

### Summary of changes

- call graphQL endpoint UpdateClientConfig for vault function

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -155,55 +155,66 @@ public class PayPalWebCheckoutClient: NSObject {
     public func vault(_ vaultRequest: PayPalVaultRequest, completion: @escaping (Result<PayPalVaultResult, CoreSDKError>) -> Void) {
         analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
         analyticsService?.sendEvent("paypal-web-payments:vault-wo-purchase:started")
-        
-        var vaultURLComponents = URLComponents(url: config.environment.paypalVaultCheckoutURL, resolvingAgainstBaseURL: false)
-        let queryItems = [URLQueryItem(name: "approval_session_id", value: vaultRequest.setupTokenID)]
-        vaultURLComponents?.queryItems = queryItems
-        
-        guard let vaultCheckoutURL = vaultURLComponents?.url else {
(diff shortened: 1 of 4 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -155,55 +155,66 @@ public class PayPalWebCheckoutClient: NSObject {
...
-                        !tokenID.isEmpty, !approvalSessionID.isEmpty
-                    else {
-                        self.notifyVaultFailure(with: PayPalError.payPalVaultResponseError, completion: completion)
-                        return
+        Task {
+            do {
+                _ = try await clientConfigAPI.updateClientConfig(
+                    token: vaultRequest.setupTokenID,
+                    fundingSource: PayPalWebCheckoutFundingSource.paypal.rawValue
+                )
```
sshropshire, line 164: This will introduce a small amount of latency to the launch of `ASWebAuthenticationSession`. Is there a way to update CCO without having to `await` and suspend the `Task`? On Android, Karthik suggested making the CCO update without suspending to prevent adding extra wait time to the launch of the CCT.

KunJeongPark: We wanted to make sure that the patch was done before the webpage was launched.
Otherwise you might end up with a scenario where the CCO values are not detected in cart on xorouternodeweb and they try to do another patch. So if we kick off patching, and it is not detected in time, there might be another patching done in xorouternodeweb.

They have logic there where if none is detected, they provide default values

sshropshire: Oh ok. Should Android do the same thing?

KunJeongPark: I think Ravi had mentioned this in meeting and that's why we had opted to try to await it in Android.

KunJeongPark: I haven't noticed any lag.

sshropshire (approved):

kgangineni (approved):

PR #350, opened 2025-09-05
patchCCOWithAppSwitchEligibility

KunJeongPark:
### Reason for changes

Implement patchCCOWithAppSwitchEligibility to get eligibility and urls for app switch

### Summary of changes

- Add PatchCCOWithAppSwitchEligibiligy and PatchCcoWithAppSwitchEligibiligyVariables

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
+++ b/UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
@@ -1,0 +1,222 @@
+// swiftlint:disable indentation_width
+
+import Foundation
+import XCTest
+@testable import CorePayments
+@testable import TestShared
+
+final class PatchCCOWithAppSwitchEligibility_Tests: XCTestCase {
+
+    private var sut: PatchCCOWithAppSwitchEligibility!
+    private var mockNetworkingClient: MockNetworkingClient!
+    private var mockAuthSTS: MockAuthenticationSecureTokenServiceAPI!
+    private let coreConfig = CoreConfig(clientID: "fake-client-id", environment: .sandbox)
+    private let expectedQueryString = """
+        mutation PatchCcoWithAppSwitchEligibility(
+            $contextId: String!,
+            $experimentationContext: externalExperimentationContextInput,
+            $osType: externalOSType!,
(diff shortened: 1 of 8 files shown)


UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -0,0 +1,222 @@
...
+@testable import TestShared
+
+final class PatchCCOWithAppSwitchEligibility_Tests: XCTestCase {
+
+    private var sut: PatchCCOWithAppSwitchEligibility!
+    private var mockNetworkingClient: MockNetworkingClient!
+    private var mockAuthSTS: MockAuthenticationSecureTokenServiceAPI!
+    private let coreConfig = CoreConfig(clientID: "fake-client-id", environment: .sandbox)
+    private let expectedQueryString = """
+        mutation PatchCcoWithAppSwitchEligibility(
```
sshropshire, line 15: nit: Moving this query to a sharable constant would make this code more DRY.

sshropshire (approved):

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
KunJeongPark: Do you mean make the constant name more specific or make this constant available outside this test?
In this file, I could rename it patchCcoWithAppSwitchEligibilityQuery or have like enum TestQueries with static let patchCcoWithAppSwitchEligibility.

I just moved this here because of function length lint error.

kgangineni: no, in android we have assets, we can save plain text file, save query in plain text file and read it from there, not sure if it's possible in iOS

kgangineni (approved):

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
KunJeongPark: I've seen that before but I don't think we currently have that in iOS for testing. 
I will make a separate ticket for that.

PR #352, opened 2025-09-11
PayPalWebCheckout HandleReturnURL for App Switch

KunJeongPark:
### Reason for changes

- Add `handleReturnURL` function to handle url back from app switch

### Summary of changes

- Added `handleReturnURL` to `PayPalWebCheckout` to handle app switch return, unit tests
- This function will be extended to handle both checkout and vault returns in later PR along with distinction between
    `appSwitchCheckoutCompletion` and `appSwitchVaultCompletion`
    There will also be enums persisted after app switch launch of which flow was launched, 
    or discerned implicitly by which completion was persisted.
- This function allows testing of app switch flows for upcoming PRs: 
   - Toggling flows between web checkout and app switch checkout
   - Implementing the same flows for PayPal vault without purchase
   - Singleton class for routing incoming url's from app switch to appropriate class
   - Implementing controls to make sure no app switch attempts are made before initial launch completion

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -8,6 +8,8 @@ public class PayPalWebCheckoutClient: NSObject {
 
     let config: CoreConfig
 
+    var appSwitchCompletion: ((Result<PayPalWebCheckoutResult, CoreSDKError>) -> Void)?
+
     private let clientConfigAPI: UpdateClientConfigAPI
     private let webAuthenticationSession: WebAuthenticationSession
     private let networkingClient: NetworkingClient
@@ -234,6 +236,42 @@ public class PayPalWebCheckoutClient: NSObject {
                 }
             }
         }
+    }
+
+    // MARK: - App Switch Method
+
+    public func handleReturnURL(_ url: URL) {
+
+        guard let completion = appSwitchCompletion else { return }
+        defer { appSwitchCompletion = nil }
+
+        let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        let items = components?.queryItems ?? []
+        func queryValue(_ name: String) -> String? {
+            items.first { $0.name.caseInsensitiveCompare(name) == .orderedSame }?.value
+        }
+
+        let path = url.path.lowercased()
+
+        if path.contains("/cancel") {
+            notifyCheckoutCancelWithError(with: PayPalError.checkoutCanceledError, completion: completion)
+            return
+        }
+
+        if path.contains("/success"),
+        let orderID = queryValue("token"), !orderID.isEmpty,
+        let payerID = queryValue("PayerID") ?? queryValue("payer_id") ?? queryValue("payerId"), !payerID.isEmpty {
+            notifyCheckoutSuccess(for: PayPalWebCheckoutResult(orderID: orderID, payerID: payerID), completion: completion)
+            return
+        }
+
+        // get new error type
+        if path.contains("/fail") {
+            notifyCheckoutFailure(with: PayPalError.malformedResultError, completion: completion)
+            return
+        }
+
+        notifyCheckoutFailure(with: PayPalError.malformedResultError, completion: completion)
     }
 
     private func getQueryStringParameter(url: String, param: String) -> String? {
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -248,7 +248,7 @@ public class PayPalWebCheckoutClient: NSObject {
         let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
         let items = components?.queryItems ?? []
         func queryValue(_ name: String) -> String? {
-            items.first { $0.name.caseInsensitiveCompare(name) == .orderedSame }?.value
+            items.first { $0.name.compare(name) == .orderedSame }?.value
         }
 
         let path = url.path.lowercased()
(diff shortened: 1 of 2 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -236,6 +238,42 @@ public class PayPalWebCheckoutClient: NSObject {
...
+    public func handleReturnURL(_ url: URL) {
+
+        guard let completion = appSwitchCompletion else { return }
+        defer { appSwitchCompletion = nil }
+
+        let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        let items = components?.queryItems ?? []
+        func queryValue(_ name: String) -> String? {
+            items.first { $0.name.caseInsensitiveCompare(name) == .orderedSame }?.value
+        }
```
sshropshire: random q: is there a reason this function isn't declared at the `class` level?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -236,6 +238,42 @@ public class PayPalWebCheckoutClient: NSObject {
...
+
+        if path.contains("/success"),
+        let orderID = queryValue("token"), !orderID.isEmpty,
+        let payerID = queryValue("PayerID") ?? queryValue("payer_id") ?? queryValue("payerId"), !payerID.isEmpty {
+            notifyCheckoutSuccess(for: PayPalWebCheckoutResult(orderID: orderID, payerID: payerID), completion: completion)
+            return
+        }
+
+        // get new error type
```
sshropshire: q: what's the new error type mean?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: We have another function that takes a url and query param name that is defined at class level.
I wanted one that is specific for this function to return value from URLComponents

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
```diff
@@ -326,5 +326,100 @@ class PayPalClient_Tests: XCTestCase {
...
     }
+
+    // MARK: - handleReturnURL tests
+
+    func testHandleReturnURL_success_callsAppSwitchCompletionWithResult() {
+        var received: Result<PayPalWebCheckoutResult, CoreSDKError>?
+        payPalClient.appSwitchCompletion = { received = $0 }
+
+        let url = URL(string:
+            "https://appSwitchURL/success?token=ORDER123&PayerID=PAYER456&switch_initiated_time=1757431432185")!
```
sshropshire, line 337: test q: do we need to also cover use cases where the "payer id" parameter key is `payer_id` or `payerId`?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: It would be webSession error. For web checkout, we have PayPalError.webSession(detailedErrorFromDownStream).
I don't know yet if Venice app return url includes detailed error message in case of checkout failure.

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
KunJeongPark: I can include that!

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
```diff
@@ -326,5 +326,100 @@ class PayPalClient_Tests: XCTestCase {
...
+
+        payPalClient.handleReturnURL(url)
+
+        switch received {
+        case .success(let result)?:
```
sshropshire, line 342: Interesting I've never seen a switch statement like this. What does the question mark `?` do here?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire:
```suggestion
        // check for web session error
```
Ok yeah I'd modify the comment then to something like this. It's good to provide maximum context, we might need these bread crumbs in the future for maintenance.

sshropshire (approved):

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
KunJeongPark: received is an optional value and ? after .success(let result) means match .success only if optional received is not nil. It allows unwrapped statements after the success case branch.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: Good call

UnitTests/PayPalWebPaymentsTests/PayPalWebCheckoutClient_Tests.swift
KunJeongPark: I guess I'd need to change my queryValue function to be case sensitive, remove caseInsensitive option in compare

KunJeongPark: Addressed in https://github.com/paypal/paypal-ios/pull/352/commits/f081a7a6acd51dcb159c277bf416cdb8e399d35c

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
KunJeongPark: Addressed in https://github.com/paypal/paypal-ios/pull/352/commits/f081a7a6acd51dcb159c277bf416cdb8e399d35c

PR #353, opened 2025-09-12
Changes for toggling app switch for checkout

KunJeongPark:
### Reason for changes

- Implement app switch if opted in, eligible or no errors in app launch

### Summary of changes

- Create appSwitchIfEligible property in PayPalWebCheckoutRequest
- Toggle existing paypal web checkout with app switch if eligible flow
- If not app switch launched, fallback to web checkout flow

TODO: unit tests

### Checklist

- [ ] Added a changelog entry

- @KunJeongPark

### DIFF
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -73,6 +73,7 @@ public class PayPalWebCheckoutClient: NSObject {
 
                 case .fallback(let reason):
                     // TODO: analytics
+                    _ = reason
                     startWebCheckoutFlow(request: request, completion: completionOnce)
                 }
             } else {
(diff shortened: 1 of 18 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -274,6 +243,143 @@ public class PayPalWebCheckoutClient: NSObject {
...
+                    url: authURL,
+                    context: self,
+                    sessionDidDisplay: { [weak self] didDisplay in
+                        if didDisplay {
+                            self?.analyticsService?.sendEvent("paypal-web-payments:checkout:auth-challenge-presentation:succeeded")
+                        } else {
+                            self?.analyticsService?.sendEvent("paypal-web-payments:checkout:auth-challenge-presentation:failed")
+                        }
+                    },
+                    sessionDidComplete: { [weak self] url, error in
```
KunJeongPark, line 286: While I was testing vault web flow, I had a strange issue with infinite spinner upon return back to the demo app, completion was not being invoked.
I think it's because of the demo app not having a strong reference to the PayPalWebCheckoutClient
in the vaultViewModel in demo app. I faintly remember an inbound few years ago and this is the reason we removed [weak self] in sessionDidComplete block since we have no control over merchant integrations.
Removing weak self here solved the issue since the client with completion reference was not deallocated.

PR #371, opened 2026-02-09
Fix DocC generation simulator destination for CI compatibility

MikeThorntonPayPal:
### Reason for changes

Fixes multiple issues preventing documentation and test workflows from running successfully on GitHub Actions CI runners.

### 1. Documentation Generation Simulator Destination
The `ci` script was using a specific device destination (`platform=iOS Simulator,name=iPhone 17`) which doesn't work with `xcodebuild docbuild`. Documentation generation only needs a generic destination since it compiles code but doesn't execute it.

**Solution:** Changed to `generic/platform=iOS Simulator` for documentation builds.

### 2. Unit Tests Workflow Simulator Destination
The tests workflow was also using `iPhone 17` without specifying an OS version, causing it to fail on CI runners.

**Solution:** Updated to `platform=iOS Simulator,name=iPhone 17,OS=26.2` to explicitly specify both device and OS version.

### 3. GitHub Pages Branch Initialization
The publish-docs workflow had faulty logic for creating the `gh-pages` branch when it doesn't exist, causing the workflow to fail on first run.

**Solution:** Fixed the branch creation logic to properly handle the case where `gh-pages` doesn't exist yet.

### 4. Non-functional CSS Customization
Removed custom CSS injection code that attempted to hide module icons and add environment badges. The CSS selectors didn't work with DocC's JavaScript-generated DOM, and the code added unnecessary complexity.

**Solution:** Removed all CSS customization. Documentation now uses default DocC styling, ensuring local and GitHub Pages documentation appear identical.

### Checklist

- [x] Documentation workflow runs successfully on CI
- [x] Tests workflow runs successfully on CI
- [x] GitHub Pages deploys correctly
- [x] Documentation appears correctly on GitHub Pages

### Authors
- @michthornton


.github/workflows/publish-docs.yml
```diff
@@ -38,71 +38,15 @@ jobs:
...
       - name: Publish to GitHub Pages
         run: |
           # Switch to gh-pages branch and publish
-          git fetch origin gh-pages:gh-pages || git checkout --orphan gh-pages
-          git checkout gh-pages
+          if git fetch origin gh-pages:gh-pages 2>/dev/null; then
+            git checkout gh-pages
+          else
+            git checkout --orphan gh-pages
+            git rm -rf .
```
sshropshire, line 48: q: Does this line remove all files?

MikeThorntonPayPal: Yes, `git rm -rf .`  removes all files from the git index (staging area) in the current         
  directory and all subdirectories.                                                                                    
                                                                                                                       
  Here's what's happening in the code:                                                                                 
   ```                                                                                                                    
  if git fetch origin gh-pages:gh-pages 2>/dev/null; then                                                              
    git checkout gh-pages                                                                                              
  else                                                                                                                 
    git checkout --orphan gh-pages  # Creates a new orphan branch (no history)                                         
    git rm -rf .                     # Removes all files from git staging                                              
  fi                                                                                                                   
  ```
                                                                                                                     
  When creating a new orphan gh-pages branch for the first time:                                                       
  1. `git checkout --orphan gh-pages` creates a new branch with no history, but the working directory still contains all files from the previous branch                                                                                       
  2. `git rm -rf .` removes all those files from the staging area so the orphan branch starts clean                      
  3. The workflow then adds only the documentation files that belong on the gh-pages branch                            
                                                                                                                       
This is the "correct" way to initialize a clean gh-pages branch - you want to remove all the source code files and only publish the generated documentation (.github/workflows/publish-docs.yml:47-48).

sshropshire: Yeah that makes sense. I guess the docs aren't removed then? The way it reads it seems like the branch won't contain any files. I might be missing something here though.

MikeThorntonPayPal: Good catch. The git rm -rf . only runs when creating a brand new gh-pages branch (the else branch).                                                                                                             
                                                                                                                       
  After that line, the workflow immediately proceeds to copy the generated docs into the branch:                       
  - For preview deployments: copies to preview/ directory (line 83: cp -R docs/* preview/)                             
  - For production deployments: copies to the root (line 103: cp -R docs/* .)                                          
                                                                                                                       
  So the flow for a new branch is:                                                                                     
  1. Create orphan branch (empty)                                                                                      
  2. Remove any tracked files with git rm -rf .                                                                        
  3. Immediately populate with generated docs                                                                          
  4. Commit and push                                                                                                   
                                                                                                                       
  The git rm -rf . is necessary because git checkout --orphan creates an orphan branch but keeps all the files from the previous branch in the staging area. We need a clean slate before copying in the docs.                              
                                                                                                                       
  For existing branches (the if path), we just checkout and the subsequent logic handles replacing/updating the docs.

sshropshire (approved):

PR #380, opened 2026-03-31
HTTP and Networking Protocol Oriented Programming

sshropshire:
### Summary of changes

- Create `HTTPClient` protocol and `NetworkingClient` protocol to adhere to the best practice of Protocol Oriented Programming.
- Rename `HTTP` to `HTTPNetworkingClient`
- Simplify `MockHTTPClient` and `MockNetworkingClient` by making them implement the new protocols rather than subclassing concrete types. 
### Checklist

- @sshropshire

### DIFF
--- a/Sources/CorePayments/Networking/HTTPNetworkingClient.swift
+++ b/Sources/CorePayments/Networking/HTTPNetworkingClient.swift
@@ -1,13 +1,12 @@
 import Foundation
 
-/// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+/// This class is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
 ///
 /// `HTTPNetworkingClient` transforms REST & GraphQL requests into the proper HTTP format and executes network requests.
 @_documentation(visibility: private)
 public class HTTPNetworkingClient: NetworkingClient {
         
-    // MARK: - Internal Properties
-    
+    // MARK: - Private Properties
     private let http: HTTPClient
     private let coreConfig: CoreConfig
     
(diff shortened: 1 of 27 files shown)


Sources/CorePayments/Networking/HTTPNetworkingClient.swift
```diff
@@ -0,0 +1,120 @@
+import Foundation
+
+/// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
+///
+/// `HTTPNetworkingClient` transforms REST & GraphQL requests into the proper HTTP format and executes network requests.
+@_documentation(visibility: private)
+public class HTTPNetworkingClient: NetworkingClient {
+        
+    // MARK: - Internal Properties
```
MikeThorntonPayPal: nitpick: comment says "Internal Properties" but properties are `private`

Sources/CorePayments/Networking/HTTPNetworkingClient.swift
```diff
@@ -0,0 +1,120 @@
+import Foundation
+
+/// This method is exposed for internal PayPal use only. Do not use. It is not covered by Semantic Versioning and may change or be removed at any time.
```
MikeThorntonPayPal: nitpick: comment says "This method" but it's a class

MikeThorntonPayPal (commented):
I think there may be some conflicts with feature/app-switch. That branch added ContentType support on RESTRequest for the OAuth token endpoint and an additionalHeaders param on the GraphQL fetch for passing a Bearer token. Neither appears to be currently accounted for in HTTPNetworkingClient or the NetworkingClient protocol here.

Sources/CorePayments/Networking/HTTPNetworkingClient.swift
sshropshire: Oh yeah those will need to be cleaned up too. That's a good callout.

MikeThorntonPayPal (approved):

PR #381, opened 2026-04-16
v3: Migrate analytics events to type-safe enums

MikeThorntonPayPal:
## What
Replaces the string-based FPTI analytics event pattern with type-safe enums, per the iOS section of the [PPCP v3 Proposed Breaking Changes](https://paypal.atlassian.net/wiki/spaces/CSSE/pages/2900525846/PPCP+v3+Proposed+Breaking+Changes#iOS) doc.

Additionally cleans up the analytics API surface:
- `analyticsService` → `analytics`
- `sendEvent` → `track`
- `analytics` is now a non-optional `let` on client classes, removing optional-chaining noise from every call site.

## Why
Raw strings are verbose at call sites, easy to typo, and hard to grep for when dashboards need to change. Enums give us autocomplete, compile-time safety, and one single place to see every event a module can emit. Following the pattern suggested in Confluence: `cardAnalytics.notify(.checkoutSuccess)` → here, `analyticsService.sendEvent(CardAnalyticsEvent.approveOrderSucceeded)`.

## How
- New `AnalyticsEventName` protocol in `CorePayments`.
- Per-module enums: `CardAnalyticsEvent`, `PayPalWebAnalyticsEvent`, `PaymentButtonAnalyticsEvent`.
- `AnalyticsService` gains a generic `track(_:correlationID:buttonType:)` overload; string overload demoted to `internal`.
- All 32 call sites migrated.
- Roundtrip unit tests assert the exact FPTI string per case, so nobody accidentally renames a live event.

### Why `AnalyticsService` changed from `struct` to `class`
The goal was to make `analytics` a non-optional `let` on client classes. The problem is that `orderID` and `setupToken` need to change per method call — `vault()` sets a `setupToken`, `approveOrder()` sets an `orderID`.

With a **struct** (value type), a `let` property is fully immutable — you can't mutate any of its properties. So the old code had to use `var` and reassign a brand-new `AnalyticsService` instance each time, which forced it to be both `var` (reassignable) and optional (no value at init time), meaning every call site needed `analyticsService?.sendEvent(...)`.

With a **class** (reference type), a `let` property holds an immutable *reference* to the object, but the object's own `var` properties can still be mutated. This lets clients hold a single `let analytics` reference initialized at `init` time while still updating `orderID`/`setupToken` per-flow — no optional chaining needed.

## Target branch
`v3-beta` (per the unified v3 CI thread)

## Verification
- FPTI event names unchanged — asserted by `*AnalyticsEvent_Tests`.
- `AnalyticsService_Tests` extended with a generic-overload test.
- Local `xcodebuild … test` green.

### DIFF
--- a/Sources/CardPayments/CardClient.swift
+++ b/Sources/CardPayments/CardClient.swift
@@ -53,7 +53,7 @@ public class CardClient: NSObject {
     ///                 - `.failure(CoreSDKError)`: Describes the reason for failure.
     public func vault(_ vaultRequest: CardVaultRequest, completion: @escaping (Result<CardVaultResult, CoreSDKError>) -> Void) {
         analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
-        analyticsService?.sendEvent(CardAnalyticsEvent.vaultStarted)
+        analyticsService?.sendEvent(CardAnalyticsEvent.Vault.started)
         Task {
             do {
                 let result = try await vaultAPI.updateSetupToken(cardVaultRequest: vaultRequest).updateVaultSetupToken
@@ -64,7 +64,7 @@ public class CardClient: NSObject {
                         self.notify3dsVaultFailure(with: CardError.threeDSecureURLError, completion: completion)
                         return
                     }
-                    analyticsService?.sendEvent(CardAnalyticsEvent.vaultAuthChallengeRequired)
+                    analyticsService?.sendEvent(CardAnalyticsEvent.Vault.authChallengeRequired)
                     startVaultThreeDSecureChallenge(url: url, setupTokenID: vaultRequest.setupTokenID, completion: completion)
                 } else {
                     let vaultResult = CardVaultResult(setupTokenID: result.id, status: result.status, didAttemptThreeDSecureAuthentication: false)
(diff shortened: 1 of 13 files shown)


Sources/CardPayments/CardClient.swift
```diff
@@ -53,7 +53,7 @@ public class CardClient: NSObject {
     ///                 - `.failure(CoreSDKError)`: Describes the reason for failure.
     public func vault(_ vaultRequest: CardVaultRequest, completion: @escaping (Result<CardVaultResult, CoreSDKError>) -> Void) {
         analyticsService = AnalyticsService(coreConfig: config, setupToken: vaultRequest.setupTokenID)
-        analyticsService?.sendEvent("card-payments:vault-wo-purchase:started")
+        analyticsService?.sendEvent(CardAnalyticsEvent.Vault.started)
```
sshropshire: This is a solid start. We should make `AnalyticsService` a non-null `let` property. I'm looking at the code it doesn't seem like there's a good reason for it be a `var`. Renaming the property to just `analytics` and `sendEvent` to just `track` would be dope too now that we have typed "Event" enums.

So we'd end up with a clean:
```swift
analytics.track(CardAnalyticsEvent.Vault.started)
```

The big picture goal here is readability. Right now it feels like the analytics code is a bit noisy and distracts from the true intent of the surrounding code. If we clean this up it'll be easier for us (and possibly Claude) to follow.

sshropshire (approved):

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -35,31 +45,40 @@ public struct AnalyticsService {
...
+
+    /// Sends an analytics event by raw FPTI event name.
+    ///
+    /// Prefer the `AnalyticsEventName` overload. This string-based entry point is
+    /// retained for cross-module internal use and unit tests that need to inspect
+    /// the emitted event name directly.
+    func track(_ name: String, correlationID: String? = nil, buttonType: String? = nil) {
         Task(priority: .background) {
-            await performEventRequest(name, correlationID: correlationID, buttonType: buttonType)
+            await performTrackRequest(name, correlationID: correlationID, buttonType: buttonType)
```
sshropshire:
```suggestion
            await sendEvent(name, correlationID: correlationID, buttonType: buttonType)
```
This is a super nitpick/take-or-leave, but `sendEvent()` would read nicely here internally. I'm not sure if it can be made private as well to avoid confusion with the "sendEvent" method we just removed.

PR #384, opened 2026-04-27
[V3 Beta] Streamline Card Payments UI

sshropshire:
### Summary of changes

- This PR streamlines the Card Payments UI into a single scroll view.
- This PR also makes use of `environmentObject` and `ObservableObject` to clean up UI state management logic.

## After Refactor

https://github.com/user-attachments/assets/c3874bdd-a263-4d0c-aded-62e21c4d6f66

## Current State on `main`

https://github.com/user-attachments/assets/2793736a-3232-458a-ab43-79bd05ae2dc9

### Checklist

- @sshropshire


Demo/Demo/SwiftUIComponents/CommonComponents/StepHeader.swift
```diff
@@ -0,0 +1,13 @@
...
+
+struct StepHeader: View {
+    
+    let text: String
+    
+    var body: some View {
+        Text(text)
+            .font(.system(size: 20))
+            .frame(maxWidth: .infinity, alignment: .leading)
+            .font(.headline)
```
MikeThorntonPayPal: Two .font() modifiers are applied to the same Text: .font(.system(size: 20)) and .font(.headline). The first one wins in SwiftUI, so .font(.headline) on line 11 is silently ignored. Remove one or the other.

Demo/Demo/CardPayments/CardPaymentView.swift
```diff
@@ -0,0 +1,129 @@
...
+                    if let cardResult = viewModel.approveOrderResult.value {
+                        CardResultView(cardResult: cardResult)
+                        CompleteOrder(request: createOrderRequest)
+                    }
+                    if let captureResult = viewModel.captureAuthorizeResult.value {
+                        OrderView(order: captureResult)
+                    }
+                    ScrollAnchor(id: "bottomAnchor")
+                }
+                .onChange(of: viewModel.stepCount) { _ in
```
MikeThorntonPayPal: `onChange(of:)` with trailing closure `{ _ in } ` is the deprecated iOS 14-style API. If the deployment target supports iOS 17+, use the newer `onChange(of:) { oldValue, newValue in }`

sshropshire: Should we upgrade the demo app to iOS 17? We probably should bump the min supported version for v3 of our SDK to give us more flexibility. I could improve this PR a good bit if we bump the version too.

Demo/Demo/SwiftUIComponents/CommonComponents/StepHeader.swift
sshropshire: Fixed in [c374693](https://github.com/paypal/paypal-ios/pull/384/commits/c37469374a9eba7e4537dab8ccad539f23979020).

PR #386, opened 2026-05-11
v3: tighten CoreSDKError nullability (code, domain)

MikeThorntonPayPal:
## Summary
- Flip `CoreSDKError.code` from `Int?` to `Int` and `CoreSDKError.domain` from `String?` to `String`
- All 29 instantiation sites (source, tests, demo) already pass non-nil values — no other files modified
- Add `v3_MIGRATION_GUIDE.md` documenting the two nullability changes

## Heads-up
`CoreSDKError.code` and `CoreSDKError.domain` are accessed by merchants in error-handling code and appear in public developer.paypal.com docs. Any merchant using `error.code` as `Int?` will get a compile error (e.g., if they nil-check it).

## What stays optional
- `CoreSDKError.errorDescription` — required by `LocalizedError` protocol conformance
- `CardResult.status` / `CardVaultResult.status` — explicitly set to `nil` after 3DS flow
- `Card.cardholderName` / `billingAddress` — merchant input; not all transactions require these
- `Address` fields — address components vary by country; PayPal API marks them optional
- `PaymentSource.Card.*` — `@_documentation(visibility: private)` Decodable response model
- `PaymentButton.insets` / `.label` — nil carries semantic meaning (dynamic resolution / no label)

## Test plan
- [x] `xcodebuild build` — CorePayments scheme compiles
- [x] `xcodebuild test` — UnitTests scheme passes (all suites, 0 failures)


Sources/CorePayments/CoreSDKError.swift
```diff
@@ -4,15 +4,15 @@ import Foundation
...
 
     /// A string containing the error domain.
-    public let domain: String?
+    public let domain: String
 
     /// A string containing the localized description of the error.
     public let errorDescription: String?
 
-    public init(code: Int?, domain: String?, errorDescription: String?) {
+    public init(code: Int, domain: String, errorDescription: String?) {
```
sshropshire, line 16: Honestly this shouldn't ever be constructed by merchants. We may be able to skip mentioning changes to this in the migration guide. We also should have a `@_documentation(visibility: private)` to prevent docs from being generated for it.

sshropshire (approved):

Sources/CorePayments/CoreSDKError.swift
MikeThorntonPayPal: great callout. I'll do that

PR #387, opened 2026-05-12
v3: Extract TrackingEventsAPIProtocol (POP)

MikeThorntonPayPal:
## Summary
- Create `TrackingEventsAPIProtocol` with the single `sendEvent(with:)` method
- Conform `TrackingEventsAPI` to the new protocol
- Change `AnalyticsService` to depend on `TrackingEventsAPIProtocol` instead of the concrete `TrackingEventsAPI` class
- Simplify `MockTrackingEventsAPI` to conform to the protocol instead of subclassing `TrackingEventsAPI`

Follows the same pattern established in #380 for `HTTPClient` and `NetworkingClient`.

## Test plan
- [x] CI unit tests pass
- [x] Verify `AnalyticsService_Tests` and `TrackingEventsAPI_Tests` pass locally

### DIFF
--- a/Sources/CorePayments/Networking/TrackingEventsAPIProtocol.swift
+++ b/Sources/CorePayments/Networking/TrackingEventsAPIProtocol.swift
@@ -1,0 +1,7 @@
+import Foundation
+
+/// Protocol defining the interface for sending FPTI analytics events.
+protocol TrackingEventsAPIProtocol {
+
+    func sendEvent(with analyticsEventData: AnalyticsEventData) async throws -> HTTPResponse
+}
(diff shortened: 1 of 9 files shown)


Sources/CorePayments/Networking/TrackingEventsAPIProtocol.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+/// Protocol defining the interface for sending FPTI analytics events.
+protocol TrackingEventsAPIProtocol {
```
sshropshire: Hey @MikeThorntonPayPal what do you think about this rule in the [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/)?

Image

Is there an opportunity for us to name this protocol with an `ing` suffix? e.g. `AnalyticsReporting`.

sshropshire (approved):

Sources/CorePayments/Networking/TrackingEventsAPIProtocol.swift
MikeThorntonPayPal: Good call, updated to AnalyticsEventTracking. Also renamed the mock to MockAnalyticsEventTrackingAPI to match.

PR #401, opened 2026-06-03
[Patch] App Switch — URL Opening Reliability

NastassiaRodzik:
### Reason for changes

AppSwitch must open the PayPal app via Universal Link, not fall back to Safari. This requires passing `universalLinksOnly: true` to `UIApplication.open(_:options:completionHandler:)`.
Not doing so will result in sending false positive `paypal-web-payments:checkout:app-switch-open:succeeded` analytics event, instead of expected `paypal-web-payments:checkout:app-switch-open:failed`. 

### Summary of changes

- Add `options` parameter to URLOpener `open(_ url)` method.
- Replace the `UIApplication: URLOpener` extension with a `DefaultURLOpener` wrapper that forwards to UIApplication without method shadowing.

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik

### Screenshots / Video

**Happy Path:** 

Video:
https://github.com/user-attachments/assets/2589ab1c-dcf4-43cb-b65e-d99299a0fa28

Analytics: 
```
paypal-web-payments:checkout:started 
paypal-web-payments:checkout:app-switch-open:succeeded 
paypal-web-payments:checkout:succeeded 
```

**Universal link cannot be opened scenario:** 

Video: 

https://github.com/user-attachments/assets/b7c10173-aada-407b-9a4a-e39f67cbdffc

Analytics **before** the fix: 

```
paypal-web-payments:checkout:started
paypal-web-payments:checkout:app-switch-open:succeeded  // despite App Switch failure
paypal-web-payments:checkout:succeeded
```

Analytics **after** the fix: 

```
paypal-web-payments:checkout:started 
paypal-web-payments:checkout:app-switch-open:failed 
paypal-web-payments:checkout:fallback-to-web:cannot_open_url  // fallback to web event is sent, as expected 
paypal-web-payments:checkout:succeeded 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+
+* CorePayments
+  * Add `options` parameter to `URLOpener openURL()` method
+
 ## 2.0.1 (2025-11-03)
 
 * PayPalWebPayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,6 @@
 
 ## unreleased
 
-* CorePayments
-  * Add `options` parameter to `URLOpener openURL()` method
 
 ## 2.0.1 (2025-11-03)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,6 @@
 
 # PayPal iOS SDK Release Notes
 
-## unreleased
-
-
 ## 2.0.1 (2025-11-03)
 
 * PayPalWebPayments
--- a/Sources/CorePayments/UIApplication+URLOpener.swift
+++ b/Sources/CorePayments/UIApplication+URLOpener.swift
@@ -2,20 +2,35 @@ import UIKit
 
 @_documentation(visibility: private)
 public protocol URLOpener {
-    func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?)
+    func open(
+        _ url: URL,
+        options: [UIApplication.OpenExternalURLOptionsKey: Any],
+        completionHandler completion: ((Bool) -> Void)?
+    )
     func isPayPalAppInstalled() -> Bool
 }
 
-extension UIApplication: URLOpener {
+@_documentation(visibility: private)
+public struct DefaultURLOpener: URLOpener {
+
+    private let application: UIApplication
+
+    public init(application: UIApplication = .shared) {
+        self.application = application
+    }
+
+    public func open(
+        _ url: URL,
+        options: [UIApplication.OpenExternalURLOptionsKey: Any],
+        completionHandler completion: ((Bool) -> Void)?
+    ) {
+        application.open(url, options: options, completionHandler: completion)
+    }
 
     public func isPayPalAppInstalled() -> Bool {
         guard let payPalURL = URL(string: "paypal://") else {
             return false
         }
-        return canOpenURL(payPalURL)
-    }
-
-    public func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?) {
-        UIApplication.shared.open(url, options: [:], completionHandler: completion)
+        return application.canOpenURL(payPalURL)
     }
 }
--- a/Sources/CorePayments/UIApplication+URLOpener.swift
+++ b/Sources/CorePayments/UIApplication+URLOpener.swift
@@ -10,6 +10,12 @@ public protocol URLOpener {
     func isPayPalAppInstalled() -> Bool
 }
 
+public extension URLOpener {
+    func open(_ url: URL, completionHandler: ((Bool) -> Void)? = nil) {
+        open(url, universalLinksOnly: true, completionHandler: completionHandler)
+    }
+}
+
 extension UIApplication: URLOpener {
 
     public func isPayPalAppInstalled() -> Bool {
@@ -21,7 +27,7 @@ extension UIApplication: URLOpener {
 
     public func open(
         _ url: URL,
-        universalLinksOnly: Bool,
+        universalLinksOnly: Bool = true,
         completionHandler completion: ((Bool) -> Void)?
     ) {
         let options: [UIApplication.OpenExternalURLOptionsKey: Any] = universalLinksOnly ? [.universalLinksOnly: true] : [:]
--- a/Sources/CorePayments/UIApplication+URLOpener.swift
+++ b/Sources/CorePayments/UIApplication+URLOpener.swift
@@ -11,7 +11,7 @@ public protocol URLOpener {
 }
 
 @_documentation(visibility: private)
-public struct DefaultURLOpener: URLOpener {
+internal struct DefaultURLOpener: URLOpener {
 
     private let application: UIApplication
 
--- a/Sources/CorePayments/UIApplication+URLOpener.swift
+++ b/Sources/CorePayments/UIApplication+URLOpener.swift
@@ -4,33 +4,27 @@ import UIKit
 public protocol URLOpener {
     func open(
         _ url: URL,
-        options: [UIApplication.OpenExternalURLOptionsKey: Any],
+        universalLinksOnly: Bool,
         completionHandler completion: ((Bool) -> Void)?
     )
     func isPayPalAppInstalled() -> Bool
 }
 
-@_documentation(visibility: private)
-internal struct DefaultURLOpener: URLOpener {
-
-    private let application: UIApplication
-
-    public init(application: UIApplication = .shared) {
-        self.application = application
-    }
-
-    public func open(
-        _ url: URL,
-        options: [UIApplication.OpenExternalURLOptionsKey: Any],
-        completionHandler completion: ((Bool) -> Void)?
-    ) {
-        application.open(url, options: options, completionHandler: completion)
-    }
+extension UIApplication: URLOpener {
 
     public func isPayPalAppInstalled() -> Bool {
         guard let payPalURL = URL(string: "paypal://") else {
             return false
         }
-        return application.canOpenURL(payPalURL)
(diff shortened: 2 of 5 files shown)


Sources/CorePayments/UIApplication+URLOpener.swift
```diff
@@ -2,20 +2,35 @@ import UIKit
 
 @_documentation(visibility: private)
 public protocol URLOpener {
-    func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?)
+    func open(
+        _ url: URL,
+        options: [UIApplication.OpenExternalURLOptionsKey: Any],
```
MikeThorntonPayPal: Exposing this type-erased dictionary in the protocol leaks implementation details in this abstraction.  Which means the "app switch only" semantic is sort of silently lost in the abstraction and instead lives in the caller.

Might be better to use a bool in the signature, such that it's harder to misuse`universalLinksOnly: Bool = false`. This

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -14,7 +14,7 @@ public class PayPalWebCheckoutClient: NSObject {
 
     var appSwitchCompletion: ((Result<PayPalWebCheckoutResult, CoreSDKError>) -> Void)?
     var vaultAppSwitchCompletion: ((Result<PayPalVaultResult, CoreSDKError>) -> Void)?
-    var application: URLOpener = UIApplication.shared
+    var application: URLOpener = DefaultURLOpener()
```
MikeThorntonPayPal: nitpick: worth renaming this property as it no longer is the `UIApplication` type. `urlOpener` or something.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -395,9 +396,9 @@ public class PayPalWebCheckoutClient: NSObject {
     }
 
     @MainActor
-    private func openURL(_ url: URL) async -> Bool {
+    private func openURLAppSwitchOnly(_ url: URL) async -> Bool {
         await withCheckedContinuation { continuation in
-            application.open(url) { success in
+            application.open(url, options: [.universalLinksOnly: true]) { success in
```
MikeThorntonPayPal: Along with the suggestion above: this would be 

```
urlOpener.open(url, universalLinksOnly: true) { success in ... }
```

UnitTests/TestShared/MockURLOpener.swift
```diff
@@ -1,4 +1,4 @@
-import Foundation
+import UIKit
```
MikeThorntonPayPal: would not need uikit import if made the param Bool

Sources/CorePayments/UIApplication+URLOpener.swift
```diff
@@ -2,20 +2,35 @@ import UIKit
...
+        completionHandler completion: ((Bool) -> Void)?
+    )
     func isPayPalAppInstalled() -> Bool
 }
 
-extension UIApplication: URLOpener {
+@_documentation(visibility: private)
+public struct DefaultURLOpener: URLOpener {
```
MikeThorntonPayPal: I don't think we should expose this to SDK consumer.  internal would likely be a better choice for this

MikeThorntonPayPal (commented):
It would be good to add some tests for the behavioral change here, since MockURLOpener will be taking a new parameter. Would be prudent to ensure the correct option is actually passed.

Sources/CorePayments/UIApplication+URLOpener.swift
MikeThorntonPayPal: Addressed.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
MikeThorntonPayPal: Addressed.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
MikeThorntonPayPal: Addressed.

UnitTests/TestShared/MockURLOpener.swift
MikeThorntonPayPal: Addressed.

Sources/CorePayments/UIApplication+URLOpener.swift
MikeThorntonPayPal: Addressed.

Sources/CorePayments/UIApplication+URLOpener.swift
MikeThorntonPayPal: One thing still missing: `universalLinksOnly` should have a default value of `false` as originally suggested, so existing callers aren't forced to provide it. This applies to both the protocol definition and the `UIApplication` conformance.

Sources/CorePayments/UIApplication+URLOpener.swift
```diff
@@ -2,7 +2,11 @@ import UIKit
 
 @_documentation(visibility: private)
 public protocol URLOpener {
-    func open(_ url: URL, completionHandler completion: ((Bool) -> Void)?)
+    func open(
+        _ url: URL,
```
MikeThorntonPayPal, line 6: `universalLinksOnly` should have a default value of `true` as originally suggested, so existing callers are not forced to provide it. E.g. `universalLinksOnly: Bool = true`

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 
 # PayPal iOS SDK Release Notes
 
+## unreleased
+
+* CorePayments
+  * Add `options` parameter to `URLOpener openURL()` method
```
sshropshire:
```suggestion
```
No need to add private method changes to `CHANGELOG`.

Sources/CorePayments/UIApplication+URLOpener.swift
NastassiaRodzik: Good call out, fixed here [Ensure consistency for existing merchants using open(url: ) method](https://github.com/paypal/paypal-ios/pull/401/commits/1acdaa00575faad08f1151c8218163c47c379381)

MikeThorntonPayPal (approved):

Sources/CorePayments/UIApplication+URLOpener.swift
```diff
@@ -2,10 +2,20 @@ import UIKit
...
+        universalLinksOnly: Bool,
+        completionHandler completion: ((Bool) -> Void)?
+    )
```
sshropshire, line 9: This method signature differs from the equivalent [UIApplication](https://developer.apple.com/documentation/uikit/uiapplication/open(_:options:completionhandler:)) method. Is this intentional?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -14,7 +14,7 @@ public class PayPalWebCheckoutClient: NSObject {
 
     var appSwitchCompletion: ((Result<PayPalWebCheckoutResult, CoreSDKError>) -> Void)?
     var vaultAppSwitchCompletion: ((Result<PayPalVaultResult, CoreSDKError>) -> Void)?
-    var application: URLOpener = UIApplication.shared
+    var urlOpener: URLOpener = UIApplication.shared
```
sshropshire, line 17: 🔥 This is a better name thank you.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -65,7 +65,7 @@ public class PayPalWebCheckoutClient: NSObject {
         analyticsService?.sendEvent("paypal-web-payments:checkout:started")
 
         let completionOnce = makeCompletionOnce(completion)
-        let appInstalled = self.application.isPayPalAppInstalled()
+        let appInstalled = self.urlOpener.isPayPalAppInstalled()
```
sshropshire: This seems like pre-existing debt, but do we need `self` here?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -190,13 +190,14 @@ public class PayPalWebCheckoutClient: NSObject {
             }
 
             // Try to open the PayPal app (or deep link). If opening fails, fall back.
-            let opened = await openURL(url)
+            let opened = await openURLAppSwitchOnly(url)
```
```suggestion
            let opened = await attemptAppSwitch(with: url)
```
I'd go a step further and leave the "how" up to the implementation of the method. This avoids a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/). The naming I'm suggesting doesn't have to be exact, I'm providing an example to highlight a more expressive name.

Sources/CorePayments/UIApplication+URLOpener.swift
NastassiaRodzik: yes, this was suggested by Mike, to avoid leaking implementation details in URLOpener:
https://github.com/paypal/paypal-ios/pull/401#discussion_r3364100524

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire, line 204: This is a legit question I have, I'm not as familiar with Swift Concurrency: Can't we mark the the `appSwitchCompletion` property as `@MainActor` and avoid the closure?

sshropshire (approved):

Sources/CorePayments/UIApplication+URLOpener.swift
sshropshire: Oh that's interesting.

PR #402, opened 2026-06-03
[Patch] Analytics — app-switch-open:succeeded  can be lost on app suspension 

NastassiaRodzik:
### Reason for changes

The problem: `AnalyticsService.sendEvent()` wraps its HTTP request in a Swift `Task(priority: .background)`.
`Task(priority: .background)` sets Swift concurrency thread priority — it is **not** an iOS background task. When the app is suspended by the OS (which happens immediately after UIApplication.open() successfully launches the PayPal app), all Swift tasks are frozen and the network request is never sent.

Analytics events at risk: `"paypal-web-payments:checkout:app-switch-open:succeeded"`

### Summary of changes

- Add `sendEventWithBackgroundProtection` method that protects Analytics events with an iOS background task.

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik

### Evidence

When using `sendEventWithBackgroundProtection`:

App Switch flow:
https://github.com/user-attachments/assets/0f6d82b6-933c-4cd3-8fe7-d98d84564256

Analytics: 
```
paypal-web-payments:checkout:started - sent ✅ 
paypal-web-payments:checkout:app-switch-open:succeeded - sent with BG Protection ✅ 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,13 @@
 
 # PayPal iOS SDK Release Notes
 
+## UNRELEASED
+
+* CorePayments
+  * Add `sendEventWithBackgroundProtection(_:correlationID:buttonType:)` to `AnalyticsService` to request additional runtime via `beginBackgroundTask` when sending analytics during app backgrounding
+* PayPalWebPayments
+  * Use background-protected analytics delivery for the App Switch open success event to improve delivery when the merchant app transitions to the background
+
 ## 2.0.1 (2025-11-03)
 
 * PayPalWebPayments
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,6 @@
 
 # PayPal iOS SDK Release Notes
 
-## UNRELEASED
-
-* CorePayments
-  * Add `sendEventWithBackgroundProtection(_:correlationID:buttonType:)` to `AnalyticsService` to request additional runtime via `beginBackgroundTask` when sending analytics during app backgrounding
-* PayPalWebPayments
-  * Use background-protected analytics delivery for the App Switch open success event to improve delivery when the merchant app transitions to the background
-
 ## 2.0.1 (2025-11-03)
 
 * PayPalWebPayments
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -120,6 +120,7 @@ public struct AnalyticsService {
             var bgTaskID: UIBackgroundTaskIdentifier = .invalid
             var eventTask: Task<Void, Never>?
 
+            @MainActor
             func endBackgroundTaskIfNeeded() {
                 guard bgTaskID != .invalid else { return }
                 UIApplication.shared.endBackgroundTask(bgTaskID)
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -43,15 +43,25 @@ public struct AnalyticsService {
     /// - Parameter name: Event name string used to identify this unique event in FPTI.
     /// - Parameter correlationID: correlation ID associated with the request
     /// - Parameter buttonType: The type of button
-    public func sendEvent(_ name: String, correlationID: String? = nil, buttonType: String? = nil) {
+    /// - Parameter withBackgroundProtection: When `true`, requests additional runtime if the app enters
+    ///   the background before the network request completes. Delivery is best-effort and not guaranteed.
+    public func sendEvent(
+        _ name: String,
+        correlationID: String? = nil,
+        buttonType: String? = nil,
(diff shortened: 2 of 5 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -194,7 +194,7 @@ public class PayPalWebCheckoutClient: NSObject {
 
             if opened {
                 // TODO: align with android on app switch event names, communicate with analytics team on new events
-                analyticsService?.sendEvent("paypal-web-payments:checkout:app-switch-open:succeeded")
+                analyticsService?.sendEventWithBackgroundProtection("paypal-web-payments:checkout:app-switch-open:succeeded")
```
sshropshire: q: Do we need a name change for this method? It feels like the `WithBackgroundProtection` suffix is leaking an implementation detail (ref: [Article on Information Hiding](https://thehosk.medium.com/good-coding-practices-information-hiding-e1449d9339d)).

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 
 # PayPal iOS SDK Release Notes
 
+## UNRELEASED
+
+* CorePayments
+  * Add `sendEventWithBackgroundProtection(_:correlationID:buttonType:)` to `AnalyticsService` to request additional runtime via `beginBackgroundTask` when sending analytics during app backgrounding
+* PayPalWebPayments
+  * Use background-protected analytics delivery for the App Switch open success event to improve delivery when the merchant app transitions to the background
+
```
sshropshire:
```suggestion
```
I'd revert this `CHANGELOG` addition since the changes aren't meant for public consumption.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: updated

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -43,20 +43,63 @@ public struct AnalyticsService {
...
             await performEventRequest(name, correlationID: correlationID, buttonType: buttonType)
         }
     }
 
+    private func sendEventWithBackgroundProtection(
+        _ name: String,
+        correlationID: String? = nil,
+        buttonType: String? = nil
+    ) {
+        Task { @MainActor in
```
sshropshire, line 119: random q: Is there a reason that this `Task` needs to be run on the main thread? I can't tell at a glance if there are APIs that we use here that require the caller to be on the main thread.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -43,20 +43,63 @@ public struct AnalyticsService {
...
+            var bgTaskID: UIBackgroundTaskIdentifier = .invalid
+            var eventTask: Task<Void, Never>?
```
sshropshire, line 121:
```suggestion
            let bgTaskID: UIBackgroundTaskIdentifier
            let eventTask: Task<Void, Never>
```
I'm just curious here: does this code snippet ^ break compilation?

NastassiaRodzik: yes, it's because we need to capture them both by a closure on line 79 and 84

Sources/CorePayments/Networking/AnalyticsService.swift
NastassiaRodzik: @MainActor is here because sendEventWithBackgroundProtection talks to UIKit (UIApplication.shared.beginBackgroundTask / endBackgroundTask), and all UIKit-related methods need to run on Main thread in iOS

sshropshire (approved):

UnitTests/PaymentsCoreTests/Mocks/MockTrackingEventsAPI.swift
```diff
@@ -5,11 +5,18 @@ class MockTrackingEventsAPI: TrackingEventsAPI {
...
     var stubError: Error?
+    var sendEventDelay: UInt64 = 0
+    var onSendEvent: (() -> Void)?
     
     var capturedAnalyticsEventData: AnalyticsEventData?
     
     override func sendEvent(with analyticsEventData: AnalyticsEventData) async throws -> HTTPResponse {
+        if sendEventDelay > 0 {
+            try await Task.sleep(nanoseconds: sendEventDelay)
+        }
```
sshropshire, line 16: q: What are the benefits of delaying an analytics event?

NastassiaRodzik: This is for the unit test `testPerformEventRequest_whenTaskCancelled_doesNotCallAPI() ` that ensures proper task cancellation: when task is cancelled we do not send analytics data

sshropshire (approved):

PR #403, opened 2026-06-04
[Patch] Centralize app-switch constants

NastassiaRodzik:
### Reason for changes

This PR cleans up hardcoded values introduced during app-switch work so production behavior stays correct and SDK/Demo code stays aligned with a single source of truth.

### Summary of changes

- Replace raw hardcoded strings with constants 

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik


Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModelConstants.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  PayPalWebViewModelConstants.swift
```
MikeThorntonPayPal: First, I think naming these as defaults are a bit misleading, as they're specifically fixtures for demonstration.  But second, since these extend SDK model types (Amount, Vault), with demo-specific values, these extensions add discoverable API surface to SDK types that only pertain to the demo context.

Personally I think just creating static variables in the PayPalWebViewModel would be acceptable for these. But if conceptually you prefer wrapping them in an enum, we should probably just nest it within PayPalWebViewModel

```
PayPalWebViewModel {
    private enum DemoFixtures {
        static let amount = Amount(currencyCode: "USD", value: "10.00")
        static let vault = Vault(storeInVault: "ON_SUCCESS", usageType: "MERCHANT", customerType: "CONSUMER")
    }
}
```

Sources/CorePayments/PayPalCoreConstants.swift
```diff
@@ -9,4 +9,10 @@ public enum PayPalCoreConstants {
     public static let payPalSDKVersion: String = "2.0.1"
     
     public static let callbackURLScheme: String = "sdk.ios.paypal"
+    
+    public static let integrationArtifact: String = "MOBILE_SDK"
+    
+    public static let osType = "IOS"
+    
+    static let integrationChannel = "PPCP_NATIVE_SDK"
```
MikeThorntonPayPal: nitpick: I'm fine with either explicitly providing the type or not, but just pick either providing it or letting it be inferred for all three here. It seems the current pattern is to explicitly provide it from the other constants, so I'd probably just fall back to explicitly providing a type for all of them.

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -1,5 +1,11 @@
 import Foundation
 
+@_documentation(visibility: private)
+public enum TokenType: String {
```
MikeThorntonPayPal: Conceptually this is more of a networking or API concern and not specific to the CCO patching class, it would probably make sense to live in its own file or within sort of a PayPalCoreConstants file. Also there's a bit of a contradiction in signal from this being public, but having private documentation.

From a Swift standpoint, unless there's a use case for using the enum constructor with a provided string ie: TokenType(rawValue: "some_string") its usually easier to simple provide static string properties in the use case below
static let orderId: String = "ORDER_ID"
static let clientToken: String = "CLIENT_TOKEN"

which cleans up the call sites from TokenType.clientToken.rawValue to just TokenType.clientToken
But in that case, I'd probable rename this from TokenType to TokenKey (or something like that)

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -70,20 +76,21 @@ public class PatchCCOWithAppSwitchEligibility {
 
     public func patchCCOWithAppSwitchEligibility(
         token: String,
-        tokenType: String
+        tokenType: String,
+        paypalNativeAppInstalled: Bool
```
MikeThorntonPayPal: Any external consumer calling this method will break at compile time. This is a semver-violating change. If we're going to add it, I'd strongly suggest we provide this new parameter with a default value such that it isn't a breaking change. I'd also make the same suggestion here with internal/external variable names that I make in the attemptAppSwitchIfEligible.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -171,12 +176,14 @@ public class PayPalWebCheckoutClient: NSObject {
 
     private func attemptAppSwitchIfEligible(
         request: PayPalWebCheckoutRequest,
+        paypalNativeAppInstalled: Bool,
```
MikeThorntonPayPal: nitpick: I think we could make "swift-ier" and more readable at call site with something like 
```
canSwitchToApp nativeAppInstalled: Bool
```
changing the call site to read more like:
```
patchCCOAPI.patchCCOWithAppSwitchEligibility(
    token: request.orderID, 
    tokenType: .orderId,
    isAppSwitchEligible: nativeAppInstalled
)
```

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -4,6 +4,11 @@ import AuthenticationServices
 import CorePayments
 #endif
 
+enum PayPalCheckoutCallbackURL {
```
MikeThorntonPayPal: If PayPalWebCheckoutClient is the only place these strings are used, we could just nest this enum inside it and make them private.

MikeThorntonPayPal (commented):
"Centralize constants" and "pass real app install state" are two independent concerns. 1 is a refactor, the other is a behavior change (with a public API signature update). Separate PRs would keep the git history cleaner if we ever need to bisect.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModelConstants.swift
MikeThorntonPayPal: This still extends SDK model types (`Amount`, `Vault`) with demo-specific values. Please move these into `PayPalWebViewModel` as private static properties or a nested `DemoFixtures` enum as suggested.

Sources/CorePayments/PayPalCoreConstants.swift
MikeThorntonPayPal: Still inconsistent — `integrationArtifact` has an explicit `: String` annotation, but `osType` and `integrationChannel` rely on inference. Please pick one style for all three.

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
MikeThorntonPayPal: Still in `PatchCCOWithAppSwitchEligibility.swift`. Please move it to its own file or into `PayPalCoreConstants`. Also, since every call site uses `.rawValue`, consider switching to static `String` properties (e.g. `static let orderId = "ORDER_ID"`) to simplify call sites.

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
MikeThorntonPayPal: This is still a breaking change — `paypalNativeAppInstalled` has no default value, so existing callers will fail to compile. Please add a default, e.g. `paypalNativeAppInstalled: Bool = true`.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
MikeThorntonPayPal: Still named `paypalNativeAppInstalled`. The suggestion was to use a more Swifty external/internal label split, e.g. `canSwitchToApp nativeAppInstalled: Bool`, so the call site reads as `canSwitchToApp: true`.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
MikeThorntonPayPal: This is still a file-level enum. Since it's only used within `PayPalWebCheckoutClient` (tests access via `@testable import`), please nest it inside `PayPalWebCheckoutClient` and mark it `private`.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModelConstants.swift
```diff
@@ -0,0 +1,14 @@
...
+//  Demo
+//
+//  Created by Anastasia Rodzik on 6/4/26.
```
sshropshire: Please remove this header.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModelConstants.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+
+extension Amount {
+    static let defaultAmount: Amount = .init(currencyCode: "USD", value: "10.00")
```
sshropshire: Any reason this can't be placed in `Amount.swift`?

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
 
+## Unreleased
+
+* CorePayments
+  * Update `PatchCCOWithAppSwitchEligibility.patchCCOWithAppSwitchEligibility` to accept a `paypalNativeAppInstalled` parameter so app switch eligibility reflects whether the PayPal app is actually installed instead of always sending `true`
+  * Add `TokenType` enum (`orderId`, `clientToken`) to replace hardcoded token-type string values
+  * Add `integrationArtifact`, `osType`, and `integrationChannel` constants to `PayPalCoreConstants`
+* PayPalWebPayments
+  * Replace hardcoded string literals with shared constants for the checkout redirect URL, integration artifact, and token type
+
```
sshropshire: Make sure to review these comments to see if they are actual changes to the public facing API.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModelConstants.swift
NastassiaRodzik: Hi @MikeThorntonPayPal There must be a mix up with PR #401 here or in Slack thread, I haven't pushed any fixes to your comments to this PR #403  yet, I'll let you know once I do. 
Moving it to Draft state for clarity, until it's ready for re review

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: Moved the enum inside PayPalWebCheckoutClient however have to keep it internal for tests to be able to access it: 

@NastassiaRodzik
[Move PayPalCheckoutCallbackURL inside PayPalWebCheckoutClient](https://github.com/paypal/paypal-ios/pull/403/commits/35ab9aa6414d19f1c1dcdc5f1231d5df7499b3e9)

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
NastassiaRodzik: As per your suggestion above, I'm moving `paypalNativeAppInstalled: Bool` change away from this PR and will make a separate Request with it.
Will apply this comment there

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: As per suggestion above, I'm moving `paypalNativeAppInstalled: Bool` change away from this PR and will make a separate Request with it.
Will apply this comment there

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -111,12 +111,12 @@ final class PatchCCOWithAppSwitchEligibility_Tests: XCTestCase {
             mockNetworkingClient.capturedGraphQLRequest?.variables as? PatchCcoWithAppSwitchEligibilityVariables
         )
         XCTAssertEqual(vars.contextId, token)
-        XCTAssertEqual(vars.experimentationContext?.integrationChannel, "PPCP_NATIVE_SDK")
-        XCTAssertEqual(vars.osType, "IOS")
+        XCTAssertEqual(vars.experimentationContext?.integrationChannel, PayPalCoreConstants.integrationChannel)
+        XCTAssertEqual(vars.osType, PayPalCoreConstants.osType)
         XCTAssertTrue(vars.merchantOptInForAppSwitch)
         XCTAssertEqual(vars.token, token)
```
MikeThorntonPayPal, line 117: Minor thought: it might be safer to keep test assertions against hardcoded literal values rather than the same constants used in production code. If a constant were accidentally changed to the wrong value, both the production code and the test would use the same wrong value and the test would still pass. Using literals like `"PPCP_NATIVE_SDK"` here acts as an independent check that the expected value is correct.

Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -1,3 +1,5 @@
+import CorePayments
```
MikeThorntonPayPal: Nit: this adds a CorePayments framework dependency to a Demo model file just to reference `PayPalCoreConstants.osType`, which is always `"IOS"`. Might be simpler to leave the hardcoded string here and avoid the extra import.

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
NastassiaRodzik: Done

Demo/Demo/Models/CreateOrderParams.swift
NastassiaRodzik: Done

MikeThorntonPayPal (approved):

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -5,14 +5,19 @@ import FraudProtection
 
 @MainActor
 class PayPalWebViewModel: ObservableObject {
+    
+    private enum DemoFixtures {
+        static let amount = Amount(currencyCode: "USD", value: "10.00")
+        static let vault = Vault(storeInVault: "ON_SUCCESS", usageType: "MERCHANT", customerType: "CONSUMER")
```
sshropshire: This looks good. Is the private `enum` for constants a recommended best practice?

sshropshire: Also pre-existing tech debt: I wish that struct was named something more expressive than `Vault` 😬 🤦.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -40,7 +45,7 @@ class PayPalWebViewModel: ObservableObject {
             )
 
             let attributes: Attributes? = shouldVault
-            ? Attributes(vault: Vault(storeInVault: "ON_SUCCESS", usageType: "MERCHANT", customerType: "CONSUMER"))
+            ? Attributes(vault: DemoFixtures.vault)
```
sshropshire: `VaultAtrributes` would be a better name for that struct, and `vaultAttributes` a better variable name.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
NastassiaRodzik: yes enums are great for constants since we cannot create an instance of one

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
sshropshire: Also this can be one line now. The multi-line formatting is no longer necessary.

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
```diff
@@ -49,7 +49,7 @@ public class UpdateClientConfigAPI {
         let variables = UpdateClientConfigVariables(
             token: token,
             fundingSource: fundingSource,
-            integrationArtifact: "MOBILE_SDK",
+            integrationArtifact: PayPalCoreConstants.integrationArtifact,
```
sshropshire, line 52: I understand the best practice effort here to remove magic strings, but its actually less readable now. If we wanted to eliminate magic and preserve readability, technically something like this would be more Swifty:
```suggestion
            integrationArtifact: .mobileSDK,
```
But that would require adding a new enum type for `IntegrationArtifact`. I don't know how y'all feel about that.

Sources/CorePayments/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -77,12 +77,12 @@ public class PatchCCOWithAppSwitchEligibility {
 
         let variables = PatchCcoWithAppSwitchEligibilityVariables(
             contextId: token,
-            experimentationContext: ExperimentationContext(integrationChannel: "PPCP_NATIVE_SDK"),
-            osType: "IOS",
+            experimentationContext: ExperimentationContext(integrationChannel: PayPalCoreConstants.integrationChannel),
```
sshropshire, line 81: +1 this is actually less expressive than an inline string in terms of readability.

Sources/CorePayments/PayPalCoreConstants.swift
```diff
@@ -9,4 +9,16 @@ public enum PayPalCoreConstants {
     public static let payPalSDKVersion: String = "2.0.1"
     
     public static let callbackURLScheme: String = "sdk.ios.paypal"
+    
+    public static let integrationArtifact: String = "MOBILE_SDK"
+    
+    public static let osType: String = "IOS"
```
sshropshire, line 15: I think I saw a callout in Slack or Confluence that we should align on these values across platforms. I do agree, my question: can we get someone from Analytics to make the call? Or is it just team preference.

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -128,7 +128,7 @@ final class PatchCCOWithAppSwitchEligibility_Tests: XCTestCase {
         mockNetworkingClient.stubHTTPError = CoreSDKError(code: 999, domain: "networking", errorDescription: "boom")
 
         do {
-            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "t", tokenType: "CLIENT_TOKEN")
+            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "t", tokenType: TokenType.clientToken)
```
sshropshire:
```suggestion
            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "fake-token", tokenType: TokenType.clientToken)
```
We've explicitly named mock values in the past. This is pre-existing tech debt, but it's worth considering to improve readability of the tests.

UnitTests/PayPalWebPaymentsTests/PayPalClients_CheckoutAppSwitch_Toggle_Tests.swift
```diff
@@ -46,7 +46,7 @@ class PayPalClient_CheckoutAppSwitch_Toggle_tests: XCTestCase {
         mockURLOpener.mockIsPayPalAppInstalled = true
         let request = PayPalWebCheckoutRequest(orderID: "test-order-id", appSwitchIfEligible: false)
         let expectedResult = PayPalWebCheckoutResult(orderID: "test-order-id", payerID: "test-payer-id")
-        mockWebAuthenticationSession.cannedResponseURL = URL(string: "sdk.ios.paypal://x-callback-url/paypal-sdk/paypal-checkout?token=test-order-id&PayerID=test-payer-id")
+        mockWebAuthenticationSession.cannedResponseURL = URL(string: "\(PayPalWebCheckoutClient.PayPalCheckoutCallbackURL.redirectURL)?token=test-order-id&PayerID=test-payer-id")
```
sshropshire: nitpick: this I would honestly revert. This [couples the test to the implementation](https://dev.to/scottshipp/don-t-use-non-test-constants-in-unit-tests-3ej0). And the code is actually longer than the string it replaces.

A file private constant would serve the same purpose:
```suggestion
        mockWebAuthenticationSession.cannedResponseURL = checkoutSuccessInboundDeepLink
```

Sources/CorePayments/Networking/UpdateClientConfigAPI.swift
NastassiaRodzik: I'd rather leave it to be tackled in separate PR/ticket to save more time for SSID implementation

Sources/CorePayments/PayPalCoreConstants.swift
NastassiaRodzik: I called it out that in LLD it's spelled as "iOS" however I doubt we should change this existing implementation ?

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
NastassiaRodzik: I'd leave it for a separate tech debt PR as well, to have more time on SSID implementation

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
NastassiaRodzik: I'd keep DemoFixtures as Mike suggested above, as DemoFixtures enum has constants for both Vault and Checkout (non vault) while VaultAtrributes implies there constants are strictly Vault related

Sources/CorePayments/PayPalCoreConstants.swift
```diff
@@ -9,4 +9,16 @@ public enum PayPalCoreConstants {
...
+    
+    static let integrationChannel: String = "PPCP_NATIVE_SDK"
+}
+
+@_documentation(visibility: private)
+public enum TokenType {
```
sshropshire: This name goes against Apple's [API design guidelines](https://www.swift.org/documentation/api-design-guidelines/) (see example below):

Image

sshropshire (approved):

PR #404, opened 2026-06-05
Payment button analytics fix

NastassiaRodzik:
### Reason for changes

`AnalyticsService` is getting initialized with fake credentials in `PaymentButton.swift.` 
A `payment-button:initialized` event fires against the live FPTI endpoint with clientID: `"N/A" `every time any payment button is instantiated — even in sandbox or unit test environments. This pollutes live analytics with meaningless events and could cause FPTI attribution issues.
**Example:** 
Screenshot 2026-06-04 at 6 01 17 PM

### Summary of changes

- Pass AnalyticsService parameters to PaymentButton
- Do not fire analytics events inside PaymentButton unless valid AnalyticsService instance is available

### Result

Screenshot 2026-06-04 at 5 45 51 PM

Now merchants can track `paymentButton` events along with `paypal-web-payments:..` and others by Order and Client IDs

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik (nrodzik)

### DIFF
--- a/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
+++ b/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
@@ -66,7 +66,7 @@ class PayPalWebViewModel: ObservableObject {
                 self.order = order
                 self.state.createdOrderResponse = .loaded(order)
             }
-            coreConfig = try? await configManager.getCoreConfig()
+            coreConfig = try? await resolvedCoreConfig()
             print("✅ fetched orderID: \(order.id) with status: \(order.status)")
         } catch {
             DispatchQueue.main.async {
@@ -121,10 +121,19 @@ class PayPalWebViewModel: ObservableObject {
         }
     }
 
+    /// Returns the cached `CoreConfig` if available, otherwise fetches it once and caches it.
+    func resolvedCoreConfig() async throws -> CoreConfig {
+        if let coreConfig {
+            return coreConfig
+        }
+        let config = try await configManager.getCoreConfig()
+        coreConfig = config
+        return config
+    }
+
     func getPayPalClient() async throws -> PayPalWebCheckoutClient? {
         do {
-            let config = try await configManager.getCoreConfig()
-            coreConfig = config
+            let config = try await resolvedCoreConfig()
             let payPalClient = PayPalWebCheckoutClient(config: config)
             payPalDataCollector = PayPalDataCollector(config: config)
             return payPalClient
--- a/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
+++ b/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
@@ -16,6 +16,8 @@ class PayPalWebViewModel: ObservableObject {
 
     var payPalWebCheckoutClient: PayPalWebCheckoutClient?
 
+    @Published var coreConfig: CoreConfig?
+
     var orderID: String? {
         order?.id
     }
@@ -64,6 +66,7 @@ class PayPalWebViewModel: ObservableObject {
                 self.order = order
                 self.state.createdOrderResponse = .loaded(order)
             }
+            coreConfig = try? await configManager.getCoreConfig()
             print("✅ fetched orderID: \(order.id) with status: \(order.status)")
         } catch {
             DispatchQueue.main.async {
@@ -121,6 +124,7 @@ class PayPalWebViewModel: ObservableObject {
     func getPayPalClient() async throws -> PayPalWebCheckoutClient? {
         do {
             let config = try await configManager.getCoreConfig()
+            coreConfig = config
             let payPalClient = PayPalWebCheckoutClient(config: config)
             payPalDataCollector = PayPalDataCollector(config: config)
             return payPalClient
@@ -188,6 +192,7 @@ class PayPalWebViewModel: ObservableObject {
         self.state = PayPalPaymentState()
         order = nil
         checkoutResult = nil
+        coreConfig = nil
     }
 
     // for testing until singleton router class is implemented
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -29,10 +29,10 @@ public struct AnalyticsService {
     // MARK: - Internal Initializer
 
     /// Exposed for testing
-    init(coreConfig: CoreConfig, orderID: String?, trackingEventsAPI: TrackingEventsAPI) {
+    init(coreConfig: CoreConfig, trackingEventsAPI: TrackingEventsAPI) {
         self.coreConfig = coreConfig
         self.trackingEventsAPI = trackingEventsAPI
-        self.orderID = orderID
+        self.orderID = nil
         self.setupToken = nil
     }
     
--- a/Sources/CorePayments/Networking/AnalyticsService.swift
+++ b/Sources/CorePayments/Networking/AnalyticsService.swift
@@ -12,7 +12,7 @@ public struct AnalyticsService {
     private let setupToken: String?
     // MARK: - Initializer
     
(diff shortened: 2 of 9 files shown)


Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -29,7 +29,7 @@ public struct AnalyticsService {
     // MARK: - Internal Initializer
 
     /// Exposed for testing
-    init(coreConfig: CoreConfig, orderID: String, trackingEventsAPI: TrackingEventsAPI) {
+    init(coreConfig: CoreConfig, orderID: String?, trackingEventsAPI: TrackingEventsAPI) {
```
sshropshire: I think I'm looking at iOS Analytics code for the first time. I feel like `orderID` shouldn't be a parameter here. @NastassiaRodzik would you be open to removing it?

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -16,6 +16,8 @@ class PayPalWebViewModel: ObservableObject {
 
     var payPalWebCheckoutClient: PayPalWebCheckoutClient?
 
+    @Published var coreConfig: CoreConfig?
```
sshropshire, line 19: q: Does this mean that changes to the `CoreConfig` cause a re-render of the UI?

Sources/PaymentButtons/PaymentButton.swift
```diff
@@ -124,6 +120,18 @@ public class PaymentButton: UIButton {
...
 
+    /// Configures the button's analytics once a `CoreConfig` is available.
+    /// Only the first call has an effect, guaranteeing `payment-button:initialized` is sent at most once.
+    /// - Parameters:
+    ///   - coreConfig: SDK configuration used for checkout.
+    ///   - orderID: PayPal order ID associated with this button session, if available.
+    public func configure(coreConfig: CoreConfig, orderID: String?) {
+        guard !didConfigureAnalytics else { return }
+        didConfigureAnalytics = true
+        analyticsService = AnalyticsService(coreConfig: coreConfig, orderID: orderID)
```
MikeThorntonPayPal, line 129: This is a behavioral breaking change for UIKit consumers who instantiate `PayPalButton` (or subclasses) directly without using the SwiftUI `Representable`. Previously, `payment-button:initialized` and `payment-button:tapped` fired automatically. Now both are silently lost unless the consumer discovers and calls `configure(coreConfig:orderID:)`.

Since there's no compile-time signal that something changed, existing UIKit integrations will silently stop sending button analytics after upgrading.

Consider either:
1. Documenting this as a migration step in the changelog, or
2. Accepting `CoreConfig?` in the `PaymentButton` initializer so analytics are set up at construction time (matching the old behavior) when a config is available.

Sources/PaymentButtons/PaymentButton.swift
```diff
@@ -24,11 +24,8 @@ public class PaymentButton: UIButton {
     static let bundle = Bundle(for: PaymentButton.self)
     #endif
 
-    // Use an empty config and default to live environment for button analytics
-    private let analyticsService = AnalyticsService(
-        coreConfig: .init(clientID: "N/A", environment: .live),
-        orderID: "N/A"
-    )
+    private var analyticsService: AnalyticsService?
```
MikeThorntonPayPal, line 27: `didConfigureAnalytics` is redundant with `analyticsService` itself. You could drop the flag and just guard on the optional:

```swift
guard analyticsService == nil else { return }
analyticsService = AnalyticsService(coreConfig: coreConfig, orderID: orderID)
analyticsService?.sendEvent(...)
```

One less piece of state to keep in sync.

Sources/PaymentButtons/PayPalButton.swift
```diff
@@ -107,11 +116,17 @@ public extension PayPalButton {
 
         public func makeUIView(context: Context) -> PaymentButton {
             button.addTarget(context.coordinator, action: #selector(Coordinator.onAction(_:)), for: .touchUpInside)
+            if let coreConfig {
+                button.configure(coreConfig: coreConfig, orderID: orderID)
+            }
             return button
```
MikeThorntonPayPal, line 122: Because `configure` is one-shot, if the button is rendered before the order is created (common in SwiftUI where the view tree exists before user action), the first `updateUIView` call with a non-nil `coreConfig` but nil `orderID` will lock in `orderID = nil` permanently. Subsequent renders with a real `orderID` will be ignored by the guard.

In the demo this works because `coreConfig` is fetched inside `createOrder()` so both arrive together, but that ordering isn't guaranteed for all consumers.

Worth considering whether `configure` should allow updating the `orderID` if it was initially nil, or at minimum document this ordering requirement.

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -29,10 +29,10 @@ public struct AnalyticsService {
     // MARK: - Internal Initializer
 
     /// Exposed for testing
-    init(coreConfig: CoreConfig, orderID: String, trackingEventsAPI: TrackingEventsAPI) {
+    init(coreConfig: CoreConfig, trackingEventsAPI: TrackingEventsAPI) {
         self.coreConfig = coreConfig
         self.trackingEventsAPI = trackingEventsAPI
```
MikeThorntonPayPal, line 34: The internal test initializer dropped the `orderID` parameter entirely and now hardcodes it to `nil`. This means tests can no longer verify behavior when `orderID` is set through this init path. Was this intentional? If so, consider keeping the parameter as `orderID: String? = nil` so tests that need to supply an orderID still can.

NastassiaRodzik: @MikeThorntonPayPal yes is was requested from @sshropshire in thread above 
CC @sshropshire

NastassiaRodzik: @sshropshire should we keep orderID parameter in internal method?

Sources/PaymentButtons/PaymentButton.swift
NastassiaRodzik: @MikeThorntonPayPal AnalyticsService requires CoreConfig to be non optional in initializer.

The change to wait for valid CoreConfig before firing Analytics was made to avoid sending Analytics events with fake data: client and orderID being "N/A" - or do we want to keep this behavior ? 

Fwiw this and following PRs were made in accordance with the doc https://paypal.atlassian.net/wiki/x/-5F4rw which suggested to omit Analytics events unless valid CoreConfig is present. 
Let's discuss what approach we should go with

Sources/PaymentButtons/PayPalButton.swift
NastassiaRodzik: Fixed here [Option to update payment button orderId](https://github.com/paypal/paypal-ios/pull/404/commits/a87133877f86f21ab8583de1f885e25fe452453f)
Let me know what you think

NastassiaRodzik: @MikeThorntonPayPal on Android they do not have a way to update OrderID once it's available as well. 
Let me know how you want to tackle it going forward, but looks like we might need to come to consistent approach between platforms here

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
NastassiaRodzik: As I see in the code, it does trigger `updateUIView`, however `updateUIView` implementation does not change what's on screen - it only syncs non-visual state such as context.coordinator.action (button action) and configure(config: orderID)

Sources/PaymentButtons/PayPalButton.swift
NastassiaRodzik: @MikeThorntonPayPal on Android they do not have a way to update OrderID once it's available, so PaymentButton Analytics always sends it as `null`. 
Let me know how you would like to tackle it going forward

PR #410, opened 2026-06-30
Shopper Session ID Data Models

NastassiaRodzik:
### Summary of changes

- Add new PayPalUserIdentity, PayPalURLConfig and PayPalUserAction data types
- Modify CoreConfig to prepare to Shopper Session ID implementation 

### Checklist

- [x] Added a changelog entry

- Nastassia Rodzik


Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Deep link URLs used for app switch return and cancel flows.
+public struct PayPalURLConfig {
+
+    public let returnAppUrl: String
+    public let cancelAppUrl: String
+    public let fallbackSchemeUrl: String?
```
kgangineni: I think this is required in payload for createShopperSession
should be non-nullable

NastassiaRodzik: Good catch, even though I see it is still marked as optional in LLD 
Updating

NastassiaRodzik: Fixed 
[Fallback scheme url is no longer optional](https://github.com/paypal/paypal-ios/pull/410/commits/b01f2c6ed9a0129f1db038a5f7c48a5936dc8749)

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// Deep link URLs used for app switch return and cancel flows.
+public struct PayPalURLConfig {
+
+    public let returnAppUrl: String
```
MikeThorntonPayPal: Two things on these three properties, since they're public:
1. Acronym casing — the [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) say acronyms should be uniformly cased, so I'd go `returnAppURL` / `cancelAppURL` / `fallbackSchemeURL`.
2. q: should these be typed as `URL` instead of `String`? We leaned into type safety recently (the typed analytics enums, tightening `CoreSDKError` nullability), and a `URL` would let the compiler reject a malformed string at the call site instead of at runtime. Curious if there's a reason they need to stay `String` (encoding? merchant convenience?).

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// Buyer identity for Shopper Session creation.
+public enum PayPalUserIdentity {
+    case serverSideShopperSession(serverSideShopperSessionId: String)
+    case emailPhone(email: String?, phone: String?)
+    case none
```
MikeThorntonPayPal: q: does having a `case none` here risk colliding with `Optional.none`? If we ever hold a `PayPalUserIdentity?`, `.none` becomes ambiguous between "no identity" and "nil". Wondering if `.anonymous` or `.guest` would read more clearly and sidestep that. (The test file already has to write `case .none` which reads a little confusingly.)

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// Buyer identity for Shopper Session creation.
+public enum PayPalUserIdentity {
+    case serverSideShopperSession(serverSideShopperSessionId: String)
+    case emailPhone(email: String?, phone: String?)
```
MikeThorntonPayPal: nit/random q: since both associated values are optional, `.emailPhone(email: nil, phone: nil)` is representable and is effectively the same as `.none`. Is that a state we want to allow, or should at least one be non-optional so the type can't represent "empty email/phone"? Take or leave — just thinking about making invalid states unrepresentable.

Sources/CorePayments/CoreConfig.swift
```diff
@@ -6,9 +6,13 @@ public struct CoreConfig {
 
     public let environment: Environment
     public let clientID: String
+    public let merchantID: String
+    public let bnCode: String?
```
MikeThorntonPayPal, line 15: q: `bnCode` is a little cryptic for a public property — I had to guess it's the partner/BN attribution code. Could we either spell it out (`partnerAttributionID`?) or add a `///` doc comment explaining what it is and when a merchant would set it? Same goes for `merchantID` on line 9 — a one-liner doc would help since these are the first things merchants see in `CoreConfig`.

Sources/CorePayments/CoreConfig.swift
```diff
@@ -6,9 +6,13 @@ public struct CoreConfig {
...
 
-    public init(clientID: String, environment: Environment) {
+    public init(clientID: String, environment: Environment, merchantID: String, bnCode: String? = nil) {
```
MikeThorntonPayPal, line 17: Since this is a breaking change to a public initializer, let's make sure it lands in the migration guide (not just the CHANGELOG). Ref how we handled the other v3 breaking renames.

Sources/PayPalWebPayments/Models/PayPalUserAction.swift
```diff
@@ -0,0 +1,11 @@
...
+
+/// Buyer action intent for checkout or vault.
+///
+/// Use `.setupNow` with `PayPalVaultRequest` only. Passing `.setupNow` to checkout (`start()`)
+/// returns `PayPalError.invalidUserActionError`.
+public enum PayPalUserAction {
+    case `continue`
+    case payNow
+    case setupNow
+}
```
MikeThorntonPayPal: Love that you documented the `.setupNow` gotcha right on the type — that's exactly the kind of breadcrumb that saves us later. 🔥

q: does `PayPalError.invalidUserActionError` exist yet? I couldn't find it — if it's coming in a later PR, maybe soften the wording so we're not referencing a symbol that isn't there.

Demo/Demo/Networking/MerchantIntegration.swift
```diff
@@ -2,10 +2,13 @@ import Foundation
...
 enum MerchantIntegration: Hashable {
 
-    case direct(clientID: String)
+    case direct(clientID: String, merchantID: String)
 
     static var `default`: MerchantIntegration {
-        return .direct(clientID: "AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo")
+        return .direct(
+            clientID: "AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo",
+            merchantID: "test-merchant-id"
```
MikeThorntonPayPal: nit: `"test-merchant-id"` (and the long `clientID` literal) are now repeated in three places here. Pulling them into `private static let` constants would keep this DRY — pre-existing for the clientID, but since we're already touching all three sites it might be a good moment.

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebCreateOrderView.swift
```diff
@@ -1,17 +1,26 @@
...
+extension PayPalUserAction {
+
+    var title: String {
+        switch self {
+        case .continue:
+            return "CONTINUE"
+        case .payNow:
+            return "PAY NOW"
+        case .setupNow:
+            return "SETUP NOW"
```
MikeThorntonPayPal, line 13: q: the picker hard-codes `[.payNow, .continue]` and drops `.setupNow`, but the extension above still provides a `"SETUP NOW"` title for it. Intentional (since `.setupNow` is vault-only)? If so 👍 — just flagging that the unused title branch might read as dead code to the next person.

MikeThorntonPayPal (changes requested):
Really nice groundwork for Shopper Session ID — the new models are small and focused and the test coverage is a great start. 🙌

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
NastassiaRodzik: fixed

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: Great catch! 
Replaced with `anonymous`

Sources/PayPalWebPayments/Models/PayPalUserAction.swift
NastassiaRodzik: that's correct - the change will be coming in follow up PR, updated the phrasing, let me know what you think

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: that makes sense. 
I can implement this in following PRs when processing user input

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebCreateOrderView.swift
NastassiaRodzik: good call out, let me look into this more closely in the following PRs

MikeThorntonPayPal: Hey, I mentioned this in a direct message but wanted to post you for visibility. It's generally best practice to address requested code changes in the PR in which they are suggested. Otherwise, it becomes difficult to track whether or not the changes were in fact made, additionally, it requires reviewers to potentially review the same code multiple times and catch the same suggestion multiple times

NastassiaRodzik: @MikeThorntonPayPal I'd leave a title of .setupNow here to use it for Vault flow later - what do you think?

MikeThorntonPayPal (approved):

Demo/Demo/Networking/MerchantIntegration.swift
```diff
@@ -31,14 +41,23 @@ enum MerchantIntegration: Hashable {
...
-            .direct(clientID: "AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo")
+            .direct(
+                clientID: testClientID,
+                merchantID: testMerchantID
+            )
         ]
     }
+    
+    private static let testClientID = "AQTfw2irFfemo-eWG4H5UY-b9auKihUpXQ2Engl4G1EsHJe2mkpfUv_SN3Mba0v3CfrL6Fk_ecwv9EOo"
+    private static let testMerchantID = "test-merchant-id"
```
sshropshire, line 52: q: is this test code supposed to be here?

Demo/Demo/PayPalWebPayments/PayPalWebPaymentsView/PayPalWebCreateOrderView.swift
```diff
@@ -1,17 +1,30 @@
...
+        }
+    }
+    
+    static var checkoutActions: [PayPalUserAction] {
+        [.payNow, .continue]
+    }
```
sshropshire:
```suggestion
    static let checkoutActions: [PayPalUserAction] = [.payNow, .continue]
```
I feel like this makes more sense.

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
```diff
@@ -0,0 +1,15 @@
...
+
+    public let returnAppURL: URL
+    public let cancelAppURL: URL
+    public let fallbackSchemeURL: URL
+
+    public init(returnAppURL: URL, cancelAppURL: URL, fallbackSchemeURL: URL) {
+        self.returnAppURL = returnAppURL
+        self.cancelAppURL = cancelAppURL
+        self.fallbackSchemeURL = fallbackSchemeURL
+    }
```
sshropshire:
```suggestion
```
Can we remove this? The struct should have a default constructor if i'm not mistaken.

Sources/PayPalWebPayments/Models/PayPalUserAction.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
+
+/// Buyer action intent for checkout or vault.
+///
+/// Use `.setupNow` with `PayPalVaultRequest` only. Passing `.setupNow` to checkout (`start()`)
+/// will return `PayPalError.invalidUserActionError`.
+public enum PayPalUserAction {
+    case `continue`
```
sshropshire, line 8: Any chance this could be named something other than continue? Will merchants have to type the backticks as well?

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+/// Buyer identity for Shopper Session creation.
+public enum PayPalUserIdentity {
+    case serverSideShopperSession(serverSideShopperSessionId: String)
```
sshropshire:
```suggestion
    case shopperSession(shopperSessionId: String)
```

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
sshropshire: question: Why is this an `enum` again? It feels like it makes more sense as a `struct` with all the associated values.

UnitTests/CardPaymentsTests/CardClient_Tests.swift
```diff
@@ -17,7 +17,7 @@ class CardClient_Tests: XCTestCase {
         expirationYear: "2021",
         securityCode: "123"
     )
-    let config = CoreConfig(clientID: "mockClientId", environment: .sandbox)
+    let config = CoreConfig(clientID: "mockClientId", environment: .sandbox, merchantID: "mockMerchantID")
```
sshropshire:
```suggestion
    let config = CoreConfig(clientID: "mockClientID", environment: .sandbox, merchantID: "mockMerchantID")
```
nit: for consistency.

UnitTests/CorePaymentTests/PatchCCOWithAppSwitchEligibility.swift
```diff
@@ -143,7 +143,7 @@ final class PatchCCOWithAppSwitchEligibility_Tests: XCTestCase {
         mockAuthSTS.stubbedError = CoreSDKError(code: 401, domain: "auth", errorDescription: "unauthorized")
 
         do {
-            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "t", tokenType: ExternalTokenKind.clientToken, , canSwitchToApp: true)
+            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "t", tokenType: ExternalTokenKind.clientToken, canSwitchToApp: true)
```
sshropshire:
```suggestion
            _ = try await sut.patchCCOWithAppSwitchEligibility(token: "fake-token", tokenType: ExternalTokenKind.clientToken, canSwitchToApp: true)
```

UnitTests/PayPalWebPaymentsTests/PayPalUserIdentity_Tests.swift
```diff
@@ -0,0 +1,63 @@
...
+    // MARK: - .none
+
+    func testNone_isNoneCase() {
+        let identity = PayPalUserIdentity.anonymous
+
+        guard case .anonymous = identity else {
+            XCTFail("Expected .none case")
+            return
+        }
+    }
```
sshropshire:
```suggestion
```
I'm not sure if this is actually testing anything.

UnitTests/PayPalWebPaymentsTests/PayPalUserIdentity_Tests.swift
```diff
@@ -0,0 +1,63 @@
+import XCTest
+@testable import PayPalWebPayments
+
+class PayPalUserIdentity_Tests: XCTestCase {
```
sshropshire: Actually this whole test doesn't seem to be doing much.

Sources/CorePayments/CoreConfig.swift
```diff
@@ -7,8 +7,17 @@ public struct CoreConfig {
     public let environment: Environment
     public let clientID: String
 
-    public init(clientID: String, environment: Environment) {
+    /// The PayPal merchant account identifier associated with this integration.
+    public let merchantID: String
```
sshropshire, line 11: question: Why is merchant ID required? It shouldn't be required for our `:Card` module. This makes me feel like it's better for these fields to go into the request objects instead of the `CoreConfig`.

Demo/Demo/Networking/MerchantIntegration.swift
NastassiaRodzik: I believe yes, as this is test data we use on Demo to create a shopper session

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
NastassiaRodzik: This is to make initializer `public` in order for Demo / merchant to use it. 
Otherwise it is `internal` by default and merchants cannot create an instance of PayPalURLConfig..

Sources/PayPalWebPayments/Models/PayPalUserAction.swift
NastassiaRodzik: no merchants do not have to type the backticks - this is for Swift compiler to know we are "reusing" words it is familiar with, and not to complain about it with errors

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: yeah this is how I've done it initially, a struct :)
but then I figured: PayPalUserIdentity cannot have both email and serverSideShopperSessionId present, right? 
Making it an `enum` with associated values takes care of this restriction

NastassiaRodzik: ( also Uncle Bob recommended this in his Clean Architecture )

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: `serverSideShopperSessionId` and `shopperSessionId` are different things: UserIdentity.serverSideShopperSessionId is a merchant-provided pre-existing session (created server-side via the Shopper Insights API), while the actual shopperSessionID (SSID) is a new session created by the SDK via the createShopperSessionWithAppSwitchEligibility mutation

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
sshropshire: I understand if we want to go strictly with Uncle Bob's recommendations, but something feels off to me with the naming. Like `let userIdentity = PayPalUserIdentity.shopperSession("session-id")` doesn't feel semantically correct. I can see how a shopper session id is eventually linked to an identity, but it shouldn't be of type "Identity" itself. Does that make sense?

You could also technically enforce these same constraints with separate constructors:

```swift
public struct PayPalUserIdentity {
  let shopperSessionID: String?
  let email: String?
  let phone: String?

  private init(shopperSessionID: String?, email: String?, phone: String?) {
    self.shopperSessionID = shopperSessionID
    self.email = email
    self.phone = phone
  }

  public convenience init(shopperSessionID: String) {
    self.init(shopperSessionID: shopperSessionID, email: nil, phone: nil)
  }

  public convenience init(email: String?, phone: String?) {
    self.init(shopperSessionID: nil, email: email, phone: phone)
  }
}
```

sshropshire: I guess the enum jumped out to me as a code smell as well. I'm fundamentally still not a huge fan of the naming, but Swift has a ton of tools that would allow us flexibility with the API. That to me is the root cause of the complexity of both the `enum` and the example I provided.

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
sshropshire: Are they actually different though? Would the merchant server be using a different endpoint than the SDK to create a shopper session?

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
sshropshire: Oh wow ok I didn't know that before. It's surprising given that the type is `public` I didn't expect `internal` to be the default. Thanks!

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
sshropshire: Only asking because if there's an opportunity for simplification we should take it. The shorter name would go a long way for readability.

NastassiaRodzik: Yes, they will be using different endpoint + different protocol (REST /v1/customer/sessions vs GraphQL createShopperSessionWithAppSwitchEligibility)

sshropshire: Right but the GraphQL server will be using the same REST endpoint. I think they're the same entity.

NastassiaRodzik: I believe GraphQL server will also check the app switch eligibility, comparing to REST endpoint

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: Replaced `enum` with `structure`

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: So I'd keep `serverSideShopperSessionId` naming to distinguish between pre-existing shopper session that merchant provided vs shopperSession that was created by SDK + already passed app switch check. 
This is also what LLD suggests.
Let me know your thoughts

Sources/CorePayments/CoreConfig.swift
NastassiaRodzik: It's required here to pass into createShopperSessionWithAppSwitchEligibility method.
Started the thread in Slack on this

NastassiaRodzik: Update: resolved in Slack

Sources/PayPalWebPayments/Models/PayPalUserIdentity.swift
NastassiaRodzik: update: will update to `existingPayPalSession` for iOS and Android

NastassiaRodzik: Update: actually this is already resolved after making PayPalUserIdentity a structure

PR #411, opened 2026-07-01
Shopper Session ID PayPalWebCheckoutClient changes 

NastassiaRodzik:
### Summary of changes

- Add `createPayPalSession()` to PayPalWebCheckoutClient and simplify `start()` and `vault()` to accept order/token IDs directly

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik


Sources/PayPalWebPayments/APIRequests/CreateShopperSessionResponse.swift
```diff
@@ -0,0 +1,45 @@
...
+struct ShopperSessionResult: Decodable {
+
+    /// Whether the buyer's device is eligible for PayPal app-switch.
+    let appSwitchEligible: Bool
+
+    /// The app-switch redirect URL to open when eligible.
+    let redirectURL: String?
+
+    /// Fallback web checkout URL if app-switch is unavailable.
+    let checkoutFallbackUrl: String?
```
MikeThorntonPayPal: `redirectURL` (uppercase `URL`) sits next to `checkoutFallbackUrl` (lowercase `Url`) in the same struct. Let's make the acronym casing uniform per the [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/)  `checkoutFallbackURL`. If the JSON keys differ, `CodingKeys` can bridge it.

Sources/PayPalWebPayments/APIRequests/CreateShopperSessionAPI.swift
```diff
@@ -0,0 +1,113 @@
+import Foundation
+
+#if canImport(CorePayments)
+import CorePayments
+#endif
+
+/// Coordinates the GraphQL call that pre-warms a Shopper Session with app-switch eligibility.
+@_documentation(visibility: private)
+public class CreateShopperSessionAPI {
```
MikeThorntonPayPal: q: does this need to be `public`? It's `@_documentation(visibility: private)`, only injected into the internal init, and the mock subclasses it via `@testable import` (which already lifts access). Same instinct as the `PayPalURLConfig` init discussion. Just wanting to keep the public surface as small as necessary.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -444,19 +712,28 @@ public class PayPalWebCheckoutClient: NSObject {
...
-    private func notifyVaultCancelWithError(with vaultError: CoreSDKError, completion: (Result<PayPalVaultResult, CoreSDKError>) -> Void) {
+    private func notifyVaultCancelWithError(
+        with error: CoreSDKError,
+        completion: (Result<PayPalVaultResult, CoreSDKError>) -> Void
+    ) {
         analyticsService?.sendEvent("paypal-web-payments:vault-wo-purchase:canceled")
-        completion(.failure(vaultError))
+        completion(.failure(error))
     }
 }
```
MikeThorntonPayPal, line 657: this and the legacy `startWebCheckoutFlow(request:)` are ~55 of ~60 lines identical — the only real differences are the token (`orderID` vs `request.orderID`) and funding source (`.paypal` vs `request.fundingSource`). We could collapse both into a single internal `startWebCheckoutFlow(orderID:fundingSource:completion:)` that each path calls, passing those two values in. Since the legacy path never sets `sessionTask`, always clearing it is harmless, so it unifies cleanly — and your new CreateSession tests already cover it. I'd do it here rather than leave a TODO, so a future fix to the web flow can't silently miss one path.

Sources/PayPalWebPayments/APIRequests/CreateShopperSessionAPI.swift
NastassiaRodzik: great catch, fixed

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: yes absolutely! fixed

Sources/PayPalWebPayments/APIRequests/CreateShopperSessionResponse.swift
NastassiaRodzik: just received an updated mutation from team working on createSession() GQL today - they don't have `checkoutFallbackUrl` there, removed it

Sources/PayPalWebPayments/APIRequests/CreateShopperSessionVariables.swift
```diff
@@ -0,0 +1,37 @@
+import Foundation
+
+struct CreateShopperSessionVariables: Encodable {
+
+    let returnUrl: URL
+    let cancelUrl: URL
+    let fallbackSchemeUrl: URL?
```
MikeThorntonPayPal: nit: same acronym-casing rule we settled on earlier — let's make these `returnURL` / `cancelURL` / `fallbackSchemeURL` so the `URL` suffix is uniform. Since this struct is `Encodable` and the keys have to stay `returnUrl` / `cancelUrl` / `fallbackSchemeUrl` to match the GraphQL variables, add `CodingKeys` to map them back to the wire names.

MikeThorntonPayPal (approved):

PR #419, opened 2026-07-14
Shopper Session ID analytics

NastassiaRodzik:
### Reason for changes

Shopper Session ID 

### Summary of changes

- Add Analytics to SSID flow 

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik


Sources/CorePayments/AnalyticsEventData.swift
```diff
@@ -13,25 +13,31 @@ struct AnalyticsEventData: Encodable {
     enum EventParameterKeys: String, CodingKey, CaseIterable {
         case appID = "app_id"
         case appName = "app_name"
+        case appSwitchURL = "app_switch_url"
         case buttonType = "button_type"
         case clientID = "partner_client_id"
-        case clientSDKVersion = "c_sdk_ver"
         case clientOS = "client_os"
+        case clientSDKVersion = "c_sdk_ver"
```
NastassiaRodzik, line 20: No new changes except adding new parameters here: 
             
                appSwitchURL,
                errorDescription,
                isCachedSession,
                isVaultRequest,
                shopperSessionId,
                startTime. 

+ Sorted EventParameterKeys alphabetically

Sources/CorePayments/Networking/AnalyticsService.swift
```diff
@@ -43,33 +56,81 @@ public struct AnalyticsService {
...
+    public func sendEvent(
+        _ name: String,
+        correlationID: String? = nil,
+        buttonType: String? = nil,
+        appSwitchURL: URL? = nil,
+        errorDescription: String? = nil,
+        isCachedSession: Bool? = nil,
+        isVaultRequest: Bool? = nil,
+        shopperSessionId: String? = nil,
+        startTime: Int? = nil
```
sshropshire, line 74: This is fine for now, but the parameter list is starting to grow. It may make sense to revisit this method in the future and consider using a dictionary with `enum` keys to prevent having to modify the method signature every time a new parameter can be tracked.

Sources/CorePayments/AnalyticsEventData.swift
```diff
@@ -13,25 +13,31 @@ struct AnalyticsEventData: Encodable {
...
         case component = "comp"
         case correlationID = "correlation_id"
         case deviceManufacturer = "device_manufacturer"
+        case deviceModel = "mobile_device_model"
```
sshropshire, line 24: Would it be possible to rename `mobile_device_model` to `device_model`?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -32,6 +32,29 @@ public class PayPalWebCheckoutClient: NSObject {
...
+
+    /// The Shopper Session ID for the checkout/vault attempt currently in flight. Captured as soon as
+    /// the Shopper Session fetch succeeds in `start()`/`vault()`, so every analytics event fired for the
+    /// remainder of that attempt — app-switch or web fallback, success, cancellation, or failure — can
+    /// report `shopper_session_id`, not just the app-switch/`handleReturnURL` events that used to be the
+    /// only ones with access to it (back when this was `shopperSessionID`, set only from within
+    /// `attemptSessionAppSwitch`).
+    /// Cleared once the attempt reaches a terminal outcome, in `handleReturnURL`, `startWebCheckoutFlow`,
+    /// or `startVaultWebAuthFlow` — never in `start()`/`vault()`'s own defer, since those fire before the
+    /// downstream app-switch/web flow (which still needs the ID) has even started.
```
sshropshire: This comment is really verbose.

sshropshire: Do we need these comments?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -127,13 +161,30 @@ public class PayPalWebCheckoutClient: NSObject {
 
         Task {
             do {
+                defer {
+                    sessionCreationStartTime = nil
+                    isCachedSession = nil
+                    sessionTask = nil
```
sshropshire, line 152: Is there a risk that this `defer` block gets called before the completion handler is fired? Does it matter?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -572,13 +694,16 @@ public class PayPalWebCheckoutClient: NSObject {
                 url: payPalCheckoutURLComponents,
                 context: self,
                 sessionDidDisplay: { [weak self] didDisplay in
-                    let event = didDisplay
-                        ? "paypal-web-payments:checkout:auth-challenge-presentation:succeeded"
-                        : "paypal-web-payments:checkout:auth-challenge-presentation:failed"
```
sshropshire, line 577: qq: are we no longer sending this event?

sshropshire (approved):
🌊 fundamentally sound! I left some notes, but they are higher level. There may be an opportunity to refactor toward a cleaner design in the near future.

Sources/CorePayments/AnalyticsEventData.swift
NastassiaRodzik: `deviceModel` is an existing parameter that we didn't modify in this PR (it appears as new/green after I sorted the keys alphabetically, but it was there on line 30 in "before" state) - I wouldn't make a call on modifying it myself as most likely it is used by Analytics already

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: We are still sending it, but `error` event now takes place on line 684 and line 715 where we have access to the error object / error details for Analytics.
`succeeded` event is still in `sessionDidDisplay` block

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: It's perfectly fine if it fires before the completion handler: all these parameters are for internal use only and here we just cleaning them once they serve their purpose (Analytics parameters are sent and session task is completed)

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: Removed them ✅

Sources/CorePayments/AnalyticsEventData.swift
sshropshire: Ah ok. Yeah hopefully we can get that one cleaned up on the analytics side. As a best practice, I like when the `camelCase` name lines up with the `snake_case` name.

PR #422, opened 2026-07-22
Fix SSID token type handling and propagate config-update errors

joaaraujo-paypal:
### Summary of changes
- Replace `PayPalFlowType` with `TokenType` (`.orderID`, `.vaultID`, `.billingToken`) to correctly set the token query-parameter name across app-switch and web fallback/checkout URLs for order, vault, and billing-token flows.
- Stop silently swallowing `updateClientConfig` failures in `start()`/`vault()` — errors now propagate to the completion handler and are reported via analytics instead of continuing with a stale client config.
- Improve non-JSON/unstructured HTTP error messages to include status code, request URL, and raw body, making SSID/app-switch failures easier to diagnose.

- [x] Added a changelog entry

- @joaaraujo-paypal


Sources/CorePayments/Networking/HTTP.swift
```diff
@@ -35,7 +35,7 @@ class HTTP {
         guard let response = response as? HTTPURLResponse else {
             throw NetworkingError.invalidURLResponseError
         }
-        
-        return HTTPResponse(status: response.statusCode, body: data)
+
+        return HTTPResponse(status: response.statusCode, body: data, url: httpRequest.url)
```
kgangineni, line 39: any reason for url in `HTTPResponse`

MikeThorntonPayPal (commented):
Really solid fix — replacing the `oneTime`/`billingWithoutPurchase` flow types with token semantics, plus finally surfacing `updateClientConfig` failures instead of swallowing them, are both good calls. A few thoughts below, mix of naming nits and one design q 🌊

**1. q: does `TokenType` leak an implementation detail into the public API?**

`PayPalFlowType.oneTime` / `.billingWithoutPurchase` told the merchant what *they* were doing. `TokenType.orderID` / `.vaultID` / `.billingToken` tells them what *we* call the token internally — it's literally the GraphQL `tokenType` input value via `.rawValue` (`CreateShopperSessionAPI.swift`), and separately backs the URL query-param name via `tokenQueryParameterName`. Now a merchant calling `createPayPalSession(tokenType: .orderID, ...)` has to know our token taxonomy just to say "I'm doing a one-time checkout."

Would something like `PayPalSessionType` (`.checkout`, `.vaultWithoutPurchase`, ...) with an internal mapping to the raw `TokenType`/query-param name preserve the readable call site while keeping the wire-format details private? Linking the [Swift API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) since this feels like the "name conveys role, not representation" guidance.

**2. q: `preconditionFailure` in `currentTokenType` — is a crash the right failure mode here?**

```swift
private var currentTokenType: TokenType {
    guard let tokenType else {
        preconditionFailure("tokenType must be set before building a checkout/vault URL")
    }
    return tokenType
}
```

This is enforced entirely by call-order convention — every public entry point has to remember to set `tokenType` before any code path reaches `launchCheckout`/`launchVault`/`attemptAppSwitchIfEligible`. Today all paths do, but this is a public SDK running inside merchant apps — if a future change ever reaches this without going through `createPayPalSession()` first, we crash their app instead of returning a `CoreSDKError`. 😬 Would love a fallback error path instead of `preconditionFailure`, purely for defense in depth — take or leave, since I know today's call graph is safe.

**3. nit: `flow_type` derivation could be a `switch`**

```swift
URLQueryItem(name: "flow_type", value: (tokenType == .vaultID ? FlowType.vault : .checkout).rawValue),
```

Now that `TokenType` has 3 cases, a `switch` here (vs. the `==`/ternary) would force us to explicitly decide what `.billingToken` maps to instead of it silently falling into `.checkout`. 🧶 super nitpick / take-or-leave.

**4. random q: `.billingToken` — reserved for follow-up work?**

I don't see any production call site setting `tokenType = .billingToken` yet — only exercised in `PayPalWebCheckoutURLBuilder_Tests`. Totally fine if it's landing ahead of the feature that uses it, just flagging in case it was meant to be wired up in this PR.

**5. q: is the cancel → `.error(...)` change in the Demo intentional?**

`PayPalWebViewModel`/`PayPalVaultViewModel` used to set `.idle` on `PayPalError.checkoutCanceledError`/`.vaultCanceledError`; now they set `.error(message: "PayPal checkout/vault was canceled.")`, and `PayPalApprovalResultView` renders `.error` through `ErrorView`. So a plain user-initiated cancel now shows the same red error UI as a real failure. This isn't mentioned in the PR description, so wanted to check it's deliberate — if so, might be worth a distinct "canceled" visual treatment so users don't read "canceled" as "something broke." 🤦 if I'm missing context here.

Nothing here blocks — the core fix (token type correctness + surfacing config-update failures + richer error diagnostics) is exactly what was needed.

🌊 Fundamentally solid. LGTM

MikeThorntonPayPal (approved):

MikeThorntonPayPal (approved):
Nice, this addressed everything from the last pass:

- `TokenType` is internal now with `PayPalSessionType` as the public-facing API — call sites read as intent (`.checkout`/`.vaultWithoutPurchase`) again instead of leaking the GraphQL token taxonomy. 🔥
- `preconditionFailure` is gone — `tokenType` is now an `Optional` that falls back gracefully instead of crashing the host app.
- `flow_type` derivation is a proper `switch` over all three `TokenType` cases, so `.billingToken` isn't silently falling through anymore.

random q: `.billingToken` still only shows up in `PayPalWebCheckoutURLBuilder_Tests` — still assuming that's landing ahead of the feature that sets it in production, not a leftover.

🌊 Fundamentally solid. LGTM

MikeThorntonPayPal (approved):
LGTM

Sources/CorePayments/Networking/HTTP.swift
joaaraujo-paypal: It's consumed by `HTTPResponseParser.swift`, which appends the request URL to the error message when a response isn't valid/structured JSON.

PR #424, opened 2026-07-24
Latency Analytics events 

NastassiaRodzik:
### Reason for changes

SSID Analytics 

### Summary of changes

- Log API Call Latency, createShopperSession() API call
- Log User-Perceived Latency, measures elapsed time from when the merchant calls start() or vault() until the checkout experience is launched 

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -668,7 +668,6 @@ public class PayPalWebCheckoutClient: NSObject {
 
     // MARK: - Private: System Latency
 
-    /// Forwards to `systemLatency`, supplying the client's current `analyticsService` and `analyticsData`.
     private func sendSystemLatencyEvent(_ presentationType: SystemLatencyTracker.PresentationType) {
         systemLatency.send(
             presentationType: presentationType,
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -107,7 +107,12 @@ public class PayPalWebCheckoutClient: NSObject {
 
         let tokenType = sessionType.tokenType
         self.tokenType = tokenType
-        analyticsData = PayPalCheckoutAnalyticsData(tokenType: tokenType, userIdentity: userIdentity, urlConfig: urlConfig, userAction: userAction)
+        analyticsData = PayPalCheckoutAnalyticsData(
+            tokenType: tokenType,
+            userIdentity: userIdentity,
+            urlConfig: urlConfig,
+            userAction: userAction
+        )
         analyticsService?.sendEvent("paypal-web-payments:checkout:ssid-session:started")
 
         sessionTask = Task {
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -154,7 +154,7 @@ public class PayPalWebCheckoutClient: NSObject {
                 errorDescription: PayPalError.sessionNotStartedError.errorDescription,
                 checkoutAnalyticsData: analyticsData
             )
-            sendSystemLatencyEvent(.error)
+            endSystemLatencyTracking(presentationType: .error)
             DispatchQueue.main.async {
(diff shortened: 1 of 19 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -649,6 +666,16 @@ public class PayPalWebCheckoutClient: NSObject {
         }
     }
 
+    // MARK: - Private: System Latency
+
+    private func sendSystemLatencyEvent(_ presentationType: SystemLatencyTracker.PresentationType) {
```
```suggestion
    private func endSystemLatencyTracking(_ presentationType: SystemLatencyTracker.PresentationType) {
```
Would this be more appropriate? I may need some clarification.

Also, the omission of the parameter name is affecting readability for me. In my mind `sendSystemLatencyEvent(.browser)` reads like `sendSystemLatencyEvent(eventName: .browser)`. An explicit parameter name would be helpful here to make it read more clearly.

UnitTests/PayPalWebPaymentsTests/CreateShopperSessionAPI_Tests.swift
```diff
@@ -0,0 +1,117 @@
...
+        var onSendEvent: (() -> Void)?
+
+        override func sendEvent(with analyticsEventData: AnalyticsEventData) async throws -> HTTPResponse {
+            capturedEventData = analyticsEventData
+            onSendEvent?()
+            return HTTPResponse(status: 200, body: nil)
+        }
+    }
+
+    private let successGraphQLBody = """
```
sshropshire:
```suggestion
    private let createShopperSessionSuccessGraphQLBody = """
```

Sources/PayPalWebPayments/SystemLatencyTracker.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+#if canImport(CorePayments)
+import CorePayments
+#endif
+
+/// Tracks user-perceived ("system") latency for a single checkout or vault flow and emits the
+/// `paypal-web-payments:system-latency` analytics event at most once per flow.
+final class SystemLatencyTracker {
```
sshropshire, line 9: curious: Is marking classes `final` a best practice?

UnitTests/PayPalWebPaymentsTests/SystemLatencyTracker_Tests.swift
```diff
@@ -0,0 +1,102 @@
...
+@testable import CorePayments
+@testable import PayPalWebPayments
+@testable import TestShared
+
+class SystemLatencyTracker_Tests: XCTestCase {
+
+    // MARK: - Helpers
+
+    /// Captures analytics events dispatched by the (fire-and-forget) `AnalyticsService`.
+    private final class CapturingTrackingEventsAPI: TrackingEventsAPI {
```
sshropshire: It would be nice if we could extract a protocol for `TrackingEventsAPI` so we could mock without subclassing.

Sources/PayPalWebPayments/SystemLatencyTracker.swift
NastassiaRodzik: yep, for better performance (compiler knows it cannot be subclassed and makes optimizations)

sshropshire (approved):

Sources/PayPalWebPayments/SystemLatencyTracker.swift
sshropshire: Nice we should consider following up and adding finalizers to all the other SDK types.

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -137,28 +145,29 @@ public class PayPalWebCheckoutClient: NSObject {
         orderID: String,
         completion: @escaping (Result<PayPalWebCheckoutResult, CoreSDKError>) -> Void
     ) {
+        analyticsService = AnalyticsService(coreConfig: config, orderID: orderID)
```
sshropshire, line 148: This is out of scope, I do wonder now if there's a benefit to binding an `AnalyticsService` to an `orderID`. If we could move `orderID` as a parameter to `sendEvent()`, we would only need to instantiate a single analytics service instance.

sshropshire (approved):

PR #425, opened 2026-07-28
Shopper session fallback update

NastassiaRodzik:
### Reason for changes

- On Shopper Session creation error, we show error instead of falling back to PatchCCOWithAppSwitchEligibility

### Summary of changes

- Update PayPalWebCheckout Client 
- Introduce new error
- Update tests
- Add platform = iOS query param to return url (Demo app)

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Demo/Demo/PayPalVault/PayPalVaultViewModel/PayPalVaultViewModel.swift
+++ b/Demo/Demo/PayPalVault/PayPalVaultViewModel/PayPalVaultViewModel.swift
@@ -25,10 +25,18 @@ class PayPalVaultViewModel: VaultViewModel {
             phone: userPhone,
             ssid: userSSID
         )
+        let urlConfig: PayPalURLConfig
+        do {
+            urlConfig = try ShopperSessionURLConfigFactory.makeURLConfig()
+        } catch {
+            state.setupTokenResponse = .error(message: error.localizedDescription)
+            throw error
+        }
+
         client.createPayPalSession(
             sessionType: .vaultWithoutPurchase,
             userIdentity: resolvedUserIdentity,
-            urlConfig: ShopperSessionURLConfigFactory.urlConfig,
+            urlConfig: urlConfig,
             userAction: selectedUserAction
         )
 
--- a/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
+++ b/Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
@@ -42,10 +42,18 @@ class PayPalWebViewModel: ObservableObject {
         }
         payPalWebCheckoutClient = client
 
+        let urlConfig: PayPalURLConfig
+        do {
+            urlConfig = try ShopperSessionURLConfigFactory.makeURLConfig()
+        } catch {
+            state.createdOrderResponse = .error(message: error.localizedDescription)
+            throw error
+        }
+
         client.createPayPalSession(
             sessionType: .checkout,
             userIdentity: userIdentity,
-            urlConfig: ShopperSessionURLConfigFactory.urlConfig,
+            urlConfig: urlConfig,
             userAction: userAction
         )
 
--- a/Demo/Demo/SwiftUIComponents/CommonComponents/ShopperSessionURLConfigFactory.swift
+++ b/Demo/Demo/SwiftUIComponents/CommonComponents/ShopperSessionURLConfigFactory.swift
@@ -6,12 +6,22 @@ import PayPalWebPayments
 /// Used by both the PayPalWeb and Vault flows so the callback URLs stay in sync.
 enum ShopperSessionURLConfigFactory {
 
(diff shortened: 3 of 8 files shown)


Demo/Demo/SwiftUIComponents/CommonComponents/ShopperSessionURLConfigFactory.swift
```diff
@@ -9,11 +9,19 @@ enum ShopperSessionURLConfigFactory {
...
               let cancelAppURL = URL(string: "\(baseReturnUrl)/cancel") else {
             fatalError("Failed to construct Shopper Session callback URLs for base return URL: \(baseReturnUrl)")
         }
 
+        returnAppURLComponents.queryItems = (returnAppURLComponents.queryItems ?? []) + [
+            URLQueryItem(name: "platform", value: "iOS")
+        ]
+
+        guard let returnAppURL = returnAppURLComponents.url else {
+            fatalError("Failed to construct Shopper Session callback URLs for base return URL: \(baseReturnUrl)")
```
sshropshire: This will crash the merchant app. Is there an alternative we can use in this case?

sshropshire: Maybe returning an explicit error.

NastassiaRodzik: @sshropshire this is only in Demo, but I can pass the error up and show on the UI / use more gentle handling

sshropshire: Yeah might as well. I know it's extremely rare but it's worth the extra effort.

NastassiaRodzik: done 
[No fatal error](https://github.com/paypal/paypal-ios/pull/425/commits/bfed8cc6a2c95030c8d2e1762e3ea572b2c3006f)

Demo/Demo/PayPalVault/PayPalVaultViewModel/PayPalVaultViewModel.swift
```diff
@@ -25,10 +25,18 @@ class PayPalVaultViewModel: VaultViewModel {
             phone: userPhone,
             ssid: userSSID
         )
+        let urlConfig: PayPalURLConfig
+        do {
+            urlConfig = try ShopperSessionURLConfigFactory.makeURLConfig()
+        } catch {
+            state.setupTokenResponse = .error(message: error.localizedDescription)
+            throw error
```
sshropshire, line 33: We may no longer need to `throw` here since the error will now visible in the UI.

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
```diff
@@ -42,10 +42,18 @@ class PayPalWebViewModel: ObservableObject {
         }
         payPalWebCheckoutClient = client
 
+        let urlConfig: PayPalURLConfig
+        do {
+            urlConfig = try ShopperSessionURLConfigFactory.makeURLConfig()
+        } catch {
+            state.createdOrderResponse = .error(message: error.localizedDescription)
+            throw error
```
sshropshire, line 50: Same, we may be able to remove the `throw`.

sshropshire (approved):

Demo/Demo/PayPalVault/PayPalVaultViewModel/PayPalVaultViewModel.swift
NastassiaRodzik: this is to make sure Demo doesn't attempt to proceed the flow with invalid urlConfig (early return)

Demo/Demo/PayPalWebPayments/PayPalWebViewModel/PayPalWebViewModel.swift
NastassiaRodzik: same here as well: keeping it as an early return from the method that cannot proceed with invalid urlConfig

Demo/Demo/PayPalVault/PayPalVaultViewModel/PayPalVaultViewModel.swift
sshropshire: Ok I understand. I see another potential improvement that I'd say is out of scope for this PR.

After taking a closer look at the code, I see there a few pieces in `PayPalVaultViewModel.vault()` that throw. We should consider refactoring all of our ViewModels to not throw. Ideally, all errors should be routed to SwiftUI via the `state` objects we have. That way, we pass control to the user, and they can choose to retry if something went wrong.

sshropshire (approved):

sshropshire (approved):

PR #426, opened 2026-07-28
Add returnAppUrl/fallbackSchemeUrl validation error

EdgarBarocio:
### Reason for changes
To have parity with Android SDK, we check the returned returnAppUrl and fallbackSchemeUrl, and fail early in case they are both empty.

In the case of returnAppUrl, because it's not an optional value, we check for usability of the url.

Ticket: https://paypal.atlassian.net/browse/DTPPMOBILE-580

### Summary of changes

- Added new error
- Added validations before session creations to avoid calls
- Added tests

### Checklist

- [ ] Added a changelog entry

### DIFF
--- a/Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
+++ b/Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
@@ -13,3 +13,16 @@ public struct PayPalURLConfig {
         self.fallbackSchemeURL = fallbackSchemeURL
     }
 }
+
+extension PayPalURLConfig {
+
+    /// True when neither `returnAppURL` nor `fallbackSchemeURL` is usable (both nil/blank).
+    /// Mirrors Android's ReturnToAppUrlConfig validation.
+    var isReturnToAppConfigMissing: Bool {
+        func isBlank(_ url: URL?) -> Bool {
+            guard let url else { return true }
+            return url.absoluteString.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        }
+        return isBlank(returnAppURL) && isBlank(fallbackSchemeURL)
+    }
+}
--- a/Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
+++ b/Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
@@ -16,13 +16,23 @@ public struct PayPalURLConfig {
 
 extension PayPalURLConfig {
 
-    /// True when neither `returnAppURL` nor `fallbackSchemeURL` is usable (both nil/blank).
-    /// Mirrors Android's ReturnToAppUrlConfig validation.
+    /// True when neither `returnAppURL` nor `fallbackSchemeURL` is usable.
+    /// A URL is unusable when it's nil, blank, or "scheme-only" (no host and no path,
+    /// e.g. `myapp://` or `https://`) — i.e. it can't route the buyer back into the app.
+    /// Defensive early-fail so `start()`/`vault()` can fail immediately without a network call;
+    /// this is a guard, not full parity with Android's nullable `ReturnToAppUrlConfig`.
     var isReturnToAppConfigMissing: Bool {
-        func isBlank(_ url: URL?) -> Bool {
-            guard let url else { return true }
-            return url.absoluteString.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
-        }
-        return isBlank(returnAppURL) && isBlank(fallbackSchemeURL)
+        isUnusable(returnAppURL) && isUnusable(fallbackSchemeURL)
+    }
+
+    /// A URL is "unusable" for return-to-app routing if it's nil, blank after trimming, or
+    /// scheme-only (no host and no meaningful path, e.g. `myapp://`, `https://`).
+    private func isUnusable(_ url: URL?) -> Bool {
+        guard let url else { return true }
+        let trimmed = url.absoluteString.trimmingCharacters(in: .whitespacesAndNewlines)
+        if trimmed.isEmpty { return true }
+        let hasHost = !(url.host?.isEmpty ?? true)
+        let hasPath = !url.path.isEmpty && url.path != "/"
+        return !hasHost && !hasPath
     }
 }
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -108,7 +108,7 @@ public class PayPalWebCheckoutClient: NSObject {
         sessionTask = nil
         configValidationError = nil
 
-        guard !urlConfig.isReturnToAppConfigMissing else {
+        if urlConfig.isReturnToAppConfigMissing {
             configValidationError = PayPalError.returnToAppURLConfigMissingError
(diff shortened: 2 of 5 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -102,6 +105,17 @@ public class PayPalWebCheckoutClient: NSObject {
         userAction: PayPalUserAction = .continue
     ) {
         sessionTask?.cancel()
+        sessionTask = nil
+        configValidationError = nil
+
+        guard !urlConfig.isReturnToAppConfigMissing else {
```
NastassiaRodzik: A bit easier to read would be 
```
if urlConfig.isReturnToAppConfigMissing {
... 
return 
}
```

same on line 154

EdgarBarocio: can change this line, but on line 154 it reads the stored value defined on line 45

Sources/PayPalWebPayments/Models/PayPalURLConfig.swift
```diff
@@ -13,3 +13,26 @@ public struct PayPalURLConfig {
...
+
+extension PayPalURLConfig {
+
+    /// True when neither `returnAppURL` nor `fallbackSchemeURL` is usable.
+    /// A URL is unusable when it's nil, blank, or "scheme-only" (no host and no path,
+    /// e.g. `myapp://` or `https://`) — i.e. it can't route the buyer back into the app.
+    /// Defensive early-fail so `start()`/`vault()` can fail immediately without a network call;
+    /// this is a guard, not full parity with Android's nullable `ReturnToAppUrlConfig`.
+    var isReturnToAppConfigMissing: Bool {
+        isUnusable(returnAppURL) && isUnusable(fallbackSchemeURL)
```
sshropshire, line 25: question: Isn't a fallback scheme like `my-custom-scheme://` valid? At the moment, we do allow this type of scheme on Android.

EdgarBarocio: ~~the scheme is valid on iOS if it has a path or host after `my-custom-scheme://`~~

EDIT: My mistake, yes, my-custom-scheme:// is a valid url, but we're not testing for validity, we're testing for usability. That means having a path or a host after the url scheme.

sshropshire: Is that a limitation on the iOS `URL` type?

sshropshire: On Android, we just replace the `https` in either `returnAppURL` or `cancelAppURL` with `my-custom-scheme` when constructing a fallback URL. Is it the same on iOS?

EdgarBarocio: you are looking for url validity, the url is valid if it has http, https or the custon scheme, we're evaluating for usability.

empty path or empty host. This is done due to the API not returning an optional url, we have to check for empty string or any other malformations.

sshropshire: Is the custom url scheme path really significant though? I'm looking at [ShopperSessionURLConfigFactory.swift#L35](https://github.com/EdgarBarocio/paypal-ios/blob/ff665807bc468bf2c10238cf270c781630193808/Demo/Demo/SwiftUIComponents/CommonComponents/ShopperSessionURLConfigFactory.swift#L35). I'm pretty sure `x-callback-url/paypal-sdk/paypal-checkout` has no actual meaning. It's an honest question: what would happen if we changed this? Or better yet if we omitted `checkoutPath` from fallback url construction.

EdgarBarocio: sorry if I'm missunderstanding, but what is the ask about the path?

EdgarBarocio: the purpose of the PR is fail early if we don't have a correct returnAppUrl or fallbackSchemeUrl. path or fallback are not into consideration, we're checking after everything has been built, before actually making the call.

It is strictly defensive coding because the SDK does not seem to return an "empty" returnAppUrl, so we're checking for malformations.

sshropshire: Like what happens if a merchant puts `my-custom-scheme://` as a fallback URL? Why do we fail in that case? I feel like we have enough information to make it work. Otherwise providing a path becomes an implicit API parameter e.g. `my-custom-scheme://just/to/make/things/work`.

EdgarBarocio: the part of `just/to/make/things/work` determines the actions the app will take.

The OS checks if it can use `my-custom-scheme://` and when it finds it can use it, the path determines the iOS app action. eg: navigation, app actions, analytics tracking. 

The code changes are validating that it does have a valid action to take, that it has something. in case there is nothing to do, we fail early. 


The main objective was to mimic android behavior, where they also fail early in case of an invalid return or fallback.

Invalid being the nebulous parameter.

sshropshire: Oh I wasn't aware this is happening on Android now too. Do you have a link to where we're enforcing this on Android so I can review? Before, we allowed `custom-url-scheme://` by itself. The path may be helpful, but for a merchant with a single payment method integration i.e. PayPal, checkout-only integration, the path would be irrelevant.

The path portion is merchant domain, my argument is that we shouldn't be so strict. If a merchant needs to make a change and add a path to distinguish between different types of URLs, they'll figure it out in their own QA process.

To give an example of a merchant app bug, it seems like there's a defect in the Demo app currently. Under the logic you've described, we should have separate URLs for both checkout and vault if I'm following correctly? Also the return and cancel URLs seem to lack the same path information that could be useful for determining the source payment method.

EdgarBarocio: found the following reference

`Android reference implementation: paypal-android PayPalWebCheckoutClient.kt (createPayPalSession eager validation) and GetReturnToAppStrategyUseCase.kt (require() safety net), plus new PayPalWebCheckoutErrorCode.RETURN_TO_APP_URL_CONFIG_MISSING / PayPalWebCheckoutError.returnToAppUrlConfigMissingError. `

EdgarBarocio: the android change: https://github.com/paypal/paypal-android/pull/466/changes

sshropshire: Yeah you can see here in [DemoConstants.kt#L12](https://github.com/ayer-ribeiro/paypal-android/blob/3eb86f977a5424179d385801196891e21d685aeb/Demo/src/main/java/com/paypal/android/DemoConstants.kt#L12) we allow just the scheme to be specified on Android. The `://` isn't even required.

EdgarBarocio: we're having a discussion, this change may not be needed at all because this sdk never allows empty urls, the android one does.

PR #427, opened 2026-07-28
Add analytics tracking to cancel button on Login popup

EdgarBarocio:
### Reason for changes

There is an ask from the Analytics team: to report paypal:tokenize:browser-login:alert-canceled event by analogy with BT Native iOS

### Summary of changes

- Added tracking on cancel button when the alert is shown in Vaulting, only added on vaulting because this is already tracked on web checkout flow.

### Checklist

- [ ] Added a changelog entry

### Authors

- EdgarBarocio


MikeThorntonPayPal (approved):

NastassiaRodzik (changes requested):
The event name should be: 
`paypal:tokenize:browser-login:alert-canceled`

Also please check / retest the use case: it needs to be triggered on Cancel button click on web redirect popup.

MikeThorntonPayPal (approved):

PR #429, opened 2026-07-28
Analytics reporting when cancelling flow.

EdgarBarocio:
### Reason for changes

There is an ask from the Analytics team: to report paypal:tokenize:browser-login:alert-canceled event by analogy with BT Native iOS

### Summary of changes

Added tracking on cancel button when the alert is shown in Vaulting, only added on vaulting because this is already tracked on web checkout flow.

### Checklist

- [ ] Added a changelog entry

### Authors

EdgarBarocio

### DIFF
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -972,8 +972,8 @@ public class PayPalWebCheckoutClient: NSObject {
 
     private func sendBrowserLoginCancelEvent(errorDescription: String?) {
         let eventName = webSessionReturned
-            ? "paypal-web-payments:checkout:browser-login:canceled"
-            : "paypal-web-payments:checkout:browser-login:alert-canceled"
+            ? "paypal:tokenize:browser-login:canceled"
+            : "paypal:tokenize:browser-login:alert-canceled"
         analyticsService?.sendEvent(
             eventName,
             errorDescription: errorDescription,
--- a/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
+++ b/Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
@@ -1,4 +1,5 @@
 import AuthenticationServices
+import UIKit
 
 #if canImport(CorePayments)
 import CorePayments
@@ -42,6 +43,10 @@ public class PayPalWebCheckoutClient: NSObject {
 
     // MARK: - Analytics State
     private var analyticsData: PayPalCheckoutAnalyticsData?
+
+    /// Whether the web auth modal (close button) appeared after the consent alert.
+    private var webAuthSheetPresented = false
+    private var webAuthModalPollTimer: Timer?
 
     // MARK: - Initializer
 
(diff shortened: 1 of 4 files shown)


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -43,6 +44,9 @@ public class PayPalWebCheckoutClient: NSObject {
     // MARK: - Analytics State
     private var analyticsData: PayPalCheckoutAnalyticsData?
 
+    /// Indicates whether the buyer progressed past the consent alert into the web auth modal.
+    private var webSessionReturned = false
```
sshropshire: question: Does app switch present a consent dialog?

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -663,6 +690,10 @@ public class PayPalWebCheckoutClient: NSObject {
                         )
                     }
                 },
+                sessionDidCancel: { [weak self] in
+                    guard let self else { return }
+                    self.handleCheckoutWebAuthCancel(completion: completion)
```
sshropshire, line 695:
```suggestion
                    self?.handleCheckoutWebAuthCancel(completion: completion)
```

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
EdgarBarocio: no, this flow is for when the app is not installed, only on web

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
EdgarBarocio: `guard let self else { return }` already makes self safe, no need for a `?`

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```suggestion
    private var didApplicationBecomeActive = false
```

EdgarBarocio: updated

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
sshropshire: I'm cook skipping this–I noticed this is the standard elsewhere in the codebase. To me though it reads more cleanly (1 line vs. 2 lines) and doesn't look like a strong reference.

sshropshire (approved):

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
EdgarBarocio: it's a swift thing. The [weak self] makes self a weak reference inside the block, the line of 
`guard let self else { return }` is a shorthand, it used to be `guard let self = self else { return }` and since the guard already unwraps it for you, you don't need the `self?.` for inline unwrapping

sshropshire: Yeah it's syntactic sugar for creating a shadow variable. Inline unwrapping feels more expressive. That single `?` character essentially does the same thing that the guard does.

PR #431, opened 2026-07-29
Deep deprecated methods clean up

NastassiaRodzik:
### Reason for changes

- Remove Deprecated methods before V3 release 

### Summary of changes

- Deprecated methods, unused data structures are removed

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik


Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
```diff
@@ -130,6 +121,7 @@ public class PayPalWebCheckoutClient: NSObject {
         userAction: PayPalUserAction = .continue
     ) {
         sessionTask?.cancel()
+        systemLatency.reset()
```
sshropshire, line 124: q: Was this a bug fix?

sshropshire (approved):

Sources/PayPalWebPayments/PayPalWebCheckoutClient.swift
NastassiaRodzik: Not a fix for a reported bug, it's defensive: `begin()` always overwrites startTime and `send()` is guarded by it, so there's no double-send/wrong-duration in the normal path. But the session-fetch-failure branch begins a measurement without ever ending it, and `send()` leaves flow set, so tracker state could linger across flows. `createPayPalSession()` is the entry point for each new flow and already cancels the prior sessionTask, so resetting there guarantees a clean slate. Prevents a latent stale-measurement edge case rather than fixing an observed one.

sshropshire: Got it. Yeah that's the tricky part about having stateful clients.

PR #441, opened 2026-08-11
Fix analytics events

NastassiaRodzik:
### Reason for changes

Request from Analytics team

### Summary of changes

- Log proper analytics event on web fallback cancellation 

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Sources/PayPalPayments/PayPalClient.swift
+++ b/Sources/PayPalPayments/PayPalClient.swift
@@ -588,6 +588,10 @@ public class PayPalClient: NSObject {
         setupTokenID: String,
         completion: @escaping (Result<PayPalVaultResult, CoreSDKError>) -> Void
     ) {
+        analyticsService?.sendEvent(
+            "paypal-payments:checkout:auth-challenge-presentation:started",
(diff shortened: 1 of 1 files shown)


Sources/PayPalPayments/PayPalClient.swift
```diff
@@ -650,7 +650,7 @@ public class PayPalClient: NSObject {
         let sdkError = PayPalError.checkoutCanceledError
         sendBrowserLoginCancelEvent(errorDescription: sdkError.errorDescription)
         analyticsService?.sendEvent(
-            "paypal-payments:checkout:handle-return:canceled",
+            "paypal-payments:checkout:handle-return:succeeded",
```
sshropshire, line 657: Hey is this correct? It looks like we're in an error block.

NastassiaRodzik: Yep this is for `Checkout Cancelled` scenario.
As we were able to return to merchant successfully, `handle-return:succeeded` is logged here

sshropshire (approved):

PR #445, opened 2026-08-11
Enable universal links Live env

NastassiaRodzik:
### Reason for changes

Universal links on Live do not work 

### Summary of changes

- Update entitlements with Prod website 
- Made changes to pass needed return/cancel urls

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Demo/Demo/Models/CreateOrderParams.swift
+++ b/Demo/Demo/Models/CreateOrderParams.swift
@@ -49,6 +49,7 @@ struct PayPalExperienceContext: Encodable {
 
     let returnUrl: String
     let cancelUrl: String
+    let userAction: String
     var nativeApp: NativeApp?
 }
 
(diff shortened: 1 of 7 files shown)


Demo/Demo/Models/CreateOrderParams.swift
```diff
@@ -49,6 +49,7 @@ struct PayPalExperienceContext: Encodable {
 
     let returnUrl: String
     let cancelUrl: String
+    let userAction: String
```
kgangineni: we don't need to send `userAction` this after fix in the backend

NastassiaRodzik: @kgangineni but we're still seeing this issue on Sandbox and most importantly, Live, sending you the ticket with details

NastassiaRodzik: reverted

kgangineni (approved):

PR #447, opened 2026-08-19
Add V3.0.0 iOS integration guides

naolson-paypal:
## Summary
- Ports the new-structure V3.0.0 iOS docs (Install & Setup, PayPal Checkout, Payment Buttons, Card/ACDC, Migrate V2→V3, Troubleshooting, Concepts & Responsibilities, Which Guide Do I Use) out of Confluence into `Guides/`, with internal links rewritten to work on GitHub.
- Organized into a folder hierarchy (`Guides/README.md` top-level entry, `getting-started/`, `integration-guides/`) so the intended reading order is explicit rather than relying on a flat file list.
- Every technical claim was fanned out and validated against the actual App Switch source (now merged into `main` via #358) — CoreConfig shape, PayPalClient/PayPalURLConfig signatures, cancellation handling, etc. — and corrected where the guides didn't match reality.
- Venmo is intentionally excluded from this PR — it lives on a separate, still-diverging branch. Those docs live on `docs/v3-venmo-guides`, ready for the follow-up release that includes Venmo.

## Test plan
- [ ] Spot-check that relative links resolve when browsing `Guides/` on GitHub
- [ ] Confirm code samples (CoreConfig construction, createPayPalSession/start/handleReturnURL signatures) match the current main API surface


Guides/getting-started/install-and-setup.md
```diff
@@ -0,0 +1,84 @@
...
+)
+```
+
+Reuse this `config` across every method client. Switch `.sandbox` to `.live` for production.
+
+## Step 3: Register your return links
+
+PayPal checkout sends the buyer to the PayPal app (or an in-app browser) and back to your app via a Universal Link. In **App Target › Signing & Capabilities › Associated Domains**, add `applinks:example.com` and serve your AASA file at `https://example.com/.well-known/apple-app-site-association`. One domain covers both your return and cancel URLs.
+
+Register the custom-scheme fallback under `CFBundleURLTypes`, and declare the PayPal app scheme under `LSApplicationQueriesSchemes` so the SDK can detect it. `fallbackSchemeURL` is optional on `PayPalURLConfig`, but you should always set it — without it, a buyer whose Universal Link fails to open your app has no way back to checkout:
```
MikeThorntonPayPal, line 46: nit: this "optional in the API but you should always set it" caveat shows up close to verbatim in at least 3 other spots — concepts-and-responsibilities.md line 30, and migrate-v2-to-v3.md line 86 (plus the inline code comment on line 68 here). Would you be open to stating it once here (since this is where the reader actually configures it) and having the others just link back to this section? Not blocking, just noticed it while reading top to bottom.

Guides/migrate-v2-to-v3.md
```diff
@@ -0,0 +1,102 @@
...
++     guard let url = userActivity.webpageURL else { return }
++     client.handleReturnURL(url)
++ }
+```
+
+## Info.plist and entitlements
+
+Add the custom-scheme `fallbackSchemeURL` under `CFBundleURLTypes` (optional on `PayPalURLConfig`, but you should always set it), and declare `paypal` under `LSApplicationQueriesSchemes` so the SDK can detect the PayPal app. Add the Associated Domains entitlement for your return domain. (V2's web flow needed none of these for PayPal Checkout because its return handling was internal to `ASWebAuthenticationSession`, built on the hardcoded `sdk.ios.paypal` scheme.) See [Install & Setup (iOS)](getting-started/install-and-setup.md).
+
+Note that `sdk.ios.paypal` isn't retired in V3 — it's still the SDK's internal callback scheme for Card's 3D Secure `ASWebAuthenticationSession` challenge, and it also still backs the PayPal Checkout in-app-browser fallback's own redirect internally. What changed is the buyer-facing, merchant-configured return path for PayPal Checkout and Vault: that's now Universal Links (with `fallbackSchemeURL` as backup), not something built on `sdk.ios.paypal` that you need to register or handle yourself.
```
MikeThorntonPayPal, line 88: This callout is genuinely useful — that's exactly the kind of leaky-abstraction detail that would otherwise cause a confused Slack thread six months from now. 🌊 Nice catch keeping it in.

MikeThorntonPayPal (approved):

Guides/getting-started/install-and-setup.md
naolson-paypal: Fixed — consolidated into one explanation in Install & Setup, with Concepts & Responsibilities and the migrate guide now linking back to it instead of repeating it. See f086f3c.

Guides/migrate-v2-to-v3.md
naolson-paypal: Glad it's useful — that exact gap (sdk.ios.paypal quietly still doing work under the hood) seemed worth calling out explicitly rather than leaving implicit.

PR #449, opened 2026-08-20
Remove deprecated canOpenURL(:) method

NastassiaRodzik:
### Reason for changes

Apple deprecated `canOpenURL(:)` method so we need to stop relying on it 

### Summary of changes

- Remove `isPayPalAppInstalled()` method
- Pass `true` to `createShopperSession()` as well as `attemptSessionAppSwitchOrFallback()` so flow attempts app switch every time 
- Pass `paypal_installed = null` to Analytics 

### Checklist

- [ ] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Sources/CorePayments/UIApplication+URLOpener.swift
+++ b/Sources/CorePayments/UIApplication+URLOpener.swift
@@ -7,7 +7,6 @@ public protocol URLOpener {
         universalLinksOnly: Bool,
         completionHandler completion: ((Bool) -> Void)?
     )
-    func isPayPalAppInstalled() -> Bool
 }
 
 public extension URLOpener {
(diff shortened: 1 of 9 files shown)


Sources/CorePayments/UIApplication+URLOpener.swift
```diff
@@ -19,13 +18,6 @@ public extension URLOpener {
 
 extension UIApplication: URLOpener {
 
-    public func isPayPalAppInstalled() -> Bool {
```
sshropshire, line 22: Hey this is technically a breaking change. We may not be able to do this until v4.

MikeThorntonPayPal (approved):

Sources/CorePayments/UIApplication+URLOpener.swift
NastassiaRodzik: Upd: deprecated this method instead of removing it as per internal discussion

PR #450, opened 2026-08-21
[Patch] Pass proper user action to createShopperSession

NastassiaRodzik:
### Reason for changes

- We noticed userAction PayNow being applied to Continue sessions 

### Summary of changes

- Pass selected user action to `createShopperSession()`

### Checklist

- [x] Added a changelog entry

- NastassiaRodzik

### DIFF
--- a/Sources/PayPalPayments/APIRequests/CreateShopperSessionVariables.swift
+++ b/Sources/PayPalPayments/APIRequests/CreateShopperSessionVariables.swift
@@ -41,7 +41,7 @@ struct ShopperSessionExperimentationContext: Encodable {
         integrationChannel: String = PayPalCoreConstants.integrationChannel,
         isWebLLSEligible: Bool = false,
         isWebView: Bool = false,
-        paymentType: String = "PAY",
+        paymentType: String,
         buyerGUID: String? = nil,
         merchantAccountId: String
     ) {
--- a/Sources/PayPalPayments/Models/PayPalUserAction.swift
+++ b/Sources/PayPalPayments/Models/PayPalUserAction.swift
@@ -20,3 +20,17 @@ public enum PayPalUserAction {
         }
     }
 }
+
+extension PayPalUserAction {
+    
+    var externalPaymentType: String {
+        switch self {
+        case .continue:
+            return "CONTINUE"
+        case .payNow:
+            return "PAY"
+        case .setupNow:
+            return "COMMIT"
+        }
+    }
+}
--- a/UnitTests/TestShared/MockCreateShopperSessionAPI.swift
+++ b/UnitTests/TestShared/MockCreateShopperSessionAPI.swift
@@ -9,6 +9,7 @@ class MockCreateShopperSessionAPI: CreateShopperSessionAPI {
 
     // Captured call arguments for assertion
     var capturedURLConfig: PayPalURLConfig?
+    var capturedUserAction: PayPalUserAction?
     var capturedUserIdentity: PayPalUserIdentity?
     var callCount = 0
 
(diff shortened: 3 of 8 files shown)


UnitTests/TestShared/MockCreateShopperSessionAPI.swift
```diff
@@ -9,20 +9,23 @@ class MockCreateShopperSessionAPI: CreateShopperSessionAPI {
...
+        userAction: PayPalUserAction,
         userIdentity: PayPalUserIdentity?,
         analyticsData: PayPalCheckoutAnalyticsData? = nil
     ) async throws -> ShopperSessionResult {
         callCount += 1
         capturedURLConfig = urlConfig
         capturedUserIdentity = userIdentity
-
+        capturedUserAction = userAction
+
```
MikeThorntonPayPal, line 28: nit: trailing whitespace on this blank line, swiftlint should flag it but figured I'd save you the ci round trip

Sources/PayPalPayments/Models/PayPalUserAction.swift
```diff
@@ -20,3 +20,17 @@ public enum PayPalUserAction {
         }
     }
 }
+
+extension PayPalUserAction {
+    
+    var externalPaymentType: String {
+        switch self {
```
MikeThorntonPayPal, line 27: random q: now that we have two switches over the same three cases in this file (title and externalPaymentType), any thought given to backing both off a single source, like a struct/tuple keyed by case, so we're not maintaining two parallel string tables that can drift. not blocking, just flagging since these two happened to diverge already (PAY vs PAY NOW)

Sources/PayPalPayments/APIRequests/CreateShopperSessionVariables.swift
```diff
@@ -41,7 +41,7 @@ struct ShopperSessionExperimentationContext: Encodable {
         integrationChannel: String = PayPalCoreConstants.integrationChannel,
         isWebLLSEligible: Bool = false,
         isWebView: Bool = false,
-        paymentType: String = "PAY",
+        paymentType: String,
```
MikeThorntonPayPal, line 44: q: dropping the default here is correct given the bug, just double checked and CreateShopperSessionAPI.swift is the only caller so we're fine, but worth calling out in case anyone adds a second call site later and forgets this now has to be explicit

MikeThorntonPayPal (commented):
this looks good, makes sense that continue/setupNow were silently getting PAY sent to the experimentation context since the param used to default to it. tests cover all three user actions which is nice.

should this get a changelog entry though? it's a behavior fix for the shopper session that's currently live, feels similar to the cancel opType fix we logged back in 2.0.1, checklist has it unboxed

left two nits/questions inline, nothing blocking. 🌊

UnitTests/TestShared/MockCreateShopperSessionAPI.swift
NastassiaRodzik: swift lint is not complaining on this one, marking as resolved

Sources/PayPalPayments/APIRequests/CreateShopperSessionVariables.swift
NastassiaRodzik: yes it's the only caller + an internal method so we are fine here

Sources/PayPalPayments/Models/PayPalUserAction.swift
NastassiaRodzik: them being diverged is fine - one is for backend and another is for Demo UI

MikeThorntonPayPal (approved):

PR #454, opened 2026-09-21
Migrate to Nav Stack

sshropshire:
### Reason for changes

This PR migrates the iOS Demo application to use SwiftUI `@Environment` variables and `NavStack`

### Summary of changes

- Update Demo project settings to include PayPal SDK as a local Swift Package Manager package
- Migrate `FeatureSelectionView` to use SwiftUI `NavigationStack`
- Inject new `CardPaymentViewModel` into `CardPaymentView` to refresh state every time its launched
- Migrate `CardPaymentView` to leverage SwiftUI `@Environment` for dependency injection

**Before**

https://github.com/user-attachments/assets/fbf430a4-0fe8-4660-925b-ffcf87e696c5

**After**

https://github.com/user-attachments/assets/1548cf95-3055-4806-877f-49f1c31117d2

### Checklist

- @sshropshire


Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
```diff
@@ -37,39 +48,34 @@ struct FeatureSelectionView: View {
...
-                    }
-                    NavigationLink {
-                        SwiftUIPaymentButtonDemo()
-                    } label: {
-                        Text("Payment Button")
-                    }
+                    NavigationLink("Card Payment", value: Feature.cardPayment)
+                    NavigationLink("Card Vaulting", value: Feature.cardVaulting)
+                    NavigationLink("PayPal Web", value: Feature.payPalWeb)
+                    NavigationLink("PayPal Vaulting", value: Feature.payPalWeb)
```
MikeThorntonPayPal: 🐛 Think this is a copy/paste bug — this should be `Feature.payPalVaulting`. Right now tapping "PayPal Vaulting" routes to `.payPalWeb` in the `navigationDestination` switch below (line 69), so the PayPal Vaulting screen is unreachable. Worth double checking in the demo since it'd be an easy one to miss on the simulator if you didn't tap through every row.

Demo/Demo/CardPayments/CardPaymentUiState.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftUI
+import CardPayments
+
+@Observable
+class CardPaymentUiState {
```
MikeThorntonPayPal: nit: `UiState` reads very Android/Compose to me (that's basically the Jetpack `ViewModel` + `UiState` split). Given this is `@Observable` and injected straight into the environment, could we just fold this into `CardPaymentViewModel` directly, or if we want the split, name it something more idiomatic like `CardPaymentState` (which is literally what the old type was called)? Would love to keep this swift-y rather than importing the Android pattern verbatim.

Demo/Demo/CardPayments/CardPaymentView.swift
```diff
@@ -0,0 +1,130 @@
...
+                LeadingText("Order Status", weight: .bold)
+                LeadingText("\(status)")
+            }
+            LeadingText("didAttemptThreeDSecureAuthentication", weight: .bold)
+            LeadingText("\(cardResult.didAttemptThreeDSecureAuthentication)")
+        }
+    }
+}
+
+struct CompleteOrder: View {
```
MikeThorntonPayPal: nit: `CreateOrderForm` and `ApproveOrderForm` (👍 like those names) both signal "this is a form", but this one is `CompleteOrder` — no `Form`/`View` suffix, and it reads like a verb phrase rather than a type name. Would `CompleteOrderForm` be more consistent with its siblings?

Demo/Demo/CardPayments/CardPaymentViewModel.swift
```diff
@@ -0,0 +1,154 @@
...
+    
+    var approveOrderState: AsyncState<CardResult> {
+        get { uiState.approveOrderState }
+        set { uiState.approveOrderState = newValue }
+    }
+    
+    var completeOrderState: AsyncState<Order> {
+        get { uiState.completeOrderState }
+        set { uiState.completeOrderState = newValue }
+    }
```
MikeThorntonPayPal: q: is there a reason we need 5 pass-through `get`/`set` properties here instead of just exposing `uiState` on the environment and having the views read/write `viewModel.uiState.createOrderRequest` directly? As written, `CardPaymentViewModel` isn't really adding behavior on top of `CardPaymentUiState` for these — it's just forwarding. If `uiState` is meant to be a separable/testable piece, exposing it directly (or even just merging the two types) would remove ~25 lines of boilerplate that has to stay in sync by hand.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
```diff
@@ -0,0 +1,154 @@
...
+    
+    // HACK: this is used to drive the scroll-to-bottom animation
```
MikeThorntonPayPal: 🧶 take-or-leave: the comment already flags this as a hack, so calling it out — this is a UI concern (scroll position) living on the view model, derived from three separate state properties that have to be kept in the right order. Old code just did `.onChange(of: state)` on the whole (`Equatable`) state struct. Could we make `AsyncState` `Equatable` when `T: Equatable` and go back to `.onChange(of: viewModel.uiState)`-style triggering instead of this derived counter? Would remove the hack entirely rather than just labeling it.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
```diff
@@ -0,0 +1,154 @@
...
+                    expirationDate: request.cardExpirationDate,
+                    cvv: request.cardCVV
+                )
+                let cardRequest = CardRequest(orderID: orderID, card: card, sca: request.sca)
+                let result = try await cardClient.approveOrder(request: cardRequest)
+                approveOrderState = .loaded(result)
+                
+                // update card client reference
+                // TODO: make CardClient non-null
+                self.cardClient = cardClient
```
MikeThorntonPayPal: q: `self.cardClient` only gets assigned *after* `approveOrder` succeeds — so what's it actually used for downstream? If nothing reads it later, can we drop the stored property and the local/self split entirely? If something outside this file needs it (e.g. a 3DS continuation), a comment saying so would help future-us. Also — since there's already a `// TODO: make CardClient non-null` here, should this go in the "before merge" checklist or a tracked ticket instead of living as a comment? (Same q for the `// TODO: might need to pass in payee...` on line 72 and `// TODO: differentiate error from cancellation state` on line 116 — don't want these to just quietly ride along forever.)

Demo/Demo/SwiftUIComponents/CommonComponents/SegmentedEnumPicker.swift
```diff
@@ -0,0 +1,17 @@
+import SwiftUI
+
+struct SegmentedEnumPicker<T: RawRepresentable<String> & Hashable & CaseIterable>: View {
+    
+    let label: String
+    let selection: Binding<T>
+    private let values = Array(T.allCases)
+    
+    var body: some View {
```
MikeThorntonPayPal: nit: since `T` is constrained to `RawRepresentable<String>`, we've already got a stable, meaningful identifier sitting right there — `id: \.rawValue` reads a lot more intentionally than `\.hashValue` and avoids relying on synthesized hashing for view identity.

Demo/Demo/SwiftUIComponents/CommonComponents/ContainerView.swift
```diff
@@ -0,0 +1,14 @@
+import SwiftUI
+
+// Ref: https://www.swiftbysundell.com/tips/creating-custom-swiftui-container-views/
+protocol ContainerView: View {
```
MikeThorntonPayPal: q: `FormGroup` looks like the only conformer in this PR — is this protocol earning its keep yet, or could `FormGroup` just take a `@ViewBuilder` closure directly for now? Not opposed to the pattern (nice article link, by the way 🔥) if we know more container views are coming, just want to make sure we're not adding a layer of indirection for a single type.

Demo/Demo/SwiftUIComponents/CommonComponents/RoundedBlueButtonStyle.swift
```diff
@@ -2,12 +2,14 @@ import SwiftUI
 
 struct RoundedBlueButtonStyle: ButtonStyle {
     
+    var isLabelVisible = true
+    
     func makeBody(configuration: Configuration) -> some View {
         configuration.label
-            .foregroundColor(.white)
+            .foregroundColor(isLabelVisible ? .white : .clear)
```
MikeThorntonPayPal: q: since the label is only made `.clear` (not removed) while loading, will VoiceOver still announce the button's title underneath the spinner? Might be worth an `.accessibilityLabel("Loading")`-type override while `isLabelVisible == false`, otherwise this could read confusingly for VoiceOver users mid-request.

Demo/Demo.xcodeproj/project.pbxproj
```diff
@@ -895,7 +888,7 @@
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPad = "UIInterfaceOrientationPortrait UIInterfaceOrientationPortraitUpsideDown UIInterfaceOrientationLandscapeLeft UIInterfaceOrientationLandscapeRight";
 				INFOPLIST_KEY_UISupportedInterfaceOrientations_iPhone = UIInterfaceOrientationPortrait;
 				INFOPLIST_KEY_UIUserInterfaceStyle = Light;
-				IPHONEOS_DEPLOYMENT_TARGET = 16.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 17.0;
```
MikeThorntonPayPal: q: bumping the Demo app's min target from 16.0 → 17.0 (presumably for `@Observable`/`@Bindable`) is a reasonable trade for the SwiftUI rewrite, but it does mean we lose demo-app coverage on iOS 16 devices/simulators. Worth a line in the PR description calling this out explicitly so it's a conscious decision on record, not just a side effect of adopting the new environment APIs?

Sources/CardPayments/CardClient.swift
```diff
@@ -67,7 +67,11 @@ public class CardClient: NSObject {
                     analyticsService?.sendEvent("card-payments:vault-wo-purchase:auth-challenge-required")
                     startVaultThreeDSecureChallenge(url: url, setupTokenID: vaultRequest.setupTokenID, completion: completion)
                 } else {
-                    let vaultResult = CardVaultResult(setupTokenID: result.id, status: result.status, didAttemptThreeDSecureAuthentication: false)
+                    let vaultResult = CardVaultResult(
+                        setupTokenID: result.id,
+                        status: result.status,
+                        didAttemptThreeDSecureAuthentication: false
```
MikeThorntonPayPal: nit: this (and the similar reflow in `AnalyticsService.swift` and `CardClient_Tests.swift`) is a whitespace/line-wrap-only change unrelated to the nav-stack migration. Not blocking, just flagging since it widens the diff outside `Demo/` — might've come from running SwiftLint/swiftformat locally. Fine to leave if intentional cleanup, just want to make sure it wasn't accidental.

MikeThorntonPayPal (approved):

Demo/Demo/CardPayments/CardPaymentUiState.swift
sshropshire: I agree yeah. This may not be needed I'm looking into finding a better SwiftUI solution.

Demo/Demo/CardPayments/CardPaymentView.swift
sshropshire: Good catch yeah I can update.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
sshropshire: We may not yeah good call. This is basically a result of the Android port. There might be a better way I'll update this.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
sshropshire: Yeah there was a reason why I had to do this I'll need to investigate. We might be kinda stuck though if we end up getting rid of UiState. On Android, we do this after every render, but I don't think there's isn't an equivalent way to do that in SwiftUI.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
sshropshire: Yeah I'm not sure why this is necessary. I may have to look at `main` to see if we can remove. I'm not sure what the other comments on line 72 or 116 mean, I'll have to review. Most of this stuff can probably be deleted.

Demo/Demo/SwiftUIComponents/CommonComponents/ContainerView.swift
sshropshire: I'm down for simplifying it.

Demo/Demo/SwiftUIComponents/CommonComponents/RoundedBlueButtonStyle.swift
sshropshire: Yeah I'm not sure tbh. I can add the accessibility label that's a good call.

Demo/Demo/SwiftUIComponents/CommonComponents/SegmentedEnumPicker.swift
sshropshire: Yeah that's true it's cleaner that way. I can look into updating.

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
sshropshire: Yeah good call. This was a mistake.

Sources/CardPayments/CardClient.swift
sshropshire: Yeah I'm not sure if this was auto formatting or what, but honestly there are some really long lines in this repo it'd be nice to enforce a column limit to make everything more split-screen friendly.

Demo/Demo/SwiftUIComponents/FeatureSelectionView.swift
sshropshire: This is resolved in [615bc0a](https://github.com/paypal/paypal-ios/pull/454/commits/615bc0af2e9d5f651af764173fc6b16d48da19f4).

Demo/Demo/CardPayments/CardPaymentUiState.swift
sshropshire: I found a way to just get rid of `CardPaymentUiState`.

Demo/Demo/CardPayments/CardPaymentView.swift
sshropshire: This is resolved in [a0d3b39](https://github.com/paypal/paypal-ios/pull/454/commits/a0d3b3993f3083cd34c4a340cf453beee25b8ee4).

Demo/Demo/CardPayments/CardPaymentViewModel.swift
sshropshire: This got cleaned up after removing `CardPaymentUiState`.

Demo/Demo/CardPayments/CardPaymentViewModel.swift
sshropshire: I figured out a way to do this with a hash value of the relevant state.

